PR #21, opened 2015-01-06
Add module `nailgun.config`

### LINKED ISSUE #2: Create a config file
Certain information that was available in Robottelo must also be available in NailGun. For example, user credentials and the server hostname must be available. In Robottelo, this information is available via the robottelo.properties file (see robottelo.properties.sample). A good strategy might be to give NailGun its own settings file which has the same syntax as that use by Robotteo. In that case, NailGun can either operate stand-alone, or NailGun's settings file can be symlinked to Robottelo's config file.

Ichimonji10:
The module contains "tools for managing and presenting server connection
configurations." Most importantly, this module contains a class called
`ServerConfig`, and it does all the heavy lifting.

Include the new module in the documentation.

Add unit tests for the new module.

Fix #2.

### DIFF
--- a/docs/api/nailgun.rst
+++ b/docs/api/nailgun.rst
@@ -8,3 +8,9 @@
 
 .. automodule:: nailgun.client
    :members:
+
+:mod:`nailgun.config`
+=====================
+
+.. automodule:: nailgun.config
+   :members:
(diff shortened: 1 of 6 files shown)


omaciel:
Very nice implementation and design! ACK

elyezer:
Well cone. ACK

PR #23, opened 2015-02-27
Generate ``DateField`` and ``DatTimeField``.

omaciel:
Added code that generates ``DateField`` and ``DatTimeField`` entities.
This will be useful for synchronization plans and schedule tasks in
``Robottelo``.

### DIFF
--- a/nailgun/entity_fields.py
+++ b/nailgun/entity_fields.py
@@ -46,6 +46,8 @@ UTF-8 values, which is unpleasant to work with.
 from fauxfactory import (
     gen_boolean,
     gen_choice,
+    gen_date,
+    gen_datetime,
     gen_email,
     gen_integer,
     gen_ipaddr,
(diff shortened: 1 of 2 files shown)


elyezer:
Good addition. We also need something that parses the synchronization times like `hourly`

omaciel:
The ``Hourly`` parsing would be handled by the ``entity``.

elyezer:
:+1:

nailgun/entity_fields.py
```diff
@@ -198,10 +200,32 @@ def gen_value(self):
...
 
+    def __init__(self, min_date=None, max_date=None):
+        # If ``None`` is passed then ``FauxFactory`` will deal with it.
+        self.min_date = min_date
+        self.max_date = max_date
+
+    def gen_value(self):
+        """Return a value suitable for a :class:`DateTimeField`."""
+        # pylint:disable=no-self-use
+        # See comment near top of module.
```
Ichimonji10: You should be able to drop these two lines of commenting. This method does make use of `self`.

nailgun/entity_fields.py
```diff
@@ -198,10 +200,32 @@ def gen_value(self):
...
 
+    def __init__(self, min_date=None, max_date=None):
+        # If ``None`` is passed then ``FauxFactory`` will deal with it.
+        self.min_date = min_date
+        self.max_date = max_date
+
+    def gen_value(self):
+        """Return a value suitable for a :class:`DateField`."""
+        # pylint:disable=no-self-use
+        # See comment near top of module.
```
Ichimonji10: You should be able to drop these two lines of commenting. This method does make use of `self`.

Ichimonji10:
ACK pending comments

omaciel:
Comments should be addressed.

Ichimonji10:
Waiting on Travis...

PR #43, opened 2015-04-21
added method to delete content-view from specific environment

sghai:


nailgun/entities.py
```diff
@@ -643,7 +643,7 @@ class Meta(object):
         server_modes = ('sat')
 
 
-class ContentViewVersion(Entity, EntityReadMixin):
+class ContentViewVersion(Entity, EntityReadMixin, EntityDeleteMixin):
```
Ichimonji10: This is good and makes sense. It allows us to make use of `DELETE /katello/api/content_view_versions/:id`.

Ichimonji10:
Below is a tweaked version of what you've provided, with an explanation. Can you try giving it a go?

* It's really nice if we can deal with abstractions like ``Environment`` objects instead of working with IDs all the time. As a result, the method below accepts an argument named "environment".
* You'll want to poll the server or not return JSON, respectively, if an HTTP ACCEPTED or NO CONTENT response is received.
* It's obvious that a content view version is being deleted from somewhere, so mentioning "version" in the method name is unnecessary. Thus, I've renamed the method to `delete_from_environment`. Client code will end up making calls like this: `content_view.delete_from_environment()`.
* Calling `self.path('self')` helps prevent weird errors. The `'self'` argument causes the method to raise an exception if the current object has no ID.
* I pass in `environment_id` as part of the JSON payload instead of in the URL. Either approach works, but I prefer using a JSON payload when possible. JSON is easier to logically manipulate than a URL. (Also, an over-reliance on URL parameters is part of what's led to the API being as large as it is. I'd like to avoid specialty URLs when possible in case the API is ever reined in.)

```python
def delete_from_environment(self, environment, synchronous=True):
    """Delete this content view version from an environment.

    This method acts much like :meth:`nailgun.entity_mixins.EntityDeleteMixin.delete`.
    The documentation on that method describes how the deletion procedure works
    in general. This method differs only in accepting an ``environment`` parameter.

    :param environment: A :class:`nailgun.entities.Environment` object. The
        environment's ``id`` parameter *must* be specified. As a convenience, an
        environment ID may be passed in instead of an ``Environment`` object.

    """
    if isinstance(environment, Environment):
        environment_id = environment.id
    else:
        environment_id = environment
    response = client.delete(
        self.path('self'),
        auth=self._server_config.auth,
        data={'environment_id': environment_id},
        verify=self._server_config.verify,
    )
    response.raise_for_status()
    if (synchronous is True and
            response.status_code == http_client.ACCEPTED):
        return _poll_task(response.json()['id'], self._server_config)
    if response.status_code == http_client.NO_CONTENT:
        return
    return response.json()
```

Ichimonji10:
This post is an aside regarding API design. If you're just trying to get things done, you can safely skip reading this post.

A RESTful way to say "make content view version 5 belong to environments 2, 9 and 10" is this:

    PUT
    /katello/api/v2/content_view_versions/5
    {'environment': [2, 9, 10]}

A method call in the form above works when you want to explicitly list all of the environments that a content view version should belong to. However, it doesn't allow you express thoughts like "make content view version 5 belong to whatever environments it already does, plus environment 11." We can modify the call to make that happen:

    PUT
    /katello/api/v2/content_view_versions/5
    {'environment': {'add': [11]}}

We can even express even more complex statements, like "make content view version 5 belong to whatever environments it already does, minus environment 10, plus environment 11."

    PUT
    /katello/api/v2/content_view_versions/5
    {'environment': {'remove': [10], 'add': [11]}}

The developers chose to let clients express thoughts like that in a different way. As SatelliteQE/robottelo#2096 shows, this method call is possible:

    DELETE
    /katello/api/v2/content_view_versions/:id/environments/:environment_id

And that's what you're doing here. (Annoyingly, the path above is not listed in the API documentation.) The above can also *probably* be expressed as so:

    DELETE
    /katello/api/v2/content_view_versions/:id
    {'environment_id': …}

I think that the latter form is preferable, and it's what I've suggested, but either form should work.

sghai:
Thanks @Ichimonji10 for updating the method with nice explanation. But I'm still not getting how this `client.delete` will work. 

```
        response = client.delete(
            self.path('self'),
            auth=self._server_config.auth,
            data={'environment_id': environment_id},
            verify=self._server_config.verify,
        )
```

And api docs suggest this path:
```
DELETE
/katello/api/v2/content_view_versions/:id/environments/:environment_id
```

sghai:
I'm not sure what I'm missing but when I run test with below method, test failed with error: 
Can not delete <cv_name> due to associated environments: <env_name>

```
response = client.delete(
            self.path('self'),
            auth=self._server_config.auth,
            data={'environment_id': environment_id},
            verify=self._server_config.verify,
        )
```
But it works with:

```
 response = client.delete(
            self.path() + '/environments/' + str(environment_id),
            auth=self._server_config.auth,
            verify=self._server_config.verify,
        )
```

sghai:
Made changes with full path with some string manipulation..

nailgun/entities.py
```diff
@@ -952,6 +952,38 @@ def copy(self, name):
...
+        an environment ID may be passed in instead of an ``Environment``
+        object.
+
+        """
+        if isinstance(environment, Environment):
+            environment_id = environment.id
+        else:
+            environment_id = environment
+        response = client.delete(
+            self.path() + '/environments/' + str(environment_id),
```
Ichimonji10: Thank you. One last request: can you write this like so:

```python
'{0}/environments/{1}'.format(self.path(), environment_id),
```

Ichimonji10:
ACK pending comment.

nailgun/entities.py
```diff
@@ -952,6 +952,38 @@ def copy(self, name):
...
+        This method acts much like
+        :meth:`nailgun.entity_mixins.EntityDeleteMixin.delete`.  The
+        documentation on that method describes how the deletion procedure works
+        in general. This method differs only in accepting an ``environment``
+        parameter.
+
+        :param environment: A :class:`nailgun.entities.Environment` object. The
+        environment's ``id`` parameter *must* be specified. As a convenience,
+        an environment ID may be passed in instead of an ``Environment``
+        object.
```
Ichimonji10: Make sure to tab in each line after the `:param environment:` line. You can call `make docs-html` to check and see if your syntax is correct.

Ichimonji10: :param environment: A :class:`nailgun.entities.Environment` object. The
        environment's ``id`` parameter *must* be specified. As a
        convenience, an environment ID may be passed in instead of an
        ``Environment`` object.

nailgun/entities.py
sghai: :+1:

Ichimonji10:
This looks great, @sghai. Thank you. ACK

By the way, I'm starting to notice how often this hunk of code pops up:

```python
response.raise_for_status()
if (synchronous is True and
        response.status_code == ACCEPTED):
    return _poll_task(response.json()['id'], self._server_config)
if response.status_code == NO_CONTENT:
    return
return response.json()
```

Maybe it would be good to make a helper function that does this...

omaciel:
ACK and I totally agree that we may want to create a helper for this type of repeated code @Ichimonji10

Ichimonji10:
I have created issue #46 in response to the above discussion.

Issue #46, opened 2015-04-23
Refactor out common response handling code


Ichimonji10:
The following hunk of code shows up in many places throughout NailGun, especially `nailgun.entity_mixins` and `nailgun.entities`:

```python
response.raise_for_status()
if (synchronous is True and
        response.status_code == ACCEPTED):
    return _poll_task(response.json()['id'], self._server_config)
if response.status_code == NO_CONTENT:
    return
return response.json()
```

It would be nice if this idiom could be refactored out into a helper function.

Ichimonji10:
I've put together some code that easily refactors out common response handling code as outlined above. However, I wonder if common response handling code is a good design choice. Why is that?

Let's say that a given response should always return an HTTP "NO CONTENT" response. In that case, our response handling code can be:

```python
response.raise_for_status()
return
```

Or, if we'd like to get agressive about error-checking:

```python
response.raise_for_status()
if response.status_code != NO_CONTENT:
    raise APIResponseError(…)
return
```

Similar comments apply to all other responses. We can examine what the server is returning and handle that exact case. The upshot of specialized response-handling code is that all NailGun users will immediately notice if the server's resposnes suddenly change. The response-handling code can serve as an early warning system for breakages that might singal a bug. In general, I like this approach: I think that it is better to make errors obvious than to paper them over.

On the other hand, there are several counter-arguments to specialized response-handling code. First, it is the job of Satellite developers and the broader community to decide how the API should act, and it is the job of Satellite QE and their robottelo test suite to be paranoid about looking for errors in how the API is acting. Does NailGun really need to take on any of those responsibilities? If the server's response changes in an undesireable manner, is that really NailGun's problem? Second, let's follow up on the above example and say that the developers decide to change the server's response from an HTTP "NO CONTENT" to an HTTP "OK". In that case, NailGun will have to respond by changing its response-handling code to this:

```python
response.raise_for_status()
if (getattr(self._server_config, 'version', parse_version('1.2.3')) <
        parse_version('1.2.3')):
    return
else:
    return response.json()
```

That's some nasty-looking code, and it's ripe for weird errors to creep in. Code like this will proliferate through the code base, too. Do you want to maintain that? No? Neither do I. Common response handling code avoids this issue entirely.

Ichimonji10:
For the record, this is what common response handling code could look like:

```python
def _handle_response(response, server_config, synchronous=False):
    response.raise_for_status()
[...]

PR #50, opened 2015-04-28
Fix #4, #46

### LINKED ISSUE #4: Do not populate required fields with random values
By default, calls to method `EntityCreateMixin.create` and several of its siblings causes all required and empty fields to be given random values.

As an example of this, let's say that you do the following:

```python
from nailgun import entities
entities.Repository(name='foo').create()
```

If you do this, then the "content_type", "url" and "product" fields are filled in, as all three of them are marked as required. The "product" field is a foreign key field, so the above code creates a tree of bogus entities. This is great for API testing, but not so much for other users.

Closing this bug requires that the following be done:

* Turn off automatic attribute creation by default.
* Provide some mechanism for enabling automatic attribute creation if so desired by a user.

Ichimonji10:
These two changes are logically separate, but I'm bundling them together for the sake of expediency. I have a matching pull request ready to go on Robottelo, and if this is merged, then I can release 0.12.0 and unblock that PR.

For full details on the changes in this PR, please see the individual commit messages. Summary of changes:

* Do not automatically create missing values on entities. The default assumption among users (especially new users) is likely that if a value is missing, then it will not be created, and this change makes NailGun more accessible to a wide audience by catering to that assumption.
* Refactor out some common response handling code. This makes the existing code base slightly smaller, common response handling code makes silly mistakes harder to commit, and this makes it easier for interested individuals to contribute (the existing code looks more uniform and is therefore easier to study and replicate).

### DIFF
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -85,6 +85,31 @@ class HostCreateMissingError(Exception):
     """Indicates that ``Host.create_missing`` was unable to execute."""
 
 
+def _handle_response(response, server_config, synchronous=False):
+    """Handle a server's response in a typical fashion.
+
+    Do the following:
+
+    1. Check the server's response for an HTTP status code indicating an error.
+    2. Poll the server for a foreman task to complete if an HTTP 202 (accepted)
+       status code is returned and ``synchronous is True``.
+    3. Immediately return if an HTTP "NO CONTENT" response is received.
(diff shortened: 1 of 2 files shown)


omaciel:
ACK

sthirugn:
squash commits?

sthirugn:
ACK

elyezer:
:+1:

PR #59, opened 2015-05-01
Check values assigned to `OneToManyField`s

### LINKED ISSUE #55: Warn users about invalid values for `OneToManyField`s
When instantiating an `Architecture`, a user might write the following:

```python
org = Organization(id=1)
arch = Architecture(
    name='my arch',
    organization=org,
)
```

Or, they might write the following:

```python
arch = Architecture(id=5).read()
arch.organization = org
```

There's a bug in all of the above code, though. Here's the corrected first example:

```python
org = Organization(id=1)
arch = Architecture(
    name='my arch',
    organization=[org],
)
```

And here's the corrected second example:

```python
arch = Architecture(id=5).read()
arch.organization = [org]
```

The above bug will bite the user when they call `create` or its cousins. It will also bite the user when `EntityCreateMixin` is implemented and users call `update` or its cousins. NailGun should be more user-friendly by somehow warning the user about their mistake.

How should this be done? I'm not sure. One approach that _might_ work is to add a set of properties to a class when it is instantiated. For each `OneToManyField`, one property (having the same name as the field) could be added, and if a user assigns a non-iterable value to a `OneToManyField`, then a warning or exception could be raised. That would be nice because the user is alerted right when the questionable assignment is made.

Another approach is to extend `nailgun.entity_mixins.Entity.__init__`. The method already contains the following logic:

```python
# Check that a valid set of field values has been passed in.
if not set(kwargs.keys()).issubset(self._fields.keys()):
    raise NoSuchFieldError(
        'Valid fields are {0}, but received {1} instead.'.format(
            self._fields.keys(), kwargs.keys()
        )
    )
```

That logic could be extended to check and see if a user has provided anything other than an iterable. This approach is nice because it's extremely straightforward. One downside of this approach is that users are not warned when the make assignments to existing objects.

Ichimonji10:
Fix #55. Prevent users from assigning non-iterable values to `OneToManyField`s.
For example:

    >>> from nailgun.entities import Architecture
    >>> Architecture(operatingsystem=[1])
    <nailgun.entities.Architecture object at 0x7f17df759390>
    >>> Architecture(operatingsystem=1)
    File "<stdin>", line 1, in <module>
    File "nailgun/entities.py", line 239, in __init__
        super(Architecture, self).__init__(server_config, **kwargs)
    File "nailgun/entity_mixins.py", line 280, in __init__
        .format(field_name, field_value)
    nailgun.entity_mixins.BadValueError: An inappropriate value was assigned to the "operatingsystem" field. An iterable of entities and/or entity IDs should be assigned, but the following was given: 1

### DIFF
--- a/nailgun/entity_mixins.py
+++ b/nailgun/entity_mixins.py
@@ -1,5 +1,6 @@
 # -*- encoding: utf-8 -*-
 """Defines a set of mixins that provide tools for interacting with entities."""
+from collections import Iterable
 from fauxfactory import gen_choice
 from nailgun import client, config
 from nailgun.entity_fields import IntegerField, OneToManyField, OneToOneField
@@ -135,7 +136,8 @@ def _make_entities_from_ids(entity_cls, entity_objs_and_ids, server_config):
 
     :param entity_cls: An :class:`Entity` subclass.
     :param entity_obj_or_id: An iterable of
-        :class:`nailgun.entity_mixins.Entity` objects and/or entity IDs.
+        :class:`nailgun.entity_mixins.Entity` objects and/or entity IDs. All of
+        the entities in this iterable should be of type ``entity_cls``.
     :returns: A list of ``entity_cls`` objects.
 
     """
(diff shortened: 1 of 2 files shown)


sthirugn:
ACK

omaciel:
ACK

PR #60, opened 2015-05-01
Customize how entity objects are stringified

### LINKED ISSUE #53: Customize how entity objects are stringified
When a [`nailgun.entity_mixins.Entity`](http://nailgun.readthedocs.org/en/latest/api/nailgun.html#nailgun.entity_mixins.Entity) object is stringified, little useful information is provided. The full dotted path to the object's type is provided, in addition to its memory address. However, as is clear from the example below, that's not terribly useful, and it's ugly to boot:

```python
>>> location = Location().create()
>>> location.get_values()
{
    'compute_resource': [],
    'config_template': [<nailgun.entities.ConfigTemplate at 0x107c0be90>,
    <nailgun.entities.ConfigTemplate at 0x107c0b910>,
    <nailgun.entities.ConfigTemplate at 0x107c16050>,
    <nailgun.entities.ConfigTemplate at 0x107c16310>,
    <nailgun.entities.ConfigTemplate at 0x107c165d0>,
    <nailgun.entities.ConfigTemplate at 0x107c16890>,
    <nailgun.entities.ConfigTemplate at 0x107c16b50>,
    <nailgun.entities.ConfigTemplate at 0x107c16e10>,
    <nailgun.entities.ConfigTemplate at 0x107c19110>,
    <nailgun.entities.ConfigTemplate at 0x107c193d0>,
    <nailgun.entities.ConfigTemplate at 0x107c19690>,
    <nailgun.entities.ConfigTemplate at 0x107c19950>,
    <nailgun.entities.ConfigTemplate at 0x107c19c10>],
    'description': None,
    'domain': [],
    'environment': [],
    'hostgroup': [],
    'id': 18,
    'media': [],
    'name': u'\ufc8d\u1c73\uc3a5\u0564\ud17e\u7f1a\u55a2\ub819\u040f\uc88b\u569f\u478c\uadc5\uc8c0\ubeb7\u1565\u13c4\ucd9f\u588d\u4258\u42d3\ub4f5\u82e9\u3153\ub04b',
    'organization': [],
    'smart_proxy': [],
    'subnet': [],
    'user': [],
}
```

The process of stringifying entity objects should be customized. Something like this would be more useful:

    <ConfigTemplate 18>

This could backfire - not all entities have IDs. Perhaps this stringification method could try `self.id`, then `self.uuid`, then fall back to the memory address of the object?

Ichimonji10:
Fix #53:

> When a nailgun.entity_mixins.Entity object is stringified, little useful
> information is provided. The full dotted path to the object's type is
> provided, in addition to its memory address. However, […] that's not terribly
> useful, and it's ugly to boot

Implement and unit test the following methods:

* `nailgun.config.ServerConfig.__repr__`
* `nailgun.entity_mixins.Entity.__repr__`

Example usage (slightly edited):

```python
>>> from nailgun.entities import Product
>>> product = Product(id=1).read()
>>> product
nailgun.entities.Product(
    nailgun.config.ServerConfig(
        url=u'https://sat.example.com';,
        verify=False,
        version=<Version('6.1')>,
        auth=(…)
    ),
    gpg_key=None,
    description=None,
    label=u'Red_Hat_Beta',
    organization=nailgun.entities.Organization(
        nailgun.config.ServerConfig(
            url=u'https://sat.example.com';,
            verify=False,
            version=<Version('6.1')>,
            auth=(…)
        ),
        id=3
    ),
    sync_plan=None,
    id=1,
    name=u'Red Hat Beta'
)
```

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -1,6 +1,7 @@
 """Unit tests for :mod:`nailgun.config`."""
 from mock import call, mock_open, patch
 from nailgun.config import BaseServerConfig, ServerConfig
+from pkg_resources import parse_version
 from unittest import TestCase
 import json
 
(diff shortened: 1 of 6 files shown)


Ichimonji10:
I'm really starting to think that it might be good to also implement `nailgun.config.ServerConfig.__repr__`. It'll mean a bit more work, but the tests should pass without further changes of any sort, and really - server configuration data is an integral part of an entity object. Not including that information in an entity's `__repr__` is disingenuous.

Ichimonji10:
Updated.

elyezer:
```
FAIL: test_identical_args (tests.test_client.ClientTestCase)
Check that the wrapper functions have the correct signatures.
  File "tests/test_client.py", line 135, in test_identical_args
    inspect.getargspec(getattr(requests, meth)),
AssertionError: ArgSpec(args=['url'], varargs=None, keywords='kwargs', defaults=None) != ArgSpec(args=['url', 'params'], varargs=None, keywords='kwargs', defaults=(None,))
```

tests/test_config.py
```diff
@@ -183,6 +184,123 @@ def test_get(self):
...
+        """Test :class:`nailgun.config.BaseServerConfig`.
+
+        Assert that ``__repr__`` works correctly when ``url`` is specified.
+
+        """
+        target = "nailgun.config.BaseServerConfig(url='bogus')"
+        self.assertEqual(target, repr(BaseServerConfig('bogus')))
+        import nailgun  # noqa pylint:disable=unused-variable
+        # pylint:disable=eval-used
+        self.assertEqual(target, repr(eval(repr(BaseServerConfig('bogus')))))
```
elyezer: This is asserting that you can get the representation and just paste it in a python shell for example?

Also I'm interested more on the reason to do this kind of assertion :smile:

elyezer:
ACK, pending comments.

tests/test_config.py
Ichimonji10: > This is asserting that you can get the representation and just paste it in a python shell for example?

Yes. That is exactly correct.

> Also I'm interested more on the reason to do this kind of assertion :smile:

The job of both `__repr__` and `__str__` is to provide a string representation of an object. However, they provide different kinds of string representations. From the docs:

> In short, the goal of `__repr__` is to be unambiguous and `__str__` is to be readable.

`__repr__` should be *so* unambiguous that you can copy-and-paste its output in to a shell and get the same object back. From the docs for [`__repr__`](https://docs.python.org/3/reference/datamodel.html#object.__repr__):

> If at all possible, this should look like a valid Python expression that could be used to recreate an object with the same value (given an appropriate environment).

Here's another example of `__repr__` and `__str__`:

```python
>>> import datetime
>>> today = datetime.datetime.now()
>>> str(today)
'2012-03-14 09:21:58.130922'
>>> repr(today)
'datetime.datetime(2012, 3, 14, 9, 21, 58, 130922)'
```

As you can see, `__repr__` provides an "official" representation of an object that can be used to recreate a new object with the same value.

I've decided to implement `__repr__` instead of `_-str__` because the former is usable in a wider variety of situations and because it will be more useful for debugging. Check this out:

one important distinction between `__str__` and `__repr__` is what happens when a container is stringified. Check this out:

```python
>>> class Foo(object):
...   def __str__(self):
...     return 'Yellow Submarine'
... 
>>> str(Foo())
'Yellow Submarine'
>>> repr(Foo())
'<__main__.Foo object at 0x7f9ba3f86438>'
>>> str([Foo(), Foo()])
'[<__main__.Foo object at 0x7f9ba3f86438>, <__main__.Foo object at 0x7f9ba3f864a8>]'
```

What happens above is that Python calls `__repr__` on all of the elements of a container even when the container itself defines `__str__`.

Ichimonji10:
> Seems that travis does not liked the addition of `__repr__` for `ServerConfig`.

It appears that `inspect.getargspec` has been updated. I've re-installed my own NailGun development environment and am now experiencing the same error. I'll add a second commit to this PR with that fix.

Ichimonji10:
Updated again!

elyezer:
Wasn't able to identify that. Thank you @Ichimonji10 

ACK

sthirugn:
ACK

PR #62, opened 2015-05-05
Properly configure logging system

### LINKED ISSUE #61: Initialize the logging system
When a message is sent to the logging system, that message must be handled some how. A typical method of handling a logging message is to simply print it to stderr, but other handlers might send the message to a file, fire off an email, emit an SMTP message or anything else you can dream up. The logging system is organized in to an arbitrarily complex hierarchy, and messages can be sent to anywhere in this hierarchy. In NailGun's case, the logging hierarchy is quite simple:

    nailgun
    └── client

However, it may be expanded in the future:

    nailgun
    ├── client
    ├── entity_fields
    ├── entity_mixins
    └── entities

When a message is sent to any element in this logging hierarchy, the message is propagated up the hierarchy, and at each level, the message may be used by one or more handlers.

Currently, NailGun has no logging handlers. Thus, when the `nailgun.client` logging element a logging message, that message is handed up to `nailgun` logging element, and it is then discarded. Currently, when a message is logged, Python 2 prints out this warning:

> No handlers could be found for logger "nailgun.client"

And Python 3 prints out a message like this:

> Received HTTP 422 response: {
>   "error": {"id":null,"errors":{"name":["can't be blank"]},"full_messages":["DNS domain can't be blank"]}
> }

Neither of these are ideal.

According to the documentation, the logging system is automatically configured with default values when functions such as `logging.warning` and `logging.debug` are called. However, NailGun does not do that. (Nor does Robottelo!) Instead, _methods_ named `warning`, `debug`, etc. are called.

NailGun's logging system should be initialized so that error messages are properly logged.

Ichimonji10:
Fix #61:

> When a message is sent to the logging system, that message must be handled
> some how. A typical method of handling a logging message is to simply print it
> to stderr, but other handlers might send the message to a file, fire off an
> email, emit an SMTP message or anything else you can dream up.
>
> Currently, NailGun has no logging handlers. Thus, when the nailgun.client
> logging element a logging message, that message is handed up to nailgun
> logging element, and it is then discarded.

Here's what a logging message now looks like:

> WARNING:nailgun.client:Received HTTP 422 response: {
>   "error": {"id":null,"errors":{"name":["can't be blank"]},"full_messages":["DNS domain can't be blank"]}
> }

Also, replace a call to `warn` with a call to `warning`. `warn` is deprecated.

### DIFF
--- a/nailgun/__init__.py
+++ b/nailgun/__init__.py
@@ -15,3 +15,7 @@ As an end user, you'll typically want to use the classes exposed by
 :mod:`nailgun.entities`.
 
 """
+from logging import basicConfig
+
+
+basicConfig()
--- a/nailgun/client.py
+++ b/nailgun/client.py
@@ -89,7 +89,7 @@ def _log_response(response):
         response.text
     )
     if response.status_code >= 400:
-        logger.warn(message)
+        logger.warning(message)
     else:
         logger.debug(message)
 


elyezer:
ACK, pending rebase after #60 being merged in order to make travis green for this PR.

sthirugn:
ACK pending rebase/travis

Ichimonji10:
This PR has been rebased and is passing Travis.

elyezer:
Ship it.

PR #64, opened 2015-05-06
Give generated entities a server config object

### LINKED ISSUE #63: Give generated entities a server config object
When `create_missing` is called on an entity, it fills in all fields that are
required and that do not have values. When filling in a field such as an
`IntegerField` or a `StringField`, the process of creating a bogus value is
straight-forward. When filling in a `OneToOneField` or `OneToManyField`, the
process is more delicate. The correct entity class must be instantiated with an
appropriate `nailgun.config.ServerConfig` object, and all of the dependent
entity's required fields must also be filled in. Currently, method
`nailgun.entity_mixins.EntityCreateMixin.create_missing` does not do that.
Currently, the method instantiates a new entity without passing in a
`ServerConfig` object (causing the new object to search for a server config) and
`create_missing` is not called on the new entity.

Ichimonji10:
Fix #63 and add unit tests for the fix:

> When `create_missing` is called on an entity, it fills in all fields that are
> required and that do not have values. When filling in a field such as an
> `IntegerField` or a `StringField`, the process of creating a bogus value is
> straight-forward. When filling in a `OneToOneField` or `OneToManyField`, the
> process is more delicate. The correct entity class must be instantiated with
> an appropriate `nailgun.config.ServerConfig` object, and all of the dependent
> entity's required fields must also be filled in. Currently, method
> `nailgun.entity_mixins.EntityCreateMixin.create_missing` does not do that.
> Currently, the method instantiates a new entity without passing in a
> `ServerConfig` object (causing the new object to search for a server config)
> and `create_missing` is not called on the new entity.

### DIFF
--- a/tests/test_entity_mixins.py
+++ b/tests/test_entity_mixins.py
@@ -43,8 +43,10 @@ class ManyRelatedEntity(entity_mixins.Entity):
 
 
 class EntityWithDelete(entity_mixins.Entity, entity_mixins.EntityDeleteMixin):
-    """
-    An entity inheriting from :class:`nailgun.entity_mixins.EntityDeleteMixin`.
+    """An entity that can be deleted.
+
+    This entity inherits from :class:`nailgun.entity_mixins.EntityDeleteMixin`.
+
     """
 
     class Meta(object):
@@ -54,8 +56,10 @@ class EntityWithDelete(entity_mixins.Entity, entity_mixins.EntityDeleteMixin):
 
 
 class EntityWithRead(entity_mixins.Entity, entity_mixins.EntityReadMixin):
-    """
-    An entity inheriting from :class:`nailgun.entity_mixins.EntityReadMixin`.
+    """An entity that can be read.
+
+    This entity inherits from :class:`nailgun.entity_mixins.EntityReadMixin`.
+
     """
 
     def __init__(self, server_config=None, **kwargs):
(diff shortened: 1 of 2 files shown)


tests/test_entity_mixins.py
```diff
@@ -458,3 +464,41 @@ def test_create(self):
...
+                    'one_to_many': OneToManyField(SampleEntity, required=True),
+                    'one_to_one': OneToOneField(SampleEntity, required=True),
+                }
+                super(FKEntityWithCreateRead, self).__init__(
+                    server_config,
+                    **kwargs
+                )
+
+        entity = FKEntityWithCreateRead(self.server_config)
+        with mock.patch.object(OneToOneField, 'gen_value') as gen_value_1:
```
elyezer: The with statement can receive more than one context manager like below:

```python
with (
        mock.patch.object(OneToOneField, 'gen_value') as gen_value_1,
        mock.patch.object(OneToManyField, 'gen_value') as gen_value_2):
```

elyezer:
ACK, good catch and fix.

tests/test_entity_mixins.py
Ichimonji10: I get a syntax error under both Python 2.7 and 3.4 when attempting that.

omaciel:
APT

Ichimonji10:
Rebased.

PR #65, opened 2015-05-06
Provide examples of using ServerConfig objects

### LINKED ISSUE #44: Provide examples of using ServerConfig objects
The API documentation briefly mentions how to effectively use `ServerConfig` objects in the docstring for class `Entity`:

> An entity object is useless if you are unable to use it to communicate with a server. The solution is to provide a nailgun.config.ServerConfig when instantiating a new entity. This configuration object is stored as an instance variable named _server_config and used by methods such as nailgun.entity_mixins.Entity.path().
>
> 1. If the server_config argument is specified, then that is used.
> 2. Otherwise, if nailgun.entity_mixins.DEFAULT_SERVER_CONFIG is set, then that is used.
> 3. Otherwise, call nailgun.config.ServerConfig.get().

That's it, though. And the docstring for `ServerConfig.get` is pretty terrible. The existing API documentation should be fixed up, and a nice example of how to effectively use this capability should be provided.

Ichimonji10:
Fix #44:

> The API documentation briefly mentions how to effectively use `ServerConfig`
> objects in the docstring for class `Entity`:
>
> > An entity object is useless if you are unable to use it to communicate with a
> > server. The solution is to provide a nailgun.config.ServerConfig when
> > instantiating a new entity. This configuration object is stored as an instance
> > variable named _server_config and used by methods such as
> > nailgun.entity_mixins.Entity.path().
> >
> > 1. If the server_config argument is specified, then that is used.
> > 2. Otherwise, if nailgun.entity_mixins.DEFAULT_SERVER_CONFIG is set, then that
> > is used.
> > 3. Otherwise, call nailgun.config.ServerConfig.get().
>
> That's it, though. And the docstring for `ServerConfig.get` is pretty terrible.
> The existing API documentation should be fixed up, and a nice example of how to
> effectively use this capability should be provided.

### DIFF
--- a/nailgun/config.py
+++ b/nailgun/config.py
@@ -149,7 +149,7 @@ class BaseServerConfig(object):
 
     @classmethod
     def get(cls, label='default', path=None):
-        """Get a server configuration.
+        """Read a server configuration from a configuration file.
 
         :param label: A string. The configuration identified by ``label`` is
             read.
@@ -276,7 +276,10 @@ class ServerConfig(BaseServerConfig):
 
     @classmethod
     def get(cls, label='default', path=None):
-        """If ``auth`` is a two element list, convert it to a tuple.
+        """Read a server configuration from a configuration file.
+
+        This method extends :meth:`nailgun.config.BaseServerConfig.get`. Please
+        read up on that method before trying to understand this one.
 
         The entity classes rely on the requests library to be a transport
         mechanism. The methods provided by that library, such as ``get`` and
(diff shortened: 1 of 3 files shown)


Ichimonji10:
I think @omaciel may be especially interested in this pull request.

nailgun/config.py
```diff
@@ -285,12 +288,13 @@ def get(cls, label='default', path=None):
...
         However, the JSON decoder does not recognize a tuple as a type, and
-        represents sequences of elements as a tuple. Compensate for that.
+        represents sequences of elements as a tuple. Compensate for that by
+        converting ``auth`` to a two element tuple if it is a two element list.
 
         This override is done here, and not in the base class, because the base
         class may be extracted out into a separate library and used in other
         contexts. In those contexts, the presence of a list may not matter or
-        may be desireable.
+        may be desirable.
```
elyezer: :+1:

elyezer:
ACK

nailgun/config.py
Ichimonji10: `:set spell` :flushed:

omaciel:
ACK and thank you!

PR #66, opened 2015-05-07
Silence spurious insecure connection warnings

Ichimonji10:
From comments in the code itself:

> The urllib3 module (which requests uses) refuses to make insecure HTTPS
> connections. You can override this behaviour by passing `verify=False` to any
> of its methods. For example:
>
>     requests.get('https://insecure.example.com', verify=False)
>
> However, urllib3 will still raise an InsecureRequestWarning if this is done.
> This is problematic. If a user has explicitly stated that they want HTTPS
> connections to be made and we are pestering the user with warnings about that
> fact, then we are effectively training the user to ignore warnings. The
> statement below filters out that warning.

Example of NailGun usage with this patch:

```python
>>> org = entities.Organization(id=1).read()
>>> cfg = ServerConfig.get()
>>> cfg.verify = True
>>> cfg.save()
>>> org = entities.Organization(id=1).read()
…
requests.exceptions.SSLError: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:600)
```

Example of NailGun usage without this patch:

```python
>>> org = entities.Organization(id=1).read()
/home/ichimonji10/.virtualenvs/nailgun3/lib/python3.4/site-packages/requests/packages/urllib3/connectionpool.py:768: InsecureRequestWarning: Unverified HTTPS request is being made. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.org/en/latest/security.html
  InsecureRequestWarning)
```

### DIFF
--- a/nailgun/client.py
+++ b/nailgun/client.py
@@ -16,11 +16,39 @@ Each function is modified with the following behaviours:
 
 """
 from json import dumps
+from warnings import simplefilter
 import logging
 import requests
 
 
 logger = logging.getLogger(__name__)  # pylint:disable=invalid-name
+
+# The urllib3 module (which requests uses) refuses to make insecure HTTPS
+# connections. You can override this behaviour by passing `verify=False` to any
+# of its methods. For example:
+#
+#     requests.get('https://insecure.example.com', verify=False)
+#
+# However, urllib3 will still raise an InsecureRequestWarning if this is done.
+# This is problematic. If a user has explicitly stated that they want HTTPS
+# connections to be made and we are pestering the user with warnings about that
+# fact, then we are effectively training the user to ignore warnings. The
+# statement below filters out that warning.
+#
+# The urllib3 module will still refuse to make HTTPS connections if
+# `verify=True`. In other words, the use of a warning filter does not affect
+# whether urllib3 enforces the security of HTTPS connections.
+#
+# Unfortunately, Python's warnings system is process-wide. Imagine an
+# application which uses both NailGun and some other library, and imagine that
+# the other library generates `InsecureRequestWarning`s. NailGun will filter
+# out the warnings produced by that library, due to the process-wide nature of
+# Python's warnings and filter system. The `warnings.catch_warnings` context
+# manager is not a good solution to this problem, as it is thread-unsafe.
+simplefilter(
+    'ignore',
+    requests.packages.urllib3.exceptions.InsecureRequestWarning,
+)
 
 
 def _content_type_is_json(kwargs):


sthirugn:
nice ACK!

omaciel:
I don't understand how this fix works... Could you explain it to me? Huge +1for getting rid of those pesky warnings though and don't block on me to merge it

Ichimonji10:
@omaciel, warnings are a special type of exception. Check out the [`exception hierarchy`](https://docs.python.org/3.4/library/exceptions.html#exception-hierarchy). As you can see, `Warning` is a sub-class of `Exception`.

I say that warnings are a "special" type of exception because they are handled differently from a normal exception. When a normal exception is raised and goes uncaught, the currently executing statement stops. When a warning exception is raised and goes uncaught, the currently executing statement completes and a message is printed to stderr.

Python allows you to hook in to the warnings system in two ways. First, you can choose to emit warnings in ways other than just printing them to stderr. You can do this with "handlers". For example, I could install a second handler that fires off an email whenever a warning is emitted. In this case, an un-caught warning would cause two things to happen:

* A message is printed to stderr.
* An email is sent.

Handlers can be arbitrarily complex. You could even install a handler that causes your browser to open a new tab and show a [sad panda](https://i.imgur.com/n67E6wm.png) each time an ERROR level error is emitted.

OK, so the first way to hook in to the warnings system is installing new handlers. The second way you can hook in to the warnings system is to filter out warnings and prevent them from being emitted at all. That's what I'm doing here.

Unfortunately, the warnings system is process-wide. This is reflected in the logic of the various methods that the warnings module exposes. For example, you can do this to temporarily filter all warnings:

```python
with warnings.catch_warnings():
    …
```

The above is not as great as it seems, though. What this does is make a copy of the current set of warning filters, installs a new filter that catches all warnings, runs the nested code, and then restores the original set of warning filters. This only behaves well in a single-threaded context. Lets's say that you have a Python application that is using two threads, and the first thread raises a CRITICAL level warning message. Let's also say that this happens while the second thread is using `warnings.catch_warnings`. In this case, the first thread's warning is filtered out by the new filter installed by the second thread, and you never get to see the critical warning. Uh oh. :(

That's a huge weakness of the logic in this pull request. All warnings of type `requests.packages.urllib3.exceptions.InsecureRequestWarning` raised absolutely anywhere in your entire application are now silenced. The reason I haven't submitted this pull request earlier is because of that major flaw. However, I think that training users to ignore warnings is a bigger risk, so I think that this change is worth it.

elyezer:
I think @omaciel's request is accomplished. Time to ship it.

omaciel:
@Ichimonji10 as always thank you for the thorough explanation. This PR is a welcoming addition to the code.

Ichimonji10:
:smile:

PR #69, opened 2015-05-14
Drop `Meta.api_names`

Ichimonji10:
Field names on an entity do not always match what the server expects to receive
in a request or what the server returns in a response. Class attribute
`Meta.api_names` is an attempt to address these mis-matches. It is a dict of
local-remote name mappings. For example, from class `ContentViewFilter`:

```python
class Meta(object):
    """Non-field information about this entity."""
    api_names = {'filter_type': 'type'}
```

Unfortunately, `Meta.api_names` is both complex and limited. It is complex
because cause and effect are distant. The presence of name mappings on
individual classes in module `nailgun.entities` causes logic in several methods
in `nailgun.entity_mixins` to execute — that's not obvious! It is limited
because it cannot handle certain use cases, such as the fact that a name mapping
should be used for the `Subscription` entity, but only when POSTing.

An easier approach is to simply extend or override inherited methods such as
`create_payload` and `read` when needed. For example, the `Media` entity extends
both in order to deal with a naming conflict between an inherited method and a
field.

Test results against an upstream system:

    ...S...S......S...............................................................S...S..........................................................................................................................................SSSS.SSSSSS..S.S.SSSSSS...SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSS
    Ran 278 tests in 469.999s

    OK (SKIP=54)

Note that these test results require the use of a small change to the test
module that will be submitted in a separate PR. (Module `test_multiple_paths` is
the only one in Robottelo that makes use of `Meta.api_names`.) Without the
change, two tests fail.

### DIFF
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -844,7 +844,7 @@ class ContentViewFilter(Entity):
                 ContentView,
                 required=True
             ),
-            'filter_type': entity_fields.StringField(required=True),
+            'type': entity_fields.StringField(required=True),
             'inclusion': entity_fields.BooleanField(),
             'name': entity_fields.StringField(required=True),
             'original_packages': entity_fields.BooleanField(),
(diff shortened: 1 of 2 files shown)


Ichimonji10:
$ git grep -i api_names | wc -l
    0

elyezer:
APT

sthirugn:
ACK

PR #70, opened 2015-05-14
Drop the inner `Meta` class from entities

### LINKED ISSUE #30: Drop the inner Meta classes on entities
The inner Meta classes are arguably a bad idea. A class should contain a bundle of data and logic that operates on that data. However, the Entity classes contain only data. By this reasoning, it is better to drop the Meta classes and add a regular class attribute to hold the same data.

I suggest creating a class attribute named `_meta`. I suggest that this class attribute be private for the following reasons:

* If `Meta`/`_meta` is changed, then *all* entity classes are affected. This is only something that should be done by someone who is an expert with NailGun and who really knows what they're doing.
* Experience with Robottelo shows that access to meta is unnecessary. The only modules that touch the current Meta class are `robottelo.entities` and `tests.foreman.api.test_multiple_path`. The former will soon be ported to NailGun, so those cases of access can be controlled. It is questionable whether the accesses to Meta in the latter case are really necessary.

This is a breaking change because the inner `Meta` class is a public attribute.

Ichimonji10:
Fix #30:

> The inner Meta classes are arguably a bad idea. A class should contain a
> bundle of data and logic that operates on that data. However, the Entity
> classes contain only data. By this reasoning, it is better to drop the Meta
> classes and add a regular class attribute to hold the same data.

A more serious issue is that the use of an inner class brings shared state to
all entity classes. One way in which this is problematic is that is prevents
safe customization of an entity's API path. One class that does this is
`ContentViewPuppetModule`. It has API paths in the following form:

    /content_views/:content_view_id/content_view_puppet_modules
    /content_views/:content_view_id/content_view_puppet_modules/:id

When a `ContentViewPuppetModule` is instantiated, it sets its API path to
something like the following:

    /content_views/42/content_view_puppet_modules

And because an entity's API path is stored as a value on an inner class, this
change is reflected in *all* `ContentViewPuppetModule` objects. That's a bug.

Test results against a downstream system:

    ...S...S......S...............................................................S...S..........................................................................................................................................SSS..S.S.SSSSSS.S..SS.SSSSSSSSSSSSSSSSSSSSSSSSSSSS
    Ran 271 tests in 473.785s

    OK (SKIP=47)

These test results do require a change be made to Robottelo. Robottelo has one
reference to the inner `Meta` class in module `test_multiple_entities`.

### DIFF
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -41,9 +41,6 @@ if version_info.major == 2:
 else:
     from http.client import ACCEPTED, NO_CONTENT  # pylint:disable=import-error
 
-# pylint:disable=too-few-public-methods
-# See NailGun issue #30: https://github.com/SatelliteQE/nailgun/issues/30
-
 # pylint:disable=too-many-lines
 # The size of this file is a direct reflection of the size of Satellite's API.
 # This file's size has already been significantly cut down through the use of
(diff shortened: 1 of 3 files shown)


elyezer:
ACK

sthirugn:
ACK

PR #72, opened 2015-05-15
Unit test overridden read() methods

Ichimonji10:
One thing that the default `read` method does is to pass the current entity
object's server configuration to the entity object being created. Both parent
and child entity objects should talk to the same server with the same
credentials and so on. Unfortunately, several of the `read` method overrides in
module `nailgun.entities` fail to do that. Add a unit test for these methods to
ensure that they do so.

`ContentViewFilterRule.read` calls `keys` on a dict, then iterates through those
keys and modifies the underlying dict. This works on Python 2 where `keys`
returns an entirely new object, but it fails in Python 3 where `keys` returns a
view object:

    $ make test
    …
    File "/home/ichimonji10/code/nailgun/tests/test_entities.py", line 435, in test_sync_plan
        entity.read()
    File "/home/ichimonji10/code/nailgun/nailgun/entities.py", line 842, in read
        for entity_field in entity._fields.keys():
    RuntimeError: dictionary changed size during iteration

Add unit tests for several of the overridden `read` methods and fix these
issues.

### DIFF
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -839,7 +839,7 @@ class ContentViewFilterRule(
         if attrs is None:
             attrs = self.read_json()
         # Field should be present in entity only if it was passed in attributes
-        for entity_field in entity._fields.keys():
+        for entity_field in entity.get_fields().keys():
             if entity_field not in attrs:
                 del entity._fields[entity_field]
         return super(ContentViewFilterRule, self).read(entity, attrs, ignore)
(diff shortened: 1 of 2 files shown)


elyezer:
Is Travis not checking `make test`?

elyezer:
ACK, pending question.

Ichimonji10:
Travis does run `make test`. However, there were no unit tests for that method, so `make test` could not catch that error.

JacobCallahan:
ACK

Ichimonji10:
@elyezer Issues like these could be more effectively prevented if this repo made use of coverage/coveralls.

elyezer:
@Ichimonji10 got it. Also I like the idea of using coverage.

PR #79, opened 2015-05-19
Use method `ServerConfig.get_client_kwargs`

Ichimonji10:
Why use this method? From the method docstring:

> This method returns a dict of attributes that can be unpacked and used as
> kwargs via the ``**`` operator. For example::
>
>     cfg = ServerConfig.get()
>     client.get(cfg.url + '/api/v2', **cfg.get_client_kwargs())
>
> This method is useful because client code may not know which attributes should
> be passed from a ``ServerConfig`` object to one of the ``nailgun.client``
> functions. Consider that the example above could also be written like this::
>
>     cfg = ServerConfig.get()
>     client.get(cfg.url + '/api/v2', auth=cfg.auth, verify=cfg.verify)
>
> But this latter approach is more fragile. It will break if ``cfg`` does not
> have an ``auth`` or ``verify`` attribute.

In other words, using `get_client_kwargs` is more robust than passing individual
`ServerConfig` object attributes to the various functions in `nailgun.client`.

Test results against a downstream system:

    ...S...S......S...............................................................S...S..........................................................................................................................................SSS..S.S.SSSSSS.S..SS.SSSSSSSSSSSSSSSSSSSSSSSSSSSS
    Ran 271 tests in 459.591s

    OK (SKIP=47)

### DIFF
--- a/docs/create_user_nailgun.py
+++ b/docs/create_user_nailgun.py
@@ -46,9 +46,8 @@ def get_organization(server_config, label):
     """
     response = client.get(
         Organization(server_config).path(),
-        auth=server_config.auth,
         data={'search': 'label={}'.format(label)},
-        verify=server_config.verify,
+        **server_config.get_client_kwargs()
     )
     response.raise_for_status()
     decoded = response.json()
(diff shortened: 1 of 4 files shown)


omaciel:
APT

elyezer:
ACK, pending rebase.

Ichimonji10:
Rebased.

elyezer:
APT

elyezer:
Travis is green and coveralls does not look to be running this. Will merge.

PR #83, opened 2015-05-21
Add fauxfactory as an install requirement

### LINKED ISSUE #81: Add python deps via pip install

elyezer:
Nailgun's entities, entity_fields and entity_mixins modules requires
that fauxfactory is installed.

Closes #81

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,6 +1,5 @@
 # For `make test`
 ddt
-fauxfactory
 mock
 
 # For `make lint`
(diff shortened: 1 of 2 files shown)


requirements-dev.txt
```diff
@@ -1,6 +1,5 @@
 # For `make test`
 ddt
-fauxfactory
```
elyezer: It will be installed when NailGun is installed.

elyezer:
Closing this in favor of #82.

PR #84, opened 2015-05-22
Fixed few issues for HostGroup Entity

oshtaier:
I have doubts in current implementation style and appreciate any improvement feedbacks

### DIFF
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -961,8 +961,8 @@ class ContentViewPuppetModule(
 
             entity = type(self)()
 
-        However, :class:`OperatingSystemParameter` requires that an
-        ``operatingsystem`` be provided, so this technique will not work. Do
+        However, :class:`ContentViewPuppetModule` requires that an
+        ``content_view`` be provided, so this technique will not work. Do
         this instead::
 
             entity = type(self)(content_view=self.content_view.id)
@@ -1226,9 +1226,14 @@ class Environment(
 
     def __init__(self, server_config=None, **kwargs):
         self._fields = {
+            'location': entity_fields.OneToManyField(Location, null=True),
             'name': entity_fields.StringField(
                 required=True,
                 str_type='alphanumeric',  # cannot contain whitespace
+            ),
+            'organization': entity_fields.OneToManyField(
+                Organization,
+                null=True,
             ),
         }
         self._meta = {
@@ -1236,6 +1241,15 @@ class Environment(
             'server_modes': ('sat'),
         }
         super(Environment, self).__init__(server_config, **kwargs)
+
+    def create_payload(self):
+        """Wrap submitted data within an extra dict.
+
+        For more information, see `Bugzilla #1151220
+        <https://bugzilla.redhat.com/show_bug.cgi?id=1151220>`_.
+
+        """
+        return {u'environment': super(Environment, self).create_payload()}
 
 
 class Errata(Entity):
@@ -1470,7 +1484,8 @@ class HostGroupClasses(Entity):
         super(HostGroupClasses, self).__init__(server_config, **kwargs)
 
 
-class HostGroup(Entity, EntityCreateMixin):
+class HostGroup(
+        Entity, EntityCreateMixin, EntityDeleteMixin, EntityReadMixin):
     """A representation of a Host Group entity."""
 
     def __init__(self, server_config=None, **kwargs):
@@ -1481,10 +1496,15 @@ class HostGroup(Entity, EntityCreateMixin):
             ),
             'domain': entity_fields.OneToOneField(Domain, null=True),
             'environment': entity_fields.OneToOneField(Environment, null=True),
+            'location': entity_fields.OneToManyField(Location, null=True),
             'medium': entity_fields.OneToOneField(Media, null=True),
             'name': entity_fields.StringField(required=True),
             'operatingsystem': entity_fields.OneToOneField(
                 OperatingSystem,
+                null=True,
+            ),
+            'organization': entity_fields.OneToManyField(
+                Organization,
                 null=True,
             ),
             'parent': entity_fields.OneToOneField(HostGroup, null=True),
@@ -1494,6 +1514,68 @@ class HostGroup(Entity, EntityCreateMixin):
         }
         self._meta = {'api_path': 'api/v2/hostgroups', 'server_modes': ('sat')}
         super(HostGroup, self).__init__(server_config, **kwargs)
+
+    def create_payload(self):
+        """Wrap submitted data within an extra dict.
+
+        For more information, see `Bugzilla #1151220
+        <https://bugzilla.redhat.com/show_bug.cgi?id=1151220>`_.
+
+        """
+        return {u'hostgroup': super(HostGroup, self).create_payload()}
+
+    def read(self, entity=None, attrs=None, ignore=()):
+        """Normalize the data received from the server.
+
+        The server returns data in this format::
+
+            {
+                'id': 15,
+                'name': 'name',
+                'title': 'title',
+                'subnet_id': null,
+                'subnet_name': null,
+                'operatingsystem_id': null,
+                'operatingsystem_name': null,
+                …
+            }
+
+        Normalize data to this format::
+
+            {
+                'id': 15,
+                'name': 'name',
+                'title': 'title',
+                'subnet': {'id': …},
+                'operatingsystem': {'id': …},
+                …
+            }
+
+        """
+        if attrs is None:
+            attrs = self.read_json()
+
+        field_names = [
+            field_name
+            for field_name, field
+            in self.get_fields().items()
+            if isinstance(field, entity_fields.OneToOneField)
+        ]
+        field_names.remove('parent')
+        for field_name in field_names:
+            field_value = attrs.pop(field_name + '_id')
+            if field_value is None:
+                attrs[field_name] = field_value
+            else:
+                attrs[field_name] = {'id': field_value}
+
+        parent_id = attrs.pop('ancestry')
+        if parent_id is None:
+            attrs['parent'] = None
+        else:
+            attrs['parent'] = {'id': parent_id}
+
+        return super(HostGroup, self).read(entity, attrs, ignore)
 
 
 class Host(  # pylint:disable=too-many-instance-attributes


nailgun/entities.py
```diff
@@ -1495,6 +1496,37 @@ def __init__(self, server_config=None, **kwargs):
         self._meta = {'api_path': 'api/v2/hostgroups', 'server_modes': ('sat')}
         super(HostGroup, self).__init__(server_config, **kwargs)
 
+    def read(self, entity=None, attrs=None, ignore=()):
+        if entity is None:
+            entity = type(self)(
+                self._server_config,
+                # pylint:disable=no-member
+                name=self.name,
+            )
```
Ichimonji10: I think this entire block of code can be dropped. `entity` is not used anywhere else in this method, so there's no need to instantiate a `HostGroup` and assign it to `entity`. `super` will do that. Also, imagine the following use case:

```python
from nailgun import entities
host_group = entities.HostGroup(id=5).read()
```

After `HostGroup(id=1)` finishes executing, a `HostGroup` object will be instantiated, and that object will have just one public instance attribute set: `id`:

```python
>>> host_group = entities.HostGroup(id=5)
>>> host_group.get_values()
{'id': 5}
>>> vars(host_group).keys()
dict_keys(['_server_config', '_fields', '_meta', 'id'])
```

When `.read()` is executed, `self.name` is referenced. However, as you can see, no such public instance attribute exists.

Ichimonji10:
The biggest thing I need here is just information on what's being done. I get that we're renaming field names as received from the server, but which field names? Is there only a certain version of Foreman or Satellite that necessitates this fix?

Also, it looks like you're trying to come up with a general-purpose solution where you examine field names and rename them according to a fixed scheme. That's pretty cool, and if your logic can be pushed out in to a private helper function, it could be used all over the place in `nailgun.entities`.

I'd suggest that we start small and:

* Add a docstring explaining what's being done.
* Figure out which server versions require this fix.
* Rename only the necessary fields and only for necessary server versions.

After that, we can looks at writing up a private helper function that works in the general case. By the way, such a private helper function could be *really* cool and help to get rid of a lot of similar-looking code!

oshtaier:
Basically, I have absolutely the same situation like we have for ```auth_source``` parameter for ```User``` class, but for more attributes.
Usually we have next situation for create procedure:
OneToOneField:
(Parameter) default_location ->
(API_create.html) default_location_id ->
(Created_Entity_name) [hash] default_location
OneToManyField:
(Parameter) location ->
(API_create.html) location_ids ->
(Created_Entity_name) [list of hashes] locations

But in our case (which will work just perfect if we change ```auth_source``` type from OneToOneField to StringField, but general creation procedure will be much harder especially if you have a lot of such parameters) we have next picture:

(Parameter) auth_source ->
(API_create.html) auth_source_id ->
(Created_Entity_name) [string] auth_source_id, [string] auth_source_name

elyezer:
Unfortunatelly host group read returns information about related objects inline instead of a separated hash:

```json
{
  "id":15,
  "name":"name",
  "title":"title",
  "subnet_id":null,
  "subnet_name":null,
  "operatingsystem_id":null,
  "operatingsystem_name":null,
  "domain_id":null,
  "domain_name":null,
  "environment_id":null,
  "environment_name":null,
  "compute_profile_id":null,
  "compute_profile_name":null,
  "ancestry":null,
  "puppet_proxy_id":null,
  "puppet_ca_proxy_id":null,
  "ptable_id":null,
  "ptable_name":null,
  "medium_id":null,
  "medium_name":null,
  "architecture_id":null,
  "architecture_name":null,
  "realm_id":null,
  "realm_name":null,
  "created_at":"2015-05-26T13:51:44Z",
  "updated_at":"2015-05-26T13:51:44Z",
  "all_puppetclasses":[],
  "locations":[],
  "organizations":[],
  "parameters":[],
  "template_combinations":[],
  "puppetclasses":[],
  "config_groups":[]
}
```

nailgun/entities.py
```diff
@@ -1463,14 +1473,23 @@ def __init__(self, server_config=None, **kwargs):
...
-        super(HostGroupClasses, self).__init__(server_config, **kwargs)
+
+    def read(self, entity=None, attrs=None, ignore=('hostgroup',)):
+        """Deal with redundant entity fields"""
+        if attrs is None:
+            attrs = self.read_json()
 
 
-class HostGroup(Entity, EntityCreateMixin):
+class HostGroup(
```
elyezer: This fits in one line :smile:

nailgun/entities.py
```diff
@@ -1463,14 +1473,23 @@ def __init__(self, server_config=None, **kwargs):
...
+                # pylint:disable=no-member
+                self.hostgroup.path('self')
+            )
         }
-        super(HostGroupClasses, self).__init__(server_config, **kwargs)
+
+    def read(self, entity=None, attrs=None, ignore=('hostgroup',)):
+        """Deal with redundant entity fields"""
+        if attrs is None:
+            attrs = self.read_json()
```
elyezer: Missing the return on this method.

nailgun/entities.py
```diff
@@ -1359,11 +1360,19 @@ def __init__(self, server_config=None, **kwargs):
...
+                # pylint:disable=no-member
+                self.host.path('self')
+            )
         }
-        super(HostClasses, self).__init__(server_config, **kwargs)
+
+    def read(self, entity=None, attrs=None, ignore=('hostgroup',)):
+        """Deal with redundant entity fields"""
+        if attrs is None:
+            attrs = self.read_json()
```
elyezer: Missing the return on this method.

nailgun/entities.py
```diff
@@ -1348,7 +1348,8 @@ def __init__(self, server_config=None, **kwargs):
         super(GPGKey, self).__init__(server_config, **kwargs)
 
 
-class HostClasses(Entity):
+class HostClasses(
+        Entity, EntityCreateMixin, EntityDeleteMixin, EntityReadMixin):
     """A representation of a Host Class entity."""
 
     def __init__(self, server_config=None, **kwargs):
```
Ichimonji10: Below, I see a reference to `self.host`, but there's no guarantee that `self.host` will exist. How about calling `_check_for_value` here?

```python
    _check_for_value('host', kwargs)
    self._fields = {…}
    super(HostClasses, self).__init__(server_config, **kwargs)
    self._meta = {
        'server_modes': ('sat'),
        # pylint:disable=no-member
        'api_path': '{0}/puppetclass_ids'.format(self.host.path('self'))
    }
```

nailgun/entities.py
Ichimonji10: I think you can drop the assignment to `attrs` completely. The method could be this:

```python
def read(self, entity=None, attrs=None, ignore=('hostgroup',)):
    """Deal with redundant entity fields."""
    return super(HostClasses, self).read(entity, attrs, ignore)
```

nailgun/entities.py
Ichimonji10: I think you can drop the assignment to `attrs` completely. Just call `super`.

nailgun/entities.py
```diff
@@ -1452,7 +1461,8 @@ def create_payload(self):
         return payload
 
 
-class HostGroupClasses(Entity):
+class HostGroupClasses(
+        Entity, EntityCreateMixin, EntityDeleteMixin, EntityReadMixin):
     """A representation of a Host Group Classes entity."""
 
     def __init__(self, server_config=None, **kwargs):
```
Ichimonji10: I see a reference to `self.hostgroup` below. Make a call to `_check_for_value` here, like so:

```python
    _check_for_value('hostgroup', kwargs)
    …
```

Doing this ensures that `self.hostgroup` will be available for use when populating `self._meta['api_path']`.

nailgun/entities.py
```diff
@@ -1579,6 +1632,11 @@ def create_missing(self):
                  |-> domain
                  '-> environment
 
+        In case, you have full set of instance attributes that is necessary to
+        pass, don't forget to put False value for your 'create()' method call,
+        so flow will not enter current 'create_missing' method and not throw
+        an exception
+
```
Ichimonji10: This is inaccurate. [The documentation](http://nailgun.readthedocs.org/en/latest/api/nailgun.entity_mixins.html#nailgun.entity_mixins.EntityCreateMixin) gives this brief description of `create_missing`:

> Populate required fields with random values. Required fields that already have a value are not populated. This method is not called by default.

Calling `.create(False)` is only necessary in Robottelo because Robottelo goes out of its way to change NailGun's default behaviour. [`robottelo/__init__.py`](https://github.com/SatelliteQE/robottelo/blob/master/robottelo/__init__.py) contains a call to [`configure_nailgun()`](https://github.com/SatelliteQE/robottelo/blob/master/robottelo/common/helpers.py#L282). What does that method do? Among other things, it enables automatic dependency creation:

```python
entity_mixins.CREATE_MISSING = True
```

nailgun/entities.py
Ichimonji10: If I can ask - why are we ignoring the `hostgroup` class? Is it because `hostgroup` is also an attribute of the `Host` entity, and therefore also listing the `hostgroup` attribute on `HostClass` is redundant? If that's the case, then we should be able to drop this method entirely (and possibly add something in to the docstring for this class explaining why the hostgroup attribute is ignored). NailGun just ignores extra parameters returned from the server. It only throws an exception when too few parameters are returned.

nailgun/entities.py

nailgun/entities.py
```diff
@@ -1487,7 +1506,7 @@ def __init__(self, server_config=None, **kwargs):
                 OperatingSystem,
                 null=True,
             ),
-            'parent': entity_fields.OneToOneField(HostGroup, null=True),
+            'ancestry': entity_fields.StringField(null=True),
```
Ichimonji10: Does this field reference a `HostGroup` entity? If so, I'm very wary about changing this from a `OneToOneField` to a `StringField`.

nailgun/entities.py
```diff
@@ -1495,6 +1514,40 @@ def __init__(self, server_config=None, **kwargs):
...
+        for attribute_from in attrs.keys():
+            for attribute_to in attribute_to_modify:
+                if attribute_from == attribute_to + '_id':
+                    attribute_value = attrs.pop(attribute_from)
+                    if attribute_value is None:
+                        attrs[attribute_to] = None
+                    else:
+                        attrs[attribute_to] = {'id': attribute_value}
+
+        return super(HostGroup, self).read(entity, attrs, ignore)
```
Ichimonji10: How about something like this?

```python
        """Normalize the data received from the server.

        The server returns data in this format::

            {
                'id': 15,
                'name': 'name',
                'title': 'title',
                'subnet_id': null,
                'subnet_name': null,
                'operatingsystem_id': null,
                'operatingsystem_name': null,
                …
            }

        Normalize data to this format::

            {
                'id': 15,
                'name': 'name',
                'title': 'title',
                'subnet': {'id': …},
                'operatingsystem': {'id': …},
                …
            }

        """
        if attrs is None:
            attrs = self.read_json()

        # Choose fields to fiddle with, then loop through choices.
        field_names = [
            field_name
            for field_name, field
            in self.get_fields().items()
            if isinstance(field, entity_fields.OneToOneField)
        ]
        del field_names['ancestry']
        for field_name in field_names:
            field_value = attrs.pop(field_name + '_id')
            if field_value is None:
                attrs[field_name] = field_value
            else:
                attrs[field_name] = {'id': field_value}

        # `attrs` is now normalized. `super` can do its work.
        return super(HostGroup, self).read(entity, attrs, ignore)
```

nailgun/entities.py
oshtaier: Already
```
class HostGroup(
        Entity, EntityCreateMixin, EntityDeleteMixin, EntityReadMixin):
```

nailgun/entities.py
oshtaier: Completely re-implemented. Thanks for pointing into right direction.
It is best solution that I find after some hours of debugging

nailgun/entities.py
oshtaier: Completely re-implemented. Thanks for pointing into right direction.
It is best solution that I find after some hours of debugging

We get only two attributes: 'hostgroup_id' and 'puppetclass_id', so ignoring correspondingly

nailgun/entities.py
oshtaier: Added. Thanks

nailgun/entities.py
oshtaier: Added. Thanks

nailgun/entities.py
oshtaier: I have no idea what that method is doing. Basically, just wanted to emphasize on thing that took few hours for me to figure out. (to avoid mentioned exception), but you are right, it will be better to do something with that ```configure_nailgun()``` method

nailgun/entities.py
oshtaier: It seems more StringField for me as, for example, "ancestry":"131" doesn't seem as a hash for me.

nailgun/entities.py
oshtaier: Wow. Thanks for such level of details. That approach seems much better

nailgun/entities.py
Ichimonji10: Hashes? I don't understand how hashes are relevant here.

The API documentation for POST and PUT requests says this:

> hostgroup[parent_id] 
> Optional, Nil Allowed.
> Value: number

I then tried making some POST requests, like so:

```python
>>> from nailgun import client, entities
>>> cfg = ServerConfig.get()
>>> path = entities.HostGroup().path()
>>> response = client.post(
...     path,
...     {'hostgroup': {'name': 'hg2', 'parent_id': 9}},
...     **cfg.get_client_kwargs()
... )
>>> response.status_code
200
>>> for field_name in ('parent', 'parent_id', 'parent_ids', 'parents'):                                   
...     field_name in response.json()
... 
False
False
False
False
>>> response.json()['ancestry']
'9'
>>> response = client.post(
...     path,
...     {'hostgroup': {'name': 'hg3', 'parent_id': response.json()['id']}},  # 18
...     **cfg.get_client_kwargs()
... )
>>> response.raise_for_status()
>>> response.json()['ancestry']
'9/18'
>>> response = client.post(
...     path,
...     {'hostgroup': {'name': 'hg4', 'parent_id': response.json()['id']}},  # 19
...     **cfg.get_client_kwargs()
... )
>>> response.raise_for_status()
>>> response.json()['ancestry']
'9/18/19'
```

As you can see, each `HostGroup` entity has a foreign key relationship to some other `HostGroup` entity. In other words, the `HostGroup` database table has a self-referencing foreign key. This foreign key is named "parent". It just so happens that when GETting a `HostGroup`, the IDs of all parents are returned as a string containing decimal IDs joined by forward slashes. (This is utterly stupid IMO. Why not return a list of integer IDs? Or, even better, just the ID of the direct parent?)

What should be done? Keep the "parent" foreign key field:

```python
'parent': entity_fields.OneToOneField(HostGroup, null=True)
```

Then, add some extra logic in to `read` that parses 'ancestry' and uses that information to populate the `parent` field.

nailgun/entities.py
```diff
@@ -1463,14 +1505,39 @@ def __init__(self, server_config=None, **kwargs):
...
+        unique identification value
+
+        """
+        if entity is None:
+            entity = type(self)(
+                self._server_config,
+                hostgroup=self.hostgroup,  # pylint:disable=E1101
+                puppetclass=self.puppetclass,  # pylint:disable=E1101
+            )
+        return super(HostGroupClasses, self).read(entity, attrs, ignore)
```
Ichimonji10: This method will fail in the following case:

```python
>>> entities.HostGroupClasses(id=1, hostgroup=2).read()
```

Ichimonji10: What the the output of `read_json()` look like?

nailgun/entities.py
```diff
@@ -1359,11 +1375,35 @@ def __init__(self, server_config=None, **kwargs):
...
+        unique identification value
+
+        """
+        if entity is None:
+            entity = type(self)(
+                self._server_config,
+                host=self.host,  # pylint:disable=E1101
+                puppetclass=self.puppetclass,  # pylint:disable=E1101
+            )
+        return super(HostClasses, self).read(entity, attrs, ignore)
```
Ichimonji10: This method will fail in the following case:

```python
>>> entities.HostClasses(id=1, host=2).read()
```

What the the output of `read_json()` look like?

nailgun/entities.py
```diff
@@ -961,8 +961,8 @@ def read(self, entity=None, attrs=None, ignore=('content_view',)):
 
             entity = type(self)()
 
-        However, :class:`OperatingSystemParameter` requires that an
-        ``operatingsystem`` be provided, so this technique will not work. Do
+        However, :class:`ContentViewPuppetModule` requires that an
+        ``content_view`` be provided, so this technique will not work. Do
```

nailgun/entities.py
```diff
@@ -1230,6 +1230,11 @@ def __init__(self, server_config=None, **kwargs):
                 required=True,
                 str_type='alphanumeric',  # cannot contain whitespace
             ),
+            'organization': entity_fields.OneToManyField(
+                Organization,
+                null=True,
+            ),
+            'location': entity_fields.OneToManyField(Location, null=True),
```
Ichimonji10: Nitpick: order these fields alphabetically:

```python
'location': entity_fields.OneToManyField(Location, null=True),
'organization': entity_fields.OneToManyFields(
    Organization,
    null=True
)
```

nailgun/entities.py
oshtaier: I think it will not be so obvious for people if we use field name that does not have any correspondence to real attribute.

And when I said 'hash', I mean your comment to the code:
        # The server returns OneToOneFields as a hash of attributes, like so:
        #
        #     {'name': 'Alice Hayworth', 'login': 'ahayworth', 'id': 1}
        #
        # OneToManyFields are a list of hashes.

And actually it usually does that in mentioned way, but okay, I can follow your comment and implement functionality in that way

nailgun/entities.py
oshtaier: not required. I can try to follow that format:
```
puppetclass_id = attrs.pop('puppetclass_id')
        if puppetclass_id is None:
            attrs['puppetclass'] = None
        else:
            attrs['puppetclass'] = {'id': puppetclass_id}
```

nailgun/entities.py
oshtaier: puppetclass_id = attrs.pop('puppetclass_id')
        if puppetclass_id is None:
            attrs['puppetclass'] = None
        else:
            attrs['puppetclass'] = {'id': puppetclass_id}

nailgun/entities.py
Ichimonji10: Ahhhh, I see how hashes are involved here. Sorry about that. :smile:

elyezer:
APT

Ichimonji10:
ACK pending Travis. Don't worry about any drops in coverage percentage - @oshtaier is working super late, and I'll file a follow-up PR that adds any needed tests.

elyezer:
ACK

PR #85, opened 2015-05-22
Update documentation

Ichimonji10:
Currently, all of the modules in the `nailgun` package are documented in a
single reST file. Similarly, all of the modules in the `tests` package are
documented in a single reST file. This technically works, but the compiled
documentation contains some enormously long pages, and the navigation trees are
not intuitive. Fix this by breaking documentation down so there is one reST file
per module. As a result, the compiled documentation contains smaller pages and
more intuitive navigation trees.

Rework the actual documentation contents in several other locations. Provide a
brief explanation of how the methods in each mixin cooperate to do work, provide
a better introduction to the documentation at the root of the API documentation,
and fix a link in the "Examples" documentation.

### DIFF
--- a/docs/api/index.rst
+++ b/docs/api/index.rst
@@ -4,8 +4,10 @@ API Documentation
 This is the NailGun API documentation. It is mostly auto generated from the
 source code. A good place to start reading is :doc:`nailgun`.
 
+The :mod:`nailgun` namespace is the public API. The :mod:`tests` is not part of
+the public API, and it is documented here for easy reference for developers.
+
 .. toctree::
-    :maxdepth: 2
 
     nailgun
     tests
--- a/nailgun/entity_mixins.py
+++ b/nailgun/entity_mixins.py
@@ -374,7 +374,21 @@ class Entity(object):
 
 
 class EntityDeleteMixin(object):
-    """A mixin that adds the ability to delete an entity."""
+    """This mixin provides the ability to delete an entity.
+
+    The methods provided by this class work together. The call tree looks like
+    this::
+
+        delete → delete_raw
+
+    In short, here is what the methods do:
+
+    :meth:`delete_raw`
+        Make an HTTP DELETE request to the server.
+    :meth:`delete`
+        Check the server's response for errors and decode the response.
+
+    """
 
     def delete_raw(self):
         """Delete the current entity.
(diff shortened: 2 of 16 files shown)


elyezer:
APT

docs/api/index.rst
```diff
@@ -4,8 +4,10 @@ API Documentation
 This is the NailGun API documentation. It is mostly auto generated from the
 source code. A good place to start reading is :doc:`nailgun`.
 
+The :mod:`nailgun` namespace is the public API. The :mod:`tests` is not part of
+the public API, and it is documented here for easy reference by developers.
```
omaciel: s/by/for?

nailgun/entity_mixins.py
```diff
@@ -424,7 +438,26 @@ def delete(self, synchronous=True):
...
+
+    In short, here is what the methods do:
+
+    :meth:`read_raw`
+        Make an HTTP GET request to the server.
+    :meth:`read_json`
+        Check the server's response for errors and decode the response.
+    :meth:`read`
+        Create a :class:`nailgun.entity_mixins.Entity` object representing the
+        created entity and populate its fields.
```
omaciel: Do we want to say something like '...populate its fields with the data returned from the server.'?

nailgun/entity_mixins.py
```diff
@@ -530,18 +563,33 @@ def read(self, entity=None, attrs=None, ignore=()):
...
-                └── create_payload
-
-    Only :meth:`create_raw` communicates with the server.
+                ├── create_payload
+                └── create_missing
+
+    In short, here is what the methods do:
+
+    :meth:`create_missing`
+        Populate required fields with bogus values. This method is not called
```
omaciel: s/bogus/random? Bogus to me implies that the values may not make sense (which is true) but the values we provide comply with the expected type at least.

Ichimonji10: :+1:

nailgun/entity_mixins.py
Ichimonji10: :+1:

docs/api/index.rst
Ichimonji10: Sure, will do.

omaciel:
ACK

Issue #12, opened 2014-12-11
Implement EntityUpdateMixin


Ichimonji10:
It'd be super handy to have a standard mechanism for updating existing entities.

Ichimonji10:
Incomplete solution: https://github.com/Ichimonji10/nailgun/compare/master...Ichimonji10:update-mixin

omaciel:
Beautiful! Looking forward to seeing it in action @Ichimonji10

Ichimonji10:
I've hopefully completed work on the `EntityUpdateMixin`. As a bonus, I've revamped and expanded the tests in module `tests.test_entity_mixins`. I'm now adding a second commit to that branch which expands the documentation a bit. What needs to be done before I submitt a PR that will close this issue?

* Finish expanding the documentation.
* Test the final result. Update Robottelo with a few tests that make use of the new `EntityUpdateMixin` and make sure everything's sane.
* Get #106 merged.

Issue #11, opened 2014-12-11
Implement EntitySearchMixin


Ichimonji10:
It'd be super handy to be able to search for entities with a standard search mechanism rather than adding methods to entities on an ad-hoc basis.

Ichimonji10:
Here's a santized version of some notes I wrote up on the subject several weeks ago.

EntitySearchMixin
=================

I think it's about time to create an EntitySearchMixin class. This occurred to me while looking over some Jenkins failures recently. The jenkins failure involved some custom search code, and the process of troubleshooting the search code would be easier if a standard toolset for searching was available. I propose adding an EntitySearchMixin class which provides the following methods:

    search
    └── search_json
        └── search_raw
            └── search_payload

In the near term, the `search` method will not be implemented, as there are some significant barriers to implementing it. However, the other three `search_*` methods can be immediately created.

Search Structure
----------------

When a GET request is issued, a JSON payload must be provided. The payload has this basic structure:

    {
        'filter': '…',
        'filter2': '…',
        'search': 'key=… AND key2=…',
    }

The above is not as easy to implement as it appears at first glance, due to some annoying details. For example, here are some filters named by the API documentation:

* per_page (e.g. 20)
* order (e.g. 'name DESC')
* sort (e.g. {'name', 'DESC'})
* name (e.g. 'ACME Corp')
* organization_id (e.g. 5)

Notice that some of the filters, such as `per_page` and `sort`, are very generic and can apply to all search results, but other filters such as `name` and `organization_id` are specific to the type of entity being searched. Notice also that the `sort` and `order` filters are both very generic, but appear to conflict with each other.

The search string itself is also highly variable. Foreman API paths expect key-value pairs to be joined with the '=' character, whereas Katello API paths expect key-value pairs to be joined with the ':' character. Upstream will unify the syntax over time, but it is not the case right now. Also, search strings are arbitrarily complex. The AND and OR operators may be used, and I suspect that queries can be made more complex by adding parentheses. Also, search terms must be escaped so that characters like '=' do not interfere. The API documentation provides no guidance on how to perform escaping, and some questioning in IRC yielded "my guess is wrap it in double quotes".

Robottelo Search Implementation
-------------------------------

To handle all of the above, here is what I suggest doing:

### Mark Fields

Add a `search_filter=False` argument to `Field.__init__`'s parameter list. This will make it possible to mark individual fields as being filters in a very natural way. For example:

    class Product(…):
[...]

PR #88, opened 2015-05-27
Drop unnecessary entity classes

Ichimonji10:
Drop the `HostClasses` and `HostGroupClasses` entity classes. The first class
exposes the following calls:

* GET    /api/hosts/:host_id/puppetclass_ids
* POST   /api/hosts/:host_id/puppetclass_ids
* DELETE /api/hosts/:host_id/puppetclass_ids/:id

The second class exposes the following calls:

* GET    /api/hostgroups/:hostgroup_id/puppetclass_ids
* POST   /api/hostgroups/:hostgroup_id/puppetclass_ids
* DELETE /api/hostgroups/:hostgroup_id/puppetclass_ids/:id

These calls appear quite useful, but in practice, their presence is confusing an
unnecessary. Consider `GET /api/hosts/:host_id/puppetclass_ids`. It is one of
three highly similar paths:

1. GET /api/puppetclasses
2. GET /api/hosts/:host_id/puppetclasses
3. GET /api/hosts/:host_id/puppetclass_ids

The first call returns all puppet classes by default. You can filter down
exactly how many puppet classes are returned by passing in a JSON payload such
as `{'host_id': 42}`. The second call returns hashes of information about puppet
classes that belong to a host. And the third call returns IDs of puppet classes
that belong to a host.

As you can see, the first API call is the most powerful, the second API call is
less powerful and the third API call is least powerful. As a result, one can
just use the first API call and ignore the other two, and no functionality is
lost.

Similar comments apply to all of the other paths. You can update the set of
puppet classes by making an HTTP POST call to
`/api/hosts/:host_id/puppetclass_ids`. However, it is preferable to make an HTTP
PUT call to `/api/hosts/:id` with a JSON payload of `{'host':
{'puppet_class_ids': [1, 2, 3]}}`. This is preferable not only because you're
using the most general and powerful mechanism available, but also because a PUT
is more appropriate than  POST.

All the above is a nice design discussion, but how will NailGun provide clients
with the ability to do things like get puppet classes belonging to a host or
update the puppet classes belonging to a host? These two actions can be made
possible by writing the `EntitySearchMixin` and `EntityUpdateMixin` as described
in #11 and #12. Statements like the following should be made available:

    # Update the puppet classes belonging to a host.
    host = entities.Host(id=1).read()
    host.puppet_classes.extend([PuppetClass(id=2), PuppetClass(id=3)])
    host.update(fields=['puppet_classes'])

    # Get puppet classes belonging to a host.
    hosts = entities.Host(puppet_classes=[2, 3]).search()


Ichimonji10:
This PR is a direct result of working on #84. It became clear to me that there's some confusion over the design of NailGun. This PR should help reduce confusion both by getting rid of unnecessary classes and by serving as a reference in future discussions.

elyezer:
ACK

omaciel:
ACK pending fixing Travis.

Ichimonji10:
Rebased.

Ichimonji10:
Rebased again.

elyezer:
ACK

PR #91, opened 2015-05-27
Fix AbstractComputeResource inheritance

Ichimonji10:
The entity class `AbstractComputeResource` defines a set of fields that all
inheriting classes use. For example, all compute resources have a name and
provider. Child classes should be able to extend that set of fields. For
example, `LibvirtComputeResource` adds the `display_type` and
`set_console_password` fields.

Currently, that inheritance is broken. The following fails:

    from nailgun.entities import LibvirtComputeResource
    LibvirtComputeResource(display_type='VNC')

Similarly, any field which is defined on a child class cannot be assigned to
directly. Fix this.

### DIFF
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -402,7 +402,20 @@ class AbstractComputeResource(
     """A representation of a Compute Resource entity."""
 
     def __init__(self, server_config=None, **kwargs):
-        self._fields = {
+        # A user may decide to write this if trying to figure out what provider
+        # a compute resource has:
+        #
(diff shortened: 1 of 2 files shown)


Ichimonji10:
Thank you to @omaciel for reporting this bug. If this is merged, I'll cherry-pick the commit and release 0.15.1 ASAP.

elyezer:
ACK

omaciel:
ACK and merge

omaciel:
Thank you ;)

Ichimonji10:
Thank *you*. I've released NailGun 0.15.1 too.

PR #97, opened 2015-06-01
Mark "test-coverage" as a phony make target

Ichimonji10:
Also, rearrange the `.PHONY` list of targets to match the actual ordering of the
targets in the body of the make file.


elyezer:
APT

JacobCallahan:
APT

Ichimonji10:
Rebased.

PR #98, opened 2015-06-01
Add support for RHCI deployment testing

JacobCallahan:
Added in paths for RHCI's satellite-based deployments.
POST fusor/api/v2/deployments
PUT fusor/api/v2/deployments/:id/deploy
DELETE fusor/api/v2/deployments/:id

Added three unit tests for the added paths.

### DIFF
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -2902,6 +2902,99 @@ class Repository(
         return super(Repository, self).read(entity, attrs, ignore)
 
 
+class RHCIDeployment(
+        Entity, EntityCreateMixin, EntityDeleteMixin, EntityReadMixin):
+    """A representation of a RHCI deployment entity."""
+
+    def __init__(self, server_config=None, **kwargs):
+        self._fields = {
+            'name': entity_fields.StringField(required=True),
+            'organization_id': entity_fields.IntegerField(required=True),
+            'lifecycle_environment_id': entity_fields.IntegerField(
+                required=True),
+            'deploy_rhev': entity_fields.BooleanField(required=True),
+            'rhev_engine_host_id': entity_fields.IntegerField(required=True),
+            'rhev_storage_type': entity_fields.StringField(required=True),
+            'rhev_engine_admin_password': entity_fields.StringField(),
+        }
+        self._meta = {
+            'api_path': 'fusor/api/v21/deployments',
+            'server_modes': ('sat'),
+        }
+        super(RHCIDeployment, self).__init__(server_config, **kwargs)
+
+    def path(self, which=None):
+        """Extend ``nailgun.entity_mixins.Entity.path``.
+
+        The format of the returned path depends on the value of ``which``:
+
+        deploy
+            /deployments/<id>/deploy
+
+        ``super`` is called otherwise.
+
+        """
+        if which == 'deploy':
+            return '{0}/{1}'.format(
+                super(RHCIDeployment, self).path(which='self'),
+                which
+            )
+        return super(RHCIDeployment, self).path(which)
+
+    def add_hypervisors(self, hypervisor_ids):
+        """Helper for creating an RHCI deployment.
+
+        :param hypervisor_ids: A list of RHEV hypervisor ids to be added to the
+            deployment.
+        :returns: The server's response, with all JSON decoded.
+        :raises: ``requests.exceptions.HTTPError`` If the server responds with
+            an HTTP 4XX or 5XX message.
+
+        """
+        response = client.put(
+            self.path(),
+            {'discovered_host_ids': hypervisor_ids},
+            **self._server_config.get_client_kwargs()
+        )
+        return _handle_response(response, self._server_config)
+
+    def deploy(self, params):
+        """Kickoff the RHCI deployment.
+
+        :param params: Parameters that are encoded to JSON and passed in
+            with the request. See the API documentation page for a list of
+            parameters and their descriptions.
+        :returns: The server's response, with all JSON decoded.
+        :raises: ``requests.exceptions.HTTPError`` If the server responds with
+            an HTTP 4XX or 5XX message.
+
+        """
+        response = client.put(
+            self.path('deploy'),
+            params,
+            **self._server_config.get_client_kwargs()
+        )
+        return _handle_response(response, self._server_config)
+
+    def update(self, params):
+        """Update the parameters of an existing deployments.
+
+        :param params: Parameters that are encoded to JSON and passed in
+            with the request. See the API documentation page for a list of
+            parameters and their descriptions.
+        :returns: The server's response, with all JSON decoded.
+        :raises: ``requests.exceptions.HTTPError`` If the server responds with
+            an HTTP 4XX or 5XX message.
+
+        """
+        response = client.put(
+            self.path(),
+            params,
+            **self._server_config.get_client_kwargs()
+        )
+        return _handle_response(response, self._server_config)
+
+
 class RoleLDAPGroups(Entity):
     """A representation of a Role LDAP Groups entity."""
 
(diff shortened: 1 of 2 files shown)


JacobCallahan:
Some of this code is subject to change as the project develops and we are able to test it in a more automated manner.

elyezer:
Looks good to me. ACK

Ichimonji10:
I see the following method signatures:

    add_hypervisors(self, hypervisor_ids)
    deploy(self, params)
    update(self, params)

Thanks for adding the explicit argument to the `add_hypervisors()` method. Can you do the same for the `deploy()` method too? I'm not going to bug you about the `update()` method. That's method is going to be blown away eventually, and I'm not sure exactly what its replacement will look like.

JacobCallahan:
@Ichimonji10 right now, the deploy method only accepts one optional parameter, but could be expanded to more in future builds. Also, the update method can be used to update any deployment value, so i'd like to keep that opened up as params for now.

thanks for the help!

nailgun/entities.py
```diff
@@ -2875,6 +2875,99 @@ def read(self, entity=None, attrs=None, ignore=()):
...
+                super(RHCIDeployment, self).path(which='self'),
+                which
+            )
+        return super(RHCIDeployment, self).path(which)
+
+    def add_hypervisors(self, hypervisor_ids):
+        """Helper for creating an RHCI deployment.
+
+        :param hypervisor_ids: A list of RHEV hypervisor ids to be added to the
+        deployment.
```
Ichimonji10: Indenting this should fix (one of?) the `make docs-html` error.

tests/test_entities.py
```diff
@@ -132,6 +134,7 @@ def test_base_path_with_which(self, entity, path, which):
         (entities.Product, 'repository_sets'),
         (entities.Repository, 'sync'),
         (entities.Repository, 'upload_content'),
+        (entities.RHCIDeployment, 'deploy')
```
JacobCallahan: added a ```,``` here to fix a failure.

Ichimonji10: :+1:

nailgun/entities.py
```diff
@@ -2875,6 +2875,99 @@ def read(self, entity=None, attrs=None, ignore=()):
...
+
+        The format of the returned path depends on the value of ``which``:
+
+        deploy
+            /deployments/<id>/deploy
+
+        ``super`` is called otherwise.
+
+        """
+        if which in ('deploy'):
```
Ichimonji10: I believe `make lint` is complaining about the unnecessary parentheses here. `if which == 'deploy':` should satisfy it.

tests/test_entities.py
```diff
@@ -78,6 +78,7 @@ def test_path_without_which(self, entity, path):
         (entities.Product, '/products', 'repository_sets/2396/enable'),
         (entities.Repository, '/repositories', 'sync'),
         (entities.Repository, '/repositories', 'upload_content'),
+        (entities.RHCIDeployment, '/deployments'),
```
Ichimonji10: I think `(entities.RHCIDeployment, '/deployments', 'deploy')` is expected here. (Did I ask you to write something else? Sorry.)

nailgun/entities.py
JacobCallahan: ahhhhhh.... hope that is it!

tests/test_entities.py
JacobCallahan: that makes a bit more sense :)

tests/test_entities.py
```diff
@@ -100,6 +101,7 @@ def test_self_path_with_which(self, entity, path, which):
     @data(
         (entities.ConfigTemplate, '/config_templates', 'build_pxe_default'),
         (entities.ConfigTemplate, '/config_templates', 'revision'),
+        (entities.RHCIDeployment, '/deployments', 'deploy'),
```
Ichimonji10: Unfortunately, this won't work. This test basically does the following:

    path = entities.RHCIDeployment().path()

It then makes some assertions about `path`. That won't work here, given that the "deploy" path is only valid when an ID is provided.

JacobCallahan: hmmmm... want to scrap it altogether?

Ichimonji10: Yep. Just drop this line, and add `(entities.RHCIDeployment, '/deployments')` to [`test_path_without_which`](https://github.com/SatelliteQE/nailgun/blob/master/tests/test_entities.py#L41).

Yes, this is a PITA. But I've been bitten by omitting these tests in the past.

Ichimonji10:
> @Ichimonji10 right now, the deploy method only accepts one optional parameter, but could be expanded to more in future builds. Also, the update method can be used to update any deployment value, so i'd like to keep that opened up as params for now.

OK. Thank you for the feedback. :smile: Figuring out some approach that works across the board is a bit of a nut. We can fall back to just accepting a dict of params, but it'd also be cool if we could offload that work from the end user.

JacobCallahan:
@Ichimonji10 during the next sprint demo, i'll ask the devs if they plan on adding anything to the deploy method. that way we can lock that down. But i don't foresee the update being anything other than a params dict, since there are currently close to 20 parameters that can be updated.

tests/test_entities.py
JacobCallahan: Yeah, RHCI is a bit awkward, but that should be a good compromise.

Ichimonji10:
APT

Ichimonji10:
I'm OK with the coverage decrease. The un-tested methods can be tested by adding in a generic test case that works for any method that simply makes a call to the server and calls `_handle_response()`.

JacobCallahan:
I've looked at the coveralls breakdown and can't seem to find what they're identifying as the lapse in coverage (based on coloring).

Ichimonji10:
Try executing `make test-coverage && coverage html` locally. The output is *much* easier to comprehend than the awful coveralls.io website.

For one thing, the methods you've added are not being tested.

JacobCallahan:
what i'm getting is ```AssertionError: ArgSpec(args=['url', 'params'], varargs=None, keywords='kwargs', defaults=(None,)) != ArgSpec(args=['url'], varargs=None, keywords='kwargs', defaults=None)``` for the test_no_such_path test. is there a way to add the 'params' argument?

Ichimonji10:
You're getting that for `test_no_such_path`?

JacobCallahan:
```
 make test-coverage && coverage html
coverage run --source nailgun -m unittest discover --start-directory tests --top-level-directory .
.........................................................................................................................................F.....
FAIL: test_identical_args (tests.test_client.ClientTestCase)
Check that the wrapper functions have the correct signatures.
  File "/home/jcallaha/Programming/nailgun/tests/test_client.py", line 141, in test_identical_args
    inspect.getargspec(getattr(requests, meth)),
AssertionError: ArgSpec(args=['url', 'params'], varargs=None, keywords='kwargs', defaults=(None,)) != ArgSpec(args=['url'], varargs=None, keywords='kwargs', defaults=None)

Ran 143 tests in 0.083s

FAILED (failures=1)
```

Ichimonji10:
Do a `pip install --upgrade requests`.

JacobCallahan:
that passed!

JacobCallahan:
@Ichimonji10 tests passing locally and in Travis. I'm unsure how to go about closing that -0.03% coverage decrease reported be coveralls, though.

Ichimonji10:
@JacobCallahan There's quite a few methods in NailGun that look just like `deploy()` and `add_hypervisors()`. They just make some request to the server, then hand the response to `_handle_response()`. How would you go about closing that `0.03%` drop in coverage? Come up with some very general unit test that does the following:

1. Mock the `nailgun.client.*` methods and the `_handle_response()` method.
2. Call the method under test.
3. Make some assertions about the call. For example, assert that `_handle_response()` was called once. Or add some extra assertions if the method you have is especially complicated.

That request is really a bit beyond the scope of this PR, and I don't expect you to do what I've outlined. (It would be *very* cool if a separate PR adding that general unit test was added, though!)

JacobCallahan:
Rebased. Will leave PR as-is unless there are any additional changes identified.

Ichimonji10:
Looks good to me.

PR #100, opened 2015-06-01
Fix `create()` method signatures

Ichimonji10:
Method `EntityCreateMixin.create` has the following method signature:

    create(create_missing=None)

Several methods in module `nailgun.entities` extend or override it. Those other
methods should have the same signature.


omaciel:
APT

elyezer:
ACK

PR #104, opened 2015-06-03
Updated missing dependencies in Host.create_missing()

abalakh:
Some entities in Host.create_missing() were created without required
connections through organization and location.
API tests which are using create_missing() are passing without this fix, as
API doesn't check for requirements to be satisfied (so API allows to
associate entities from different locs and orgs, which is not good).
But UI tests are failing as corresponding entities aren't present in
drop-downs and can't be selected.
The fix is required for SatelliteQE/robottelo#748.

### DIFF
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -1670,12 +1670,24 @@ class Host(  # pylint:disable=too-many-instance-attributes
                 'Found instance attributes: {0}'.format(self.get_values())
             )
         super(Host, self).create_missing()
+        # See: https://bugzilla.redhat.com/show_bug.cgi?id=1227854
+        self.name = self.name.lower()
         self.mac = self._fields['mac'].gen_value()
         self.root_pass = self._fields['root_pass'].gen_value()
 
         # Flesh out the dependency graph shown in the docstring.
-        self.domain = Domain(self._server_config).create(True)
-        self.environment = Environment(self._server_config).create(True)
+        self.domain = Domain(
+            self._server_config,
+            # pylint:disable=no-member
+            location=[self.location],
+            organization=[self.organization],
+        ).create(True)
+        self.environment = Environment(
+            self._server_config,
+            # pylint:disable=no-member
+            location=[self.location],
+            organization=[self.organization],
+        ).create(True)
         self.architecture = Architecture(self._server_config).create(True)
         self.ptable = PartitionTable(self._server_config).create(True)
         self.operatingsystem = OperatingSystem(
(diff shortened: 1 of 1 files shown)


nailgun/entities.py
```diff
@@ -1670,12 +1670,21 @@ def create_missing(self):
...
-        self.domain = Domain(self._server_config).create(True)
-        self.environment = Environment(self._server_config).create(True)
+        self.domain = Domain(
+            self._server_config,
+            location=[self.location],
+            organization=[self.organization],
+        ).create(True)
+        self.environment = Environment(
+            self._server_config,
+            location=[self.location],
```
elyezer: I'm wondering why a list is needed here.

nailgun/entities.py
```diff
@@ -1670,12 +1670,21 @@ def create_missing(self):
                 'Found instance attributes: {0}'.format(self.get_values())
             )
         super(Host, self).create_missing()
+        self.name = self.name.lower()
```
elyezer: Why this is needed?

elyezer:
ACK, pending fixing travis and questions.

nailgun/entities.py
abalakh: Host name is created in lowercase. create_missing() is generating name with both lower- and uppercase symbols. If we don't lower it here, than we'll receive ```self.name``` e.g. ```TeStHoSt```, when the actual name of created host is ```testhost```. It causes problem with, for example, search (```hosts.search()``` func). So we either generate name here properly or have to lower received name each time we use the host in our tests.

nailgun/entities.py
Ichimonji10: `location` is a one to many field. This is appropriate.

nailgun/entities.py

nailgun/entities.py
elyezer: :+1: thanks for the clarification.

Ichimonji10:
Travis is failing because the linter thinks `self.location` and `self.organization` are invalid references. But the references you've made to `self.location` and `self.organization` are OK, because the "location" and "organization" fields are required and you're referencing those fields after `super(Host, self).create_missing()` has been called. You can dismiss the warnings by adding `# pylint:disable=no-member` or `# pylint:disable=E1101`.

ACK pending Travis.

nailgun/entities.py
Ichimonji10: Does this indicate that host names should not include upper case characters or that the `hosts.search()` function is buggy?

abalakh: ``search()``` func is not buggy, it's just case-sensitive. You can pass uppercase characters to host name, but those characters will be converted to lowercase after creation anyway, so no point in doing this.

Ichimonji10: Gotcha. Well, that's some annoying server-side behaviour.

/me checks for a BZ about that.

Ichimonji10: Can you link to [BZ 1227854](https://bugzilla.redhat.com/show_bug.cgi?id=1227854), please? Something like this is fine:

```python
        # See: https://bugzilla.redhat.com/show_bug.cgi?id=1227854
        self.name = self.name.lower()
```

abalakh: Sure, done

Ichimonji10:
Thank you. Looks great! APT.

PR #106, opened 2015-06-04
Change dependencies, revamp tests/test_entities.py

Ichimonji10:
Add "unittest2" as a dependency. This package backports many features to older
versions of Python. For example, it makes the `assertRegex` assertion and
`subTest` context manager available in older versions of Python. As a result:

* Drop "ddt" as a dependency.
* Drop some testing logic that makes use of versioning information.

Reorganize `tests/test_entities.py`:

> This file is divided in to three sets of test cases (`TestCase` subclasses):
>
> 1. Tests for inherited methods.
> 2. Tests for entity-specific methods.
> 3. Other tests.

Fiddle with a few methods in module `nailgun.entities` by replacing custom
response handling logic with calls to function
`nailgun.entities._handle_response`. Add some new unit tests for module
`nailgun.entities.`

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,6 +1,6 @@
 # For `make test`
-ddt
 mock
+unittest2
 
 # For `make lint`
 flake8
--- a/tests/test_entities.py
+++ b/tests/test_entities.py
@@ -1,223 +1,23 @@
+# -*- encoding: utf-8 -*-
 """Tests for :mod:`nailgun.entities`."""
 from datetime import datetime
-from ddt import data, ddt, unpack
 from fauxfactory import gen_integer
 from nailgun import client, config, entities
 from nailgun.entity_mixins import EntityReadMixin, NoSuchPathError
-from sys import version_info
-from unittest import TestCase
+from unittest2 import TestCase
 import mock
-# pylint:disable=too-many-public-methods
+# pylint:disable=too-many-lines
+# The size of this file is a direct reflection of the size of module
+# `nailgun.entities` and the Satellite API.
 
 
-@ddt
-class PathTestCase(TestCase):
-    """Tests for extensions of :meth:`nailgun.entity_mixins.Entity.path`."""
-    longMessage = True
-
-    def setUp(self):
-        """Set ``self.server_config`` and ``self.id_``."""
-        self.server_config = config.ServerConfig('http://example.com')
-        self.id_ = gen_integer(min_value=1)
-        if version_info.major == 2:
-            self.re_assertion = self.assertRegexpMatches
-        else:
-            self.re_assertion = self.assertRegex  # pylint:disable=no-member
-
-    @data(
-        (entities.AbstractDockerContainer, '/containers'),
-        (entities.ActivationKey, '/activation_keys'),
-        (entities.ConfigTemplate, '/config_templates'),
-        (entities.ContentView, '/content_views'),
-        (entities.ContentViewVersion, '/content_view_versions'),
-        (entities.ForemanTask, '/tasks'),
-        (entities.Organization, '/organizations'),
-        (entities.Product, '/products'),
-        (entities.Repository, '/repositories'),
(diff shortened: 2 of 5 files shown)


Ichimonji10:
There are a few methods in module `nailgun.entities` that I could easily add some tests for but did not. For example, the `publish` and `promote` methods could use some unit tests, and given the updated structure of the test module, it should be fairly clear how to accomplish that. I'm simply not including those tests in this PR because I've spent *all* of my free time working on this today, and I want to tackle a few other things too.

Ichimonji10:
@JacobCallahan I've added `RHCIDeployment` unit tests for you. :heart:

omaciel:
ACK

requirements-dev.txt
```diff
@@ -1,5 +1,4 @@
 # For `make test`
-ddt
```
Ichimonji10: Incidentally, you can now generate documentation in about 16.5 seconds instead of 17.5 seconds.

tests/test_entities.py
```diff
@@ -1,223 +1,60 @@
...
+#
+# > "A variable used in a closure is defined in a loop. This will result in all
+# > closures using the same value for the closed-over variable."
+#
+# Why is it OK to ignore this warning? It's OK to ignore this warning because
+# the closure being defined (typically named `_assertions()`) is *immediately*
+# called.
+
+
+def _assert(self, assertion_func):
```
Ichimonji10: As a result of the `subTest` context manager, the test failures generated under Python 3.4 will be more rich than the test faulres generated under earlier versions of Python. If checking out some test failures listed on Travis, I recommend looking at the test failures generated under Python 3.4 first.

tests/test_entities.py
```diff
@@ -1,223 +1,60 @@
...
-                entities.System(self.server_config).path(which='base')):
-            self.assertIn('/systems', gen_path)
-            self.re_assertion(gen_path, 'systems$')
+# This file is divided in to three sets of test cases (`TestCase` subclasses):
+#
+# 1. Tests for inherited methods.
+# 2. Tests for entity-specific methods.
+# 3. Other tests.
+#
+# 1. Tests for inherited methods. ---------------------------------------- {{{1
```
Ichimonji10: The curious `{{{1` marker is a manual code folding marker. Several text editors recognize it, such as vim, emacs and jEdit. What I've done here is added in some folding markers that divide the module in to logical sections instead of syntactical sections.

Under vim, try this:

    :set foldmethod=market
    [go to some section of the file]
    zc
    zo

tests/test_entities.py
omaciel: Couldn't find an example on Travis... can you link me or paste an example? I'm curious :)

Ichimonji10: I've made a note to demonstrate this to you tomorrow. Will do!

omaciel: :+1:

Ichimonji10: Gahhh, I failed to get this done. Sorry. I've postponed this to-do until Monday. In the mean-time, have a look at #12. :smile:

elyezer:
Seems that the unittest2 package [1] provides the subtest feature back ported to python 2.

From the change log:

```
2014/10/31 - 0.7.0
This release contains backports from cPython 3.5 of all (I think) commits since 2010, as of today.
```

[1] https://pypi.python.org/pypi/unittest2

Ichimonji10:
Oh no way! If unittest2 supports the `subTest` method, then I'll totally have to rework this and make use of it. Thank you @elyezer!

Ichimonji10:
Thanks for the input, @elyezer. I've revamped this PR according to your suggestions, and the code base is much nicer as a result.

elyezer:
APT

omaciel:
APT x 2

elyezer:
I checked the coveralls report [1] and it have passed two hours ago. Will merge this.

[1] https://coveralls.io/builds/2758799

PR #108, opened 2015-06-08
Add EntityUpdateMixin, update documentation

Ichimonji10:
This pull request includes two closely related but logically separate commits. The first commit adds the `EntityUpdateMixin` and revamps unit tests to increase test coverage. The second commit adds the `EntityupdateMixin` to the `Organization` entity and updates the documentation. See the individual commits for more details.


docs/examples.rst
```diff
@@ -103,8 +103,154 @@ specification is obeyed, meaning that you can do things like provide a
...
+    >>> org.get_values()
+    {
+        'description': 'supercalifragilisticexpialidocious',
+        'id': 418,
+        'label': 'junk_org',
+        'name': 'junkier org',
+        'title': 'junkier org',
+    }
+    >>> org.description = None
+    >>> org = org.update(['description'])  # update only named fields
```
elyezer: Why no using `org.update('field1', 'field2')` instead of `org.update(['field1', 'field2'])`?

nailgun/entity_mixins.py
```diff
@@ -717,3 +749,99 @@ def create(self, create_missing=None):
...
+
+        :param fields: An iterable of field names. Only the fields named in
+            this iterable will be updated. No fields are updated if an empty
+            iterable is passed in. All fields are updated if ``None`` is passed
+            in.
+        :raises: ``KeyError`` if asked to update a field but no value is
+            available for that field on the current entity.
+
+        """
+        return self.read(attrs=self.update_json(fields))
```
elyezer: We must ensure that all entities that uses UpdateMixin have the ReadMixin :smile:

nailgun/entities.py
```diff
@@ -2195,7 +2196,11 @@ def read(self, entity=None, attrs=None, ignore=('operatingsystem',)):
 
 
 class Organization(
-        Entity, EntityCreateMixin, EntityDeleteMixin, EntityReadMixin):
+        Entity,
+        EntityCreateMixin,
+        EntityDeleteMixin,
+        EntityReadMixin,
+        EntityUpdateMixin):
```
elyezer: :tada:

nailgun/entity_mixins.py
```diff
@@ -717,3 +749,99 @@ def create(self, create_missing=None):
...
+        """Create a payload of values that can be sent to the server.
+
+        By default, this method behaves just like :func:`_payload`. However,
+        one can also specify a certain set of fields that should be returned.
+        For more information, see :meth:`update`.
+
+        """
+        values = self.get_values()
+        if fields is not None:
+            values = {field: values[field] for field in fields}
```
elyezer: By using this we clearly says that nailgun does not support python 2.6 and this is cool :)

elyezer:
ACK, pending question.

Thanks for this really nice improvement.

nailgun/entity_mixins.py
Ichimonji10: Right. Travis doesn't test 2.6, and I'm not sure how much work it would take to support 2.6. Speaking of Python versions... 3.5 just came out, and that's worth checking out.

nailgun/entities.py
Ichimonji10: I'm going to submit at least one more PR before releasing 0.17.0. I'm going to update many of the tests in Robottelo to use the `update*` methods, add `EntityUpdateMixin` to many of the entities, and post the test results.

nailgun/entity_mixins.py
Ichimonji10: Yup!

docs/examples.rst
Ichimonji10: I think that a method which requires an iterable of field names is [easier to comprehend](http://docs.python-guide.org/en/latest/writing/style/#explicit-code) and will lead to generally nicer client code. What do I mean by "nicer client code?" Consider the case where I want to update all of an organization's fields except it's name. You could do so like this:

```python
org = …  # org is an Organization object.
org = org.update(set(org.get_values().keys()) - set(['name']))
```

If `*args` were required, the same code would look like this:

```python
org = …  # org is an Organization object.
org = org.update(*(set(org.get_values().keys()) - set(['name'])))
```

elyezer: :+1:

nailgun/entities.py

nailgun/entity_mixins.py
elyezer: Python 2.6 is something that we don't need to worry about.

elyezer:
APT

docs/examples.rst
Ichimonji10: Great question, by the way. You really made me think.

omaciel:
APT

Ichimonji10:
According to coveralls.io, "coverage increased (+1.48%) to 87.07%" for pull request 108 (this PR). This web interface simply is not updating for some reason.

omaciel:
Will merge

Ichimonji10:
Thank you!

omaciel:
Couldn't do it from phone so had to wait to get back home :)

Ichimonji10:
:tada:

PR #117, opened 2015-06-17
Drop extra attribute from `OneTo*Field` classes

### LINKED ISSUE #113: Drop extra attribute from OneTo*Field classes
Here, in full, the definition of the `OneToOneField` class:

```python
class OneToOneField(Field):
    """Field that represents a reference to another entity.

    All parameters not documented here are passed to :class:`Field`.

    :param entity: Either a class or the name of a class.
    :param module: A string. A dotted module name. Used by
        :meth:`OneToOneField.gen_value`. See that method for more details.

    """

    def __init__(self, entity, module=None, *args, **kwargs):
        self.entity = entity
        super(OneToOneField, self).__init__(*args, **kwargs)

    def gen_value(self):
        """Return the class that this field references."""
        return self.entity
```

What's wrong here? The class requires that a `module` argument be provided, and the argument is stored as an instance attribute. The docstring states that `module` should be "a dotted module name" and that it is used by the `gen_value` method. However, that's not true. In reality, the `module` attribute is absolutely useless. It does nothing. The `module` argument and attribute are outdated relics of when the `OneToOneField` was a class attribute. As a result, it should be removed.

Similar comments apply to `OneToManyField`.

Ichimonji10:
Fix #113:

> the `module` attribute is absolutely useless. It does nothing. The `module`
> argument and attribute are outdated relics of when the `OneToOneField` was a
> class attribute. As a result, it should be removed.

### DIFF
--- a/nailgun/entity_fields.py
+++ b/nailgun/entity_fields.py
@@ -237,15 +237,13 @@ class OneToOneField(Field):
 
     All parameters not documented here are passed to :class:`Field`.
 
-    :param entity: Either a class or the name of a class.
-    :param module: A string. A dotted module name. Used by
-        :meth:`OneToOneField.gen_value`. See that method for more details.
+    :param nailgun.entity_mixins.Entity entity: The entity to which this field
+        points.
 
     """
 
-    def __init__(self, entity, module=None, *args, **kwargs):
+    def __init__(self, entity, *args, **kwargs):
         self.entity = entity
-        self.module = module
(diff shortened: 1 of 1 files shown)


Ichimonji10:
$ git grep 'module='
    tests/test_entities.py:            puppet_module=2,

JacobCallahan:
APT

elyezer:
APT

Ichimonji10:
The drop in coverage is unavoidable. :/

PR #122, opened 2015-06-18
Added entities.Product.list_available_rhrepos() func

abalakh:
Added ```entities.Product.list_available_rhrepos()``` func. Need it in tests to verify after ```enable_rhrepo()``` or ```disable_rhrepo()``` repository was actually enabled/disabled, not just by checking response status.
Tests results:
```
make test
python -m unittest discover --start-directory tests --top-level-directory .
.............................................................................................................................................
Ran 141 tests in 0.247s

OK
```

### DIFF
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -2667,6 +2667,23 @@ class Product(
         )
         return _handle_response(response, self._server_config, synchronous)
 
+    # pylint:disable=C0103
+    def repository_sets_available_repositories(self, reposet_id):
+        """Lists available repositories for the repository set
+
+        :param reposet_id: The RepositorySet Id.
+        :returns: Returns list of available repositories for the repository set
+
+        """
+        response = client.get(
+            self.path(
+                'repository_sets/{0}/available_repositories'
+                .format(reposet_id)
+            ),
+            **self._server_config.get_client_kwargs()
+        )
+        return _handle_response(response, self._server_config)['results']
+
 
 class PartitionTable(
         Entity, EntityCreateMixin, EntityDeleteMixin, EntityReadMixin):
--- a/tests/test_entities.py
+++ b/tests/test_entities.py
@@ -230,6 +230,8 @@ class PathTestCase(TestCase):
                 (entities.Product, 'repository_sets'),
                 (entities.Product, 'repository_sets/2396/disable'),
                 (entities.Product, 'repository_sets/2396/enable'),
+                (entities.Product,
+                 'repository_sets/2396/available_repositories'),
                 (entities.Repository, 'sync'),
                 (entities.Repository, 'upload_content'),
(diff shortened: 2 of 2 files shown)


nailgun/entities.py
```diff
@@ -2666,6 +2666,22 @@ def disable_rhrepo(self, base_arch,
         )
         return _handle_response(response, self._server_config, synchronous)
 
+    def list_available_rhrepos(self, reposet_id):
```
Ichimonji10: Can you rename this to `repository_sets_id_available_repositories`? I know that this is a very clunky name. But I'd like to make sure that every extra method we add to NailGun maps directly to some underlying path. I know that we already have several methods such as `enable_rhrepo` and `disable_rhrepo`. But I think that those methods should also be renamed and brought in line with a better naming scheme.

Ichimonji10: Note that at some point in the future I may add a `RepositorySet` entity. At that point, we could replace this method with `RepositorySet.available_repositories`.

Ichimonji10:
Thank you for testing the code you've added. :smile: ACK pending comment.

nailgun/entities.py

elyezer:
ACK, pending @Ichimonji10's comment.

nailgun/entities.py
abalakh: @Ichimonji10 I've spent some time today trying to fit this name with no success. Line
```
:meth:`nailgun.entities.Product.repository_sets_id_available_repositories`
```
from doc doesn't fit in 80 chars. I tried something like this:
```
        """Call :meth:
        `nailgun.entities.Product.repository_sets_id_available_repositories`

        """
```
and this:
```
        """Call :meth:`nailgun.entities.Product.`
        `repository_sets_id_available_repositories`

        """
```
and this:
```
        """Call :meth:|method|.

        .. |method| replace::
           `nailgun.entities.Product.repository_sets_id_available_repositories`

        """
```
and few others but the result is still wrong. Couldn't find the way to break cross-references in sphinx docs either.
Can we use ```repository_sets_available_repositories``` (without ```_id_```) instead? If no, then could you help me to break that line so sphix could still understand it?

Ichimonji10: @abalakh Sure, let's use `repository_sets_available_repositories` instead. That works.

nailgun/entities.py
```diff
@@ -2667,6 +2667,23 @@ def disable_rhrepo(self, base_arch,
         )
         return _handle_response(response, self._server_config, synchronous)
 
+    # pylint:disable=C0103
+    def repository_sets_available_repositories(self, reposet_id):
```
abalakh: lint doesn't like this new name either :smile: , had to disable it here

Ichimonji10:
ACK. Also see #126.

Issue #111, opened 2015-06-13
Add EntityUpdateMixin to all appropriate entities


Ichimonji10:
The EntityUpdateMixin has been written and added to NailGun. Unfortunately, only a smattering of entities inherit from the mixin. All entities that can be updated should inherit from the mixin.

Issue #105, opened 2015-06-04
System entity may have issue with lifecycle environments


Ichimonji10:
See: https://github.com/SatelliteQE/robottelo/pull/2377#discussion_r31722539

elyezer:
:+1:

PR #132, opened 2015-07-01
Add entities.PartitionTable.update()

abalakh:
Added ```entities.PartitionTable.update()```, as it's needed for SatelliteQE/robottelo#2502.
Part of #111.

### DIFF
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -2759,7 +2759,11 @@ class Product(
 
 
 class PartitionTable(
-        Entity, EntityCreateMixin, EntityDeleteMixin, EntityReadMixin):
+        Entity,
+        EntityCreateMixin,
+        EntityDeleteMixin,
+        EntityReadMixin,
+        EntityUpdateMixin):
     """A representation of a Partition Table entity."""
 
     def __init__(self, server_config=None, **kwargs):


nailgun/entities.py
```diff
@@ -2774,6 +2778,17 @@ def __init__(self, server_config=None, **kwargs):
         self._meta = {'api_path': 'api/v2/ptables', 'server_modes': ('sat')}
         super(PartitionTable, self).__init__(server_config, **kwargs)
 
+    def update(self, fields=None):
+        """Fetch a complete set of attributes for this entity."""
```
Ichimonji10: Is there a bugzilla bug stating what's wrong? If not, can we make one? And if so, can we link to it?

My goal is to make sure that every single work-around in NailGun is associated with a bug, even when that means filing trivial bugs or bugs against old versions of Satellite. I'm tracking progress in #105. Feel free to add this method to that list. ([Some](https://bugzilla.redhat.com/show_bug.cgi?id=1230865) [examples](https://bugzilla.redhat.com/show_bug.cgi?id=1234973) [of](https://bugzilla.redhat.com/show_bug.cgi?id=1237257) [bugs](https://bugzilla.redhat.com/show_bug.cgi?id=1235379))

abalakh: Oops, my bad. Double checked - everything works just fine without any workarounds. 
API has confusing behavior sometimes - some entities require extra dict, some - don't, and some are working ok both with and without it.
I'll remove ```update()``` and ```update_payload()```. @Ichimonji10 , should i remove the tests as well? I can't see any tests for entities-with-no-workarounds.

Ichimonji10: > API has confusing behavior sometimes - some entities require extra dict, some - don't, and some are working ok both with and without it.

Yeah. It's a real bugger.

Yes, please do remove the tests from `test_entities.py` as well.

Ichimonji10:
ACK

oshtaier:
ACK

elyezer:
:+1:

PR #133, opened 2015-07-02
File bugs for all workarounds

### LINKED ISSUE #115: File BZs for all workarounds
NailGun contains workarounds for bugs. File bugs for the workarounds in the following methods, and link to the bugs in the method docstrings:

* [x] `ActivationKey.read_raw`
* [x] `Architecture.create_payload`
* [x] `Architecture.update`
* [x] `ComputeResource.create_payload`
* [x] `DockerComputeResource.create`
* [x] `DockerComputeResrouce.read`
* [x] `ConfigTemplate.create_payload`
* [x] `ConfigTemplate.update`
* [x] `AbstractDockerContainer.create_payload`
* [x] `AbstractDockerContainer.create`
* [x] `ContentView.read`
* [x] `ContentViewFilterRule.read`
* [x] `ContentViewPuppetModule.read`
* [x] `ContentViewPuppetModule.create_payload`
* [x] `Domain.create_payload`
* [x] `Domain.create`
* [x] `Domain.read`
* [x] `Domain.update`
* [x] `Environment.create_payload`
* [x] `HostGroup.create_payload`
* [x] `HostGroup.read`
* [x] `HostGroup.update`
* [x] `Host.create_payload`
* [x] `Host.read`
* [x] `Host.update`
* [x] `LifecycleEnvironment.create_payload`
* [x] `Location.create_payload`
* [x] `Location.create`
* [x] `Location.read`
* [x] `Media.create_payload`
* [x] `Media.create`
* [x] `OperatingSystem.create_payload`
* [x] `Organization.create`
* [x] `Organization.read`
* [x] `Organization.update`
* [x] `Organization.update_payload`
* [x] `Product.read`
* [x] `Realm.create`
* [x] `Subnet.create_payload`
* [x] `Subnet.read`
* [x] `System.read`
* [x] `UserGroup.create_payload`
* [x] `UserGroup.read`
* [x] `User.create_payload`
* [x] `User.update_payload`
* [x] `User.update`

Ichimonji10:
Fix #115. File a bug for every workaround (I think). Each commit corresponds to one workaround. See the commit messages for more details.

### DIFF
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -138,6 +138,31 @@ def _check_for_value(field_name, field_values):
         )
 
 
+def _get_org_attrs(server_config, organization_label):
+    """Find an :class:`nailgun.entities.Organization` object.
+
+    :param nailgun.config.ServerConfig server_config: The server that should be
+        searched.
+    :param organization_label: A string. The label of the organization to find.
+    :raises APIResponseError: If exactly one organization is not found.
+    :returns: A dict of attributes about an organization.
+
+    """
+    response = client.get(
+        Organization(server_config).path(),
+        data={u'search': u'label={0}'.format(organization_label)},
+        **server_config.get_client_kwargs()
+    )
+    response.raise_for_status()
+    results = response.json()['results']
+    if len(results) != 1:
+        raise APIResponseError(
+            u'Could not find exactly one organization with label "{0}". '
+            u'Actual search results: {1}'.format(organization_label, results)
+        )
+    return results[0]
+
+
 class ActivationKey(
         Entity,
         EntityCreateMixin,
@@ -1090,6 +1115,23 @@ class ContentView(
             'server_modes': ('sat'),
         }
         super(ContentView, self).__init__(server_config, **kwargs)
+
+    def read(self, entity=None, attrs=None, ignore=()):
+        """Fetch an attribute missing from the server's response.
+
+        For more information, see `Bugzilla #1237257
+        <https://bugzilla.redhat.com/show_bug.cgi?id=1237257>`_.
+
+        """
+        if (getattr(self._server_config, 'version', parse_version('6.1')) <
+                parse_version('6.1')):
+            if attrs is None:
+                attrs = self.read_json()
+            attrs['organization'] = _get_org_attrs(
+                self._server_config,
+                attrs['organization']['label'],
+            )
+        return super(ContentView, self).read(entity, attrs, ignore)
 
     def path(self, which=None):
         """Extend ``nailgun.entity_mixins.Entity.path``.
@@ -2602,28 +2644,20 @@ class Product(
         return super(Product, self).path(which)
 
     def read(self, entity=None, attrs=None, ignore=()):
-        """Compensate for the weird structure of returned data."""
-        if attrs is None:
-            attrs = self.read_json()
+        """Fetch an attribute missing from the server's response.
 
-        # Satellite 6.0 does not include an ID in the `organization` hash.
+        For more information, see `Bugzilla #1237283
+        <https://bugzilla.redhat.com/show_bug.cgi?id=1237283>`_.
+
+        """
         if (getattr(self._server_config, 'version', parse_version('6.1')) <
                 parse_version('6.1')):
-            org_label = attrs.pop('organization')['label']
-            response = client.get(
-                Organization(self._server_config).path(),
-                data={'search': 'label={0}'.format(org_label)},
-                **self._server_config.get_client_kwargs()
+            if attrs is None:
+                attrs = self.read_json()
+            attrs['organization'] = _get_org_attrs(
+                self._server_config,
+                attrs['organization']['label'],
             )
-            response.raise_for_status()
-            results = response.json()['results']
-            if len(results) != 1:
-                raise APIResponseError(
-                    'Could not find exactly one organization with label "{0}".'
-                    ' Actual search results: {1}'.format(org_label, results)
-                )
-            attrs['organization'] = {'id': response.json()['results'][0]['id']}
-
         return super(Product, self).read(entity, attrs, ignore)
 
     def list_repositorysets(self, per_page=None):
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -1043,8 +1043,6 @@ class ContentViewPuppetModule(
 
             entity = type(self)(content_view=self.content_view.id)
 
-        Also, deal with the weirdly named "uuid" parameter.
-
         """
         # read() should not change the state of the object it's called on, but
         # super() alters the attributes of any entity passed in. Creating a new
@@ -1054,9 +1052,6 @@ class ContentViewPuppetModule(
                 self._server_config,
                 content_view=self.content_view,  # pylint:disable=no-member
             )
-        if attrs is None:
-            attrs = self.read_json()
-        attrs['puppet_module_id'] = attrs.pop('uuid')  # either an ID or None
         return super(ContentViewPuppetModule, self).read(entity, attrs, ignore)
 
     def create_payload(self):
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -1905,13 +1905,14 @@ class LifecycleEnvironment(
     def create_payload(self):
         """Rename the payload key "prior_id" to "prior".
 
-        A ``LifecycleEnvironment`` can be associated to another instance of a
-        ``LifecycleEnvironment``. Unusually, this relationship is represented
-        via the ``prior`` field, not ``prior_id``.
+        For more information, see `Bugzilla #1238757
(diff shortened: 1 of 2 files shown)


Ichimonji10:
Test results:

    $ make test-foreman-api-threaded
    python -m cProfile -o test-foreman-api-threaded.pstats $(which nosetests) --logging-filter=nailgun,robottelo --with-xunit --xunit-file=foreman-results.xml tests/foreman/api\
        --processes=-1 --process-timeout=300
    ...SSS.....................S..........E..........E..................E...SSSSSSSSSSSSSSSSSSSSSSS.S.......SSSSSSSSSSSSSSSSSSSSSSSSS.....E........SSSS.SSSSSSSSS............S.SSSSSSS.SSSSSS................................................S.........SSSSSSSSSSSSS...S...S...S...SSSSSSSSS..........S.......SSSSSSSS.....S...S......S...........................................................................S.S....................................................................................S...............................................................SSSSSS.SSSSSSSS.S..SSS.S...SSSSSSSSSSSSSSSSSSSSSSSSSSSS.S..................S...S.....................................................E...S..E..................EE..................EEEEEEE..EE...............SSSSSS.SS..SSSSSSSSSSSSSSSSSSSSSSSSSSSSSS..SEESS.SSSSSSSSSSSSSSSSSSSSSS.S.............................................
    Ran 867 tests in 1375.692s

    FAILED (SKIP=236, errors=19)

Every error apepars to be due to either legitimate problems on the server (a couple HTTP 500s), problems that arise from running threaded tests (database deadlocks) or broken RHCI tests (to be expected).

Ichimonji10:
For full details, see internal paste 294553.

Ichimonji10:
bump

tests/test_entities.py
```diff
@@ -1338,6 +1351,59 @@ def setUpClass(cls):
         cls.cfg_608 = config.ServerConfig('bogus url', version='6.0.8')
         cls.cfg_610 = config.ServerConfig('bogus url', version='6.1.0')
 
+    def test_missing_org_id(self):
+        """Test methods for whom no organization ID is returned.
```
omaciel: s/whom/which?

omaciel:
ACK

tests/test_entities.py
Ichimonji10: Good catch.

tests/test_entities.py
```diff
@@ -1338,6 +1351,59 @@ def setUpClass(cls):
...
+        Affected methods:
+
+        * :meth:`nailgun.entities.ContentView.read`
+        * :meth:`nailgun.entities.Product.read`
+
+        Assert that ``read_json``, ``_get_org_attrs`` and ``read`` are all
+        called once, and that the second is called with the correct arguments.
+
+        """
+        for klass in (entities.ContentView, entities.Product):
```
JacobCallahan: why "klass"?

JacobCallahan:
ACK, pending minor question.

nailgun/entities.py
```diff
@@ -138,6 +138,31 @@ def _check_for_value(field_name, field_values):
         )
 
 
+def _get_org_attrs(server_config, organization_label):
```
oshtaier: Is it specific reason to have such 'getter' only for Organizations? Or is it first one in a list of such methods?

JacobCallahan: i believe it is for the content view tests in robottelo. basically, it will help to get the organization's attributes from what is returned when creating a content view, if i am remembering it correctly.

nailgun/entities.py
```diff
@@ -1096,6 +1135,23 @@ def __init__(self, server_config=None, **kwargs):
...
         super(ContentView, self).__init__(server_config, **kwargs)
 
+    def read(self, entity=None, attrs=None, ignore=()):
+        """Fetch an attribute missing from the server's response.
+
+        For more information, see `Bugzilla #1237257
+        <https://bugzilla.redhat.com/show_bug.cgi?id=1237257>`_.
+
+        """
+        if (getattr(self._server_config, 'version', parse_version('6.1')) <
```
oshtaier: Could you please put more explanation for that statement?

oshtaier:
ACK pending questions

tests/test_entities.py
Ichimonji10: `class` is a reserved word. I could use `entity` if you prefer.

nailgun/entities.py
Ichimonji10: Ugh. The handling of version information in NailGun is one thing I am *not* happy with. `getattr` is used because `self._server_config.version` is not guaranteed to exist.

This statement can also be written as such:

```python
version = getattr(self._server_config, 'version', parse_version('6.1'))
if version < parse_version('6.1'):
    pass  # some workaround
```

This code can be read as, "if we're talking to a Satellite system older than 6.1, perform a workaround".

nailgun/entities.py
Ichimonji10: This method is not the first one in a list of such methods. This should only exist until the entity search mixin comes along and renders this redundant.

This method exists because there are several unique but highly similar bugs. Here's one of them:

> One can read a content view by issuing an HTTP GET request to `/katello/api/v2/products/:id`. This request returns JSON data in this form:
>
>     {
>         'id': …,
>         'organization' {'label': '…', 'name': '…'},
>         …
>     }
>
> Notice how the "organization" dict does not include an ID. This makes it impossible to know which organization a content view is associated with without performing a search against the server.

nailgun/entities.py
```diff
@@ -932,10 +967,14 @@ def read(self, entity=None, attrs=None, ignore=('content_view_filter',)):
             )
         if attrs is None:
             attrs = self.read_json()
-        # Field should be present in entity only if it was passed in attributes
-        for entity_field in entity.get_fields().keys():
-            if entity_field not in attrs:
-                del entity._fields[entity_field]
+        ignore = set(ignore).union([
```
Ichimonji10: The handling of the `ignore` argument is another things I don't like about NailGun. See #134.

nailgun/entities.py
Ichimonji10: It would be better if this sort of comparison could be made:

```python
from packaing.version import parse
...
if self._server_config.version < parse('6.1'):
    pass  # some workaround
```

However, this will only work if `version` is guaranteed to exist on `self._server_config`. Making that guarantee is tricky to get right. One option is to simply require that the user provide a version any time they make a `ServerConfig` object:

```python
cfg = ServerConfig('example.com', version='6.1')
```

This is straightforward, but does break backward compatibility and will render many saved server configs unusable. Another option is to provide a default value.

```python
>>> from packaging.version import Version
>>> cfg = ServerConfig('example.com')
>>> cfg.version == Version('1!0')  # epoch 1, release 0.0.0
True
```

The default version would be set to "epoch 1, release 0.0.0", which would *always* be greater than any version in epoch 0:

```python
>>> from packaging.version import Version
>>> Version('1!0') > Version('0.0.0')
True
>>> Version('1!0') > Version('0.0.1')
True
>>> Version('1!0') > Version('0.1.0')
True
>>> Version('1!0') > Version('1.0.0')
True
>>> Version('1!0') > Version('6.0.8')
True
```

...well, so long as Satellite never changes its versioning scheme:

```python
>>> from packaging.version import Version
>>> Version('1!0') > Version('1!0.0.1')
False
```

tests/test_entities.py
JacobCallahan: that looks better :)

Ichimonji10: @JacobCallahan done.

nailgun/entities.py
oshtaier: And we need that comparison to apply some workaround, right? Did we have different API in version before 6.1?

Ichimonji10: Yes, the API changes with each new release of Satellite. Sometimes bugs are fixed with a new release, so we conditionally execute a workaround only for an older release. Sometimes an entity's fields change, so we conditionally add those new fields only for a new version. And so on.

oshtaier: Huh. It can be a mess if we have 3-5 versions supported at one moment of time.
Anyway :+1: for explanation

Ichimonji10: Yeah. It definitely can be a mess. And the `nailgun.entities` module is huge and will continue to trend toward growth in the forseeable future. I can see a couple possible solution. One solution is to start making use of API specifications in the same way that python-foreman does. If nothing else, those specifications can be used to automatically populate fields. Another solution is to do something wild like make (all of? only some of?) the classes in `nailgun.entities` empty shells that just look at which version is being instantiated and delegate the instantiation process out to some other class. This isn't as outlandish as it seems:

```python
>>> from packaging.version import parse
>>> type(parse('1.0.0'))
<class 'packaging.version.Version'>
>>> type(parse('invalid version'))
<class 'packaging.version.LegacyVersion'>
```

PR #139, opened 2015-07-09
Deduplicate version specifiers

### LINKED ISSUE #17: Single Version Number
There should be a single, authoritative source for the current version of NailGun. The version number <del>will be</del> <ins>is</ins> declared in the following locations <del>once the commits in the "bootstrap" branch are merged in to master</del>:

* `setup.py`
* `docs/conf.py`

That's not great. It's easy to update one and forget about the other. The ["Single-sourcing the version across `setup.py` and your project"](https://packaging.python.org/en/latest/development.html#single-sourcing-the-version-across-setup-py-and-your-project) section of the Python Packaging User Guide provides some suggestions.

Ichimonji10:
Fix #17:

> There should be a single, authoritative source for the current version of
> NailGun. The version number is declared in the following locations:
>
> * `setup.py`
> * `docs/conf.py`
>
> That's not great. It's easy to update one and forget about the other.

Create a `VERSION` file in NailGun's root directory and read it as necessary.
Add a `MANIFEST.in` file and use it to include the `VERSION` file in packages.
See:
https://docs.python.org/3.4/distutils/sourcedist.html#specifying-the-files-to-distribute


setup.py
```diff
@@ -7,9 +14,13 @@
     LONG_DESCRIPTION = handle.read()
 
 
+with open('VERSION') as handle:
+    VERSION = handle.read().strip()
```
Ichimonji10: It's important to strip whatever's returned by `read()` as a precaution. Let's say I create a file like so:

    echo 0.18.0 > foo

There's a trailing newline in the file:

```python
>>> with open('foo') as handle:
...     handle.read()
... 
'0.18.0\n'
```

setup.py
```diff
@@ -1,5 +1,12 @@
 #!/usr/bin/env python
-"""A setuptools-based script for installing NailGun."""
+"""A setuptools-based script for installing NailGun.
+
+For more information, see:
+
+* https://packaging.python.org/en/latest/index.html
+* https://docs.python.org/distutils/sourcedist.html
```
Ichimonji10: This link explains the `MANIFEST.in` file. The Python 2.7 and 3.4 docs are significantly different, but they both cover the same material.

Ichimonji10:
I'll admit that this is not the most important issue in the issue tracker. <_< On the up side, I've learned about the `MANIFEST.in` file and some other details of Python packaging.

elyezer:
ACK, thank you for sharing what you've learned.

omaciel:
ACK

PR #140, opened 2015-07-09
Move a pylint directive

Ichimonji10:
Place a `pylint:disable` directive at the end of a line. Doing so causes pylint
to ignore the named warning for that one line, rather than for all following
code at the same indentation level — in this case, almost a thousand lines of
code.

### DIFF
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -2803,8 +2803,7 @@ class Product(
         )
         return _handle_response(response, self._server_config, synchronous)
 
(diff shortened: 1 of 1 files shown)


Ichimonji10:
Ran across this while investigating an unrelated issue.

omaciel:
ACK

omaciel:
@elyezer bump?

elyezer:
ACK

PR #143, opened 2015-07-14
Rearrange documentation

Ichimonji10:
Add brief examples of how to use NailGun to the beginning of the documentation.
This should serve as a simple reference that will help new-comers on-board.

Fiddle with existing documentation. Make the calls to `pprint` more concise and
add a few explanatory comments.

### DIFF
--- a/docs/create_organization_nailgun.py
+++ b/docs/create_organization_nailgun.py
@@ -7,7 +7,7 @@ Use NailGun to accomplish this task.
 """
 from nailgun.config import ServerConfig
 from nailgun.entities import Organization
-from pprint import PrettyPrinter
+from pprint import pprint
 
 
 def main():
@@ -17,7 +17,7 @@ def main():
         url='https://sat1.example.com',  # …to talk to this server.
     )
     org = Organization(server_config, name='junk org').create()
-    PrettyPrinter().pprint(org.get_values())
+    pprint(org.get_values())  # e.g. {'name': 'junk org', …}
     org.delete()
 
 
--- a/docs/create_organization_nailgun_v2.py
+++ b/docs/create_organization_nailgun_v2.py
@@ -2,13 +2,13 @@
 # -*- coding: utf-8 -*-
 """Create an organization, print out its attributes and delete it."""
 from nailgun.entities import Organization
-from pprint import PrettyPrinter
+from pprint import pprint
 
 
 def main():
     """Create an organization, print out its attributes and delete it."""
     org = Organization(name='junk org').create()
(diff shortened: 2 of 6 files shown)


Ichimonji10:
@kbidarkar

kbidarkar:
@Ichimonji10 , Nice to have that Quickstart section , this would definitely help.

a) I think we also add "verify=False" to that example. [ required]

Below stuff can help, even further. I leave it to you to decide if you want to add. [ optional]
b) A mention of save feature and where the config file  get's created would also  be informative. ~/.config/nailgun/server_configs.json , may not be necessary but would help to get an end-end idea that we can read it again.
c) And to read it back again, cfg = ServerConfig.get('sat61a')

Ichimonji10:
I've updated the example as per your request.

I've added a mention of the the `save()`/`get()`/etc methods like so:

    >>> server_config.save()  # Save to disk w/label 'default'. Read with get()

I want to make this example as brief as possible and give users the minimum necessary set of tools to get started. Simply saving and reading back the "default" configuration should be sufficient. Users can learn about the ability to work with multiple labels, the `get_labels` and `delete` methods, and other functionality as needed. The mention of "w/label 'default'" should hint that there's more to learn.

I can add a section mentioning SSL like so:

    ...     verify=False,  # Ignore SSL errors. DANGEROUS AND INADVISABLE.

That said, I must express a *strong* dislike for mentioning this feature in the top-level README. Ignoring SSL verification really is **dangerous and inadvisable**. When someone abuses the SSL system, the onus should be on them to make things work - we should not be teaching users that it is acceptable to undercut security technologies for convenience. Furthermore, it is likely that many people will simply copy and paste the provided examples and use the examples with little-to-no modification in their own code. Adding `verify=False` in to our examples means that other users will end up with insecure code - and they will rightly be able to blame us for providing bad advice.

JacobCallahan:
Mentioning the verify part would be good, but I definitely agree with taking it out of the code examples. 
ACK, pending that.

Ichimonji10:
I'm OK with mentioning the ability in a comment. Would something like this work?

    >>> server_config = ServerConfig(
    ...     auth=('admin', 'changeme'),      # Use these credentials…
    ...     url='https://sat1.example.com',  # …to talk to this server.
    ... )  # More options available, e.g. disabling SSL verification.

Would that work?

kbidarkar:
agreed.

Ichimonji10:
Updated.

JacobCallahan:
Re-ACK

elyezer:
:+1:

PR #145, opened 2015-07-15
Change dependencies

### LINKED ISSUE #144: Use "packaging" as a dependency
The [packaging](https://packaging.pypa.io/en/latest/) library a set of "core utilities for Python packages." One of the things it provides is a set of tools for parsing and comparing versions. The package is provided by the Python Packaging Authority.

The "setuptools" package also provides the ability to parse and compare versions. However, a look at the source code shows that setuptools simply depends on packaging and wraps what packaging provides.

packaging is a more targeted and specific tool. NailGun only uses the versioning capabilities from setuptools, and it seems reasonable for NailGun to make use of a more targeted dependency.

This is a breaking change for the following reasons:

* We're changing dependencies.
* Clients that hook tightly in to NailGun may make use of `pkg_resources.parse_version` and `pkg_resources.SetuptoolsVersion`. (Not sure why, but hey.)

Ichimonji10:
Fix #144. Drop "setuptools" as a dependency and use "packaging" instead.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -7,7 +7,7 @@ http://sphinx-doc.org/config.html
 import sys
 import os
 
-from pkg_resources import SetuptoolsVersion
+from packaging.version import Version
 
 
 # Add the NailGun root directory to the system path. This allows references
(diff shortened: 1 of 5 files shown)


Ichimonji10:
Robottelo does not reference pkg_resources anywhere, so this PR should not affect Robottelo in any way.

Anyone who is using NailGun directly instead of installing via PyPi will need to do `pip install -r requirements.txt`. @kbidarkar @sthirugn

oshtaier:
ACK

JacobCallahan:
ACK

elyezer:
:+1:

PR #149, opened 2015-07-16
Fix for a random ptable.name bug

abalakh:
Got tired of random test [failures](https://github.com/SatelliteQE/robottelo/issues/2211) [because](https://github.com/SatelliteQE/robottelo/issues/2319) [of](https://github.com/SatelliteQE/robottelo/issues/2318) [partition](https://github.com/SatelliteQE/robottelo/issues/2291) [tables](https://github.com/SatelliteQE/robottelo/issues/2229) [bug (BZ1229384)](https://bugzilla.redhat.com/show_bug.cgi?id=1229384).
This one-line PR should get rid of it.

**Pros**:
+ Tests will stop failing
+ Bug is pretty annoying, as failures are random (only when randint(1,30) throws 1), short failure log contains just common '422' error. As a result it's always time-consuming to detect the issue for each new build. 
+ We're not decreasing coverage, as there already [is](https://github.com/SatelliteQE/robottelo/blob/master/tests/foreman/api/test_partitiontable.py#L30) a test to verify that specific case
+ [BZ](https://bugzilla.redhat.com/show_bug.cgi?id=1229384) is already moved out of this release, so without the fix we will see random failures for a long time.

**Cons**:
- It's not a good approach to avoid existing bugs. If bug exists - test should fail.
- After bug is fixed, this PR should be manually reverted, there's no ```skip_if_bug_open``` func for nailgun :(

Please, share your thoughts on this.

### DIFF
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -2841,12 +2841,18 @@ class PartitionTable(
         EntityDeleteMixin,
         EntityReadMixin,
         EntityUpdateMixin):
-    """A representation of a Partition Table entity."""
+    """A representation of a Partition Table entity.
+
+    Currently a Partition Table with one character in name cannot be created.
+    For more information, see `Bugzilla #1229384
+    <https://bugzilla.redhat.com/show_bug.cgi?id=1229384>`_.
+
+    """
 
     def __init__(self, server_config=None, **kwargs):
         self._fields = {
             'layout': entity_fields.StringField(required=True),
-            'name': entity_fields.StringField(required=True),
+            'name': entity_fields.StringField(required=True, length=(2, 30)),
             'os_family': entity_fields.StringField(
                 choices=_OPERATING_SYSTEMS,
                 null=True,


Ichimonji10:
> It's not a good approach to avoid existing bugs. If bug exists - test should fail.

We can add a test in to Robottelo where we explicitly create a partition table with a single-character name.

> After bug is fixed, this PR should be manually reverted, there's no skip_if_bug_open func for nailgun :(

NailGun is capable of acting differently based on the server version. So we can set the minimum length to 2 characters for Satellite 6.1.0 and earlier and set the minimum length to 1 character for 6.1.1 (?) and beyond.

This is actually a fine PR. Let's:

* Make sure there's a test for this specific bug in Robottelo.
* Reference the BZ from this class's docstring.

Ichimonji10:
BTW, the TODO statement causes linting to fail. Go ahead and drop that line of code and reference the bug from the class docstring.

abalakh:
> We can add a test in to Robottelo where we explicitly create a partition table with a single-character name.

I've already added this test - [test_bugzilla_1229384](https://github.com/SatelliteQE/robottelo/blob/master/tests/foreman/api/test_partitiontable.py#L30)

sthirugn:
APT - Ty for taking care of this.

sthirugn:
ACK

oshtaier:
ACK. Pending @Ichimonji10 comments, especially that part:
'So we can set the minimum length to 2 characters for Satellite 6.1.0 and earlier and set the minimum length to 1 character for 6.1.1 (?) and beyond.'
We should have that fix for 6.3.0, so it is even more critical as we will definitely forget about our discussion after so long period of time

Ichimonji10:
> > We can add a test in to Robottelo where we explicitly create a partition table with a single-character name.
>
> I've already added this test - [test_bugzilla_1229384](https://github.com/SatelliteQE/robottelo/blob/master/tests/foreman/api/test_partitiontable.py#L30)

:smile: Fantastic.

> We should have that fix for 6.3.0, so it is even more critical as we will definitely forget about our discussion after such long period of time

There's a bunch of bug references and bug workarounds in nailgun. Currently, there's no automated way to use or bypass a bug workaround. All of that logic is hard-coded in. It may be feasible to contact the BZ and redmine servers and inspect the state of bugs at runtime, though there's some risks inherent in doing that.

ACK

PR #153, opened 2015-07-21
Fixes an api unicode issue for partition table names.

sthirugn:
- Previously the partition table name length was defined to be a minimum of 2 characters.  But sometimes when the name is decoded from bytes to a logical character, the name of 2 characters length becomes one character long which causes the failure again.  To fix this, the partition table name minimum length will be 4 characters now.

### DIFF
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -2859,12 +2859,15 @@ class PartitionTable(
     For more information, see `Bugzilla #1229384
     <https://bugzilla.redhat.com/show_bug.cgi?id=1229384>`_.
 
+    Note: Having a name length of 2 had failures again.  Updating the length to
+    4.
+
     """
 
     def __init__(self, server_config=None, **kwargs):
         self._fields = {
             'layout': entity_fields.StringField(required=True),
-            'name': entity_fields.StringField(required=True, length=(2, 30)),
+            'name': entity_fields.StringField(required=True, length=(4, 30)),
             'os_family': entity_fields.StringField(
                 choices=_OPERATING_SYSTEMS,
                 null=True,


sthirugn:
Unit test:
```sh
Before fix: (note it produces 2 character length names)
>>> for i in range(100):
...   len((PartitionTable().create()).name)
... 
12
28
16
7
12
27
19
5
13
18
2
13
19
25
2
21
10
12
26
20
24
2
2
7
15
3
10
14
10
24
3
11
15
20
28
3
12
11
28
11
15
13
13
19
3
26
19
25
18
4
11
24
3
15
8
28
12
11
21
15
11
21
12
21
20
24
2
5
2
8
2
20
9
18
8
9
27
15
19
9
22
20
29
7
2
24
29
21
27
20
7
28
13
4
29
7
20
23
12
14

# After fix: (note it doesn't produce 2 character length names)
>>> for i in range(100):
...   len((PartitionTable().create()).name)
... 
16
21
15
7
28
7
6
5
25
5
30
21
14
30
25
26
18
19
11
5
15
12
7
15
12
11
19
7
29
8
15
18
5
18
7
4
30
26
10
4
29
19
18
11
6
23
5
25
18
10
9
13
21
13
17
12
9
18
14
24
28
25
27
4
12
11
8
15
30
22
10
16
15
15
7
28
26
14
23
16
27
8
28
30
5
20
28
27
25
20
5
24
25
24
27
19
5
7
11
9
```

Ichimonji10:
ACK

Some unicode characters require two (16-bit) code points to be represented. These characters consist of a pair of characters: a high surrogate and a low surrogate. I strongly suspect that Satellite is correctly decoding these code point pairs as a single character, which is triggering the "whitespace" bug. We can work around it by making sure that we always generate strings at least four code points long, which is what this PR does.

sthirugn:
Merging with one ACK for the next automation run.

PR #156, opened 2015-07-22
Make `read`'s `ignore` argument easier to manipulate

### LINKED ISSUE #134: Make `read`'s `ignore` argument easier to manipulate
The `read` method has this signature:

```python
```

The `ignore` argument is a tuple of field names that NailGun should not try to populate when reading data from the server. For example, this is the signature of `DockerComputeResource.read`:

```python
def read(self, entity=None, attrs=None, ignore=('password',)):
```

This method signature works well if one wants to ignore a certain method attribute at all times. However, this method signature does not work well if one wants to conditionally ignore some field due to a bug. For example, this bit of code is used in response to a specific bug:

```python
ignore = set(ignore).union([…])  # list comprehension
if 'errata_id' in attrs:
    ignore.discard('errata')  # pylint:disable=no-member
ignore = tuple(ignore)
```

In the above code snippet, `ignore` is converted to a set, manipulated, then converted back to a tuple. Unwieldy idioms like this are likely to become more prevalent as more field-specific bugs are discovered - especially when those bugs are fixed in certain Satellite versions.

An empty tuple is used as the default argument to `read` because:

* tuples are immutable
* the field should default to an empty iterable

However, as described above, tuples are hard to work with. Some better solution should be conceived and implemented.

One alternative it to make `read` have this signature:

```python
def read(self, entity=None, attrs=None, ignore=None):
```

And somewhere in the call stack, `ignore` would be given a default value:

```python
if ignore is None:
    ignore = set()
```

This would work decently well in cases where an argument should always be ignored:

```python
def read(self, entity=None, attrs=None, ignore=None):
    if ignore is None:
        ignore = set()
    ignore.add('password')
```

It would also work very well in cases where more interesting manipulations need to happen:

```python
ignore.add([…])  # list comprehension
if 'errata_id' in attrs:
    ignore.discard('errata')  # pylint:disable=no-member
```

This solution would retain the two advantages listed earlier:

* <del>tuples are</del><ins>None is</ins> immutable
* the field should default to an empty iterable

abalakh:
As per #134 discussion, together with @oshtaier made `read`'s `ignore` argument to be instance of set instead of tuple. Default value for `ignore` changed to `None`.
This will retain the two advantages:
- None is immutable
- Field should default to an empty iterable
And introduce a new one:
- Sets are much easier to manipulate, e.g. no need to use type castings to add/remove some elements

Closes #134

`[DO NOT MERGE]` prefix till i get test run results and validate everything's OK (~ 1 hour)

### DIFF
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -381,7 +381,7 @@ class AuthSourceLDAP(
                 if not hasattr(self, field):
                     setattr(self, field, self._fields[field].gen_value())
 
-    def read(self, entity=None, attrs=None, ignore=('account_password',)):
+    def read(self, entity=None, attrs=None, ignore=None):
         """Do not read the ``account_password`` attribute. Work around a bug.
 
         For more information, see `Bugzilla #1243036
@@ -390,6 +390,9 @@ class AuthSourceLDAP(
         """
         if attrs is None:
             attrs = self.update_json([])
+        if ignore is None:
+            ignore = set()
+        ignore.add('account_password')
         return super(AuthSourceLDAP, self).read(entity, attrs, ignore)
 
 
@@ -550,7 +553,7 @@ class DockerComputeResource(AbstractComputeResource):
             id=self.create_json(create_missing)['id'],
         ).read()
 
-    def read(self, entity=None, attrs=None, ignore=('password',)):
+    def read(self, entity=None, attrs=None, ignore=None):
         """Do extra work to fetch a complete set of attributes for this entity.
 
         For more information, see `Bugzilla #1223540
@@ -562,6 +565,9 @@ class DockerComputeResource(AbstractComputeResource):
         """
         if attrs is None:
             attrs = self.read_json()
+        if ignore is None:
+            ignore = set()
+        ignore.add('password')
         if 'email' not in attrs and 'email' not in ignore:
             response = client.put(
                 self.path('self'),
@@ -963,7 +969,7 @@ class ContentViewFilterRule(
             )
         }
 
-    def read(self, entity=None, attrs=None, ignore=('content_view_filter',)):
+    def read(self, entity=None, attrs=None, ignore=None):
         """Do not read certain fields.
 
         Do not expect the server to return the ``content_view_filter``
@@ -984,14 +990,16 @@ class ContentViewFilterRule(
             )
         if attrs is None:
             attrs = self.read_json()
-        ignore = set(ignore).union([
+        if ignore is None:
+            ignore = set()
+        ignore.add('content_view_filter')
+        ignore.update([
             field_name
             for field_name in entity.get_fields().keys()
             if field_name not in attrs
         ])
         if 'errata_id' in attrs:
             ignore.discard('errata')  # pylint:disable=no-member
-        ignore = tuple(ignore)
         return super(ContentViewFilterRule, self).read(entity, attrs, ignore)
 
 
@@ -1089,7 +1097,7 @@ class ContentViewPuppetModule(
             )
         }
 
-    def read(self, entity=None, attrs=None, ignore=('content_view',)):
+    def read(self, entity=None, attrs=None, ignore=None):
         """Provide a default value for ``entity``.
 
         By default, ``nailgun.entity_mixins.EntityReadMixin.read provides a
@@ -1112,6 +1120,10 @@ class ContentViewPuppetModule(
                 self._server_config,
                 content_view=self.content_view,  # pylint:disable=no-member
             )
+
+        if ignore is None:
+            ignore = set()
+        ignore.add('content_view')
         return super(ContentViewPuppetModule, self).read(entity, attrs, ignore)
 
     def create_payload(self):
@@ -1156,7 +1168,7 @@ class ContentView(
         }
         super(ContentView, self).__init__(server_config, **kwargs)
 
-    def read(self, entity=None, attrs=None, ignore=()):
+    def read(self, entity=None, attrs=None, ignore=None):
         """Fetch an attribute missing from the server's response.
 
         For more information, see `Bugzilla #1237257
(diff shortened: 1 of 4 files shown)


oshtaier:
@JacobCallahan please review and let us know if you have any suggestions. We decided to look on the task first and ping you if we need any help:)

abalakh:
Run without changes:
```
Ran 1082 tests in 5353.651s

FAILED (SKIP=186, errors=10)
```
Run after changes:
```
Ran 1082 tests in 5370.366s

FAILED (SKIP=186, errors=9)
```
So i guess things are not bad :)
Won't remove `[DO NOT MERGE]` as I still have to work on unit tests.

nailgun/entities.py
```diff
@@ -389,6 +389,8 @@ def read(self, entity=None, attrs=None, ignore=('account_password',)):
         """
         if attrs is None:
             attrs = self.update_json([])
+        if ignore is None:
+            ignore = {'account_password'}
```
JacobCallahan: if we were to pass in ```ignore={'something_else'}```, then would we no longer be ignoring ```'account_password'```?
If so, then that would not be desirable. In that case, would something like this be better?
```
if ignore is None:
    ignore = set()
ignore.add('account_password')
```

abalakh: :facepalm:
Sure. Will update :)

JacobCallahan: :+1:

nailgun/entities.py
```diff
@@ -2052,7 +2067,7 @@ def create(self, create_missing=None):
         attrs = self.create_json(create_missing)
         return Location(self._server_config, id=attrs['id']).read()
 
-    def read(self, entity=None, attrs=None, ignore=('realm',)):
+    def read(self, entity=None, attrs=None, ignore=None):
         """Work around a bug in the server's response.
 
         Do not try to read any of the attributes listed in the ``ignore``
```
Ichimonji10: > Do not try to read the ``realm`` attribute.

nailgun/entities.py
```diff
@@ -2499,7 +2520,7 @@ def create(self, create_missing=None):
             id=self.create_json(create_missing)['id'],
         ).read()
 
-    def read(self, entity=None, attrs=None, ignore=('realm',)):
+    def read(self, entity=None, attrs=None, ignore=None):
         """Fetch as many attributes as possible for this entity.
 
         The server does not return any of the attributes listed in the
```
Ichimonji10: > Do not try to read the ``realm`` attribute.

nailgun/entities.py
```diff
@@ -3147,7 +3171,7 @@ def __init__(self, server_config=None, **kwargs):
         }
         super(RHCIDeployment, self).__init__(server_config, **kwargs)
 
-    def read(self, entity=None, attrs=None, ignore=('rhev_engine_host',)):
+    def read(self, entity=None, attrs=None, ignore=None):
         """Normalize the data returned by the server.
 
         The server's JSON response is in this form::
```
Ichimonji10: > The response does not contain a value for the ``rhev_engine_host`` field.

nailgun/entities.py
```diff
@@ -3395,7 +3422,7 @@ def create_payload(self):
         """
         return {u'subnet': super(Subnet, self).create_payload()}
 
-    def read(self, entity=None, attrs=None, ignore=('discovery',)):
+    def read(self, entity=None, attrs=None, ignore=None):
         """Fetch as many attributes as possible for this entity.
 
         The server does not return any of the attributes listed in the
```
Ichimonji10: Do not try to read the ``discovery`` attribute.

nailgun/entities.py
```diff
@@ -3649,11 +3683,7 @@ def path(self, which=None):
...
 
-    def read(
-            self,
-            entity=None,
-            attrs=None,
-            ignore=('facts', 'organization', 'type')):
+    def read(self, entity=None, attrs=None, ignore=None):
         """Fetch as many attributes as possible for this entity.
 
         The server does not return any of the attributes listed in the
```
Ichimonji10: > Do not try to read the ``facts``, ``organization`` or ``type`` attributes.

nailgun/entities.py
```diff
@@ -3816,8 +3851,11 @@ def create_payload(self):
         """
         return {u'user': super(User, self).create_payload()}
 
-    def read(self, entity=None, attrs=None, ignore=('password',)):
+    def read(self, entity=None, attrs=None, ignore=None):
         """Do not read any attributes listed in the ``ignore`` argument."""
```
Ichimonji10: > Do not read the ``password`` argument.

Ichimonji10:
Thank you! ACK pending some easy nitpicks.

JacobCallahan:
@Ichimonji10 for clarification, what do you mean by "Do not read the _____ argument"?

oshtaier:
@JacobCallahan Probably update the docstring in a proper way

Ichimonji10:
@JacobCallahan The docstrings currently contain lines like this:

> Do not try to read any of the attributes listed in the `ignore` argument. For more information, see Bugzilla...

The `ignore` argument has changed, so those docstrings are no longer accurate. I'm just asking for the docstrings to be updated.

JacobCallahan:
Ahh the docstrings! I thought you were talking about the code itself. Hence the confusion.

ACK pending @Ichimonji10's input.

Ichimonji10:
@JacobCallahan I was unclear. :disappointed:

JacobCallahan:
re-ACK

Ichimonji10:
ACK

@abalakh May this be merged? I see the "DO NOT MERGE" banner still.

oshtaier:
One moment guys. We want to add few more unit tests. Will update PR shortly

Ichimonji10:
OK. Thanks for the udpate @oshtaier. :smile:

JacobCallahan:
:+1:

abalakh:
Done! Added `test_ignore_arg_v4()` and `def test_read_v4()` & rebased the commit. Now it can be merged (if no new comments, of course) :smile:

Ichimonji10:
Re-re-ACK. :smile:

JacobCallahan:
final ACK!

PR #157, opened 2015-07-23
EntitySearchMixin

Ichimonji10:
The first commit implements and tests class `EntitySearchMixin`. The second commit integrates the mixin with several entity classes. See the individual commit messages for more details.

The current NailGun version is 0.18.2. I'd like to release version 0.19.0 today, then include this change in 0.20.0. I am marking this PR as "DO NOT MERGE" for that reason, but am posting the PR now to collect feedback and to keep track of it.

For proof that this PR functions correctly, see <del>FIXME: Fetch information from the future</del> <ins>SatelliteQE/robottelo#2610</ins>.

### DIFF
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -29,6 +29,7 @@ from nailgun.entity_mixins import (
     EntityCreateMixin,
     EntityDeleteMixin,
     EntityReadMixin,
+    EntitySearchMixin,
     EntityUpdateMixin,
     _poll_task,
 )
@@ -46,7 +47,8 @@ else:  # pragma: no cover
 # The size of this file is a direct reflection of the size of Satellite's API.
 # This file's size has already been significantly cut down through the use of
 # mixins and fields, and cutting the file down in size further would simply
-# obfuscate the design of the entities.
+# obfuscate the design of the entities. It might be possible to place entity
+# definitions in separate modules, though.
 
 # pylint:disable=attribute-defined-outside-init
 # NailGun aims to be like a traditional database ORM and allow uses of the dot
@@ -56,9 +58,14 @@ else:  # pragma: no cover
 #     product.name
 #     product.organization.id
 #
-# Unfortunately, these fields cannot simply be initialized with `None`, as the
-# server considers "nil" to be different from the absence of a value. This
-# inevitably means that instance attributes will be defined outside __init__.
+# Unfortunately, these fields cannot simply be initialized with `None`. These
+# have different effects:
+#
+#     product.description = None; product.update()
+#     del product.description; product.update()
+#
+# The first will delete the product's description, and the second will not
+# touch the product's description.
 
 
 _FAKE_YUM_REPO = 'http://inecas.fedorapeople.org/fakerepos/zoo3/'
@@ -116,7 +123,7 @@ def _check_for_value(field_name, field_values):
     example, in :class:`nailgun.entities.ContentViewPuppetModule`:
 
     >>> def __init__(self, server_config=None, **kwargs):
-    >>>     _check_for_param('content_view', kwargs)
+    >>>     _check_for_value('content_view', kwargs)
     >>>     # …
     >>>     self._meta = {
     >>>         'api_path': '{0}/content_view_puppet_modules'.format(
@@ -138,29 +145,25 @@ def _check_for_value(field_name, field_values):
         )
 
 
-def _get_org_attrs(server_config, organization_label):
+def _get_org(server_config, label):
     """Find an :class:`nailgun.entities.Organization` object.
 
     :param nailgun.config.ServerConfig server_config: The server that should be
         searched.
-    :param organization_label: A string. The label of the organization to find.
+    :param label: A string. The label of the organization to find.
     :raises APIResponseError: If exactly one organization is not found.
-    :returns: A dict of attributes about an organization.
+    :returns: An :class:`nailgun.entities.Organization` object.
 
     """
-    response = client.get(
-        Organization(server_config).path(),
-        data={u'search': u'label={0}'.format(organization_label)},
-        **server_config.get_client_kwargs()
+    organizations = Organization(server_config).search(
+        query={u'search': u'label={0}'.format(label)}
     )
-    response.raise_for_status()
-    results = response.json()['results']
-    if len(results) != 1:
+    if len(organizations) != 1:
         raise APIResponseError(
             u'Could not find exactly one organization with label "{0}". '
-            u'Actual search results: {1}'.format(organization_label, results)
+            u'Actual search results: {1}'.format(label, organizations)
         )
-    return results[0]
+    return organizations[0].read()
 
 
 class ActivationKey(
@@ -1161,10 +1164,8 @@ class ContentView(
                 parse('6.1')):
             if attrs is None:
                 attrs = self.read_json()
-            attrs['organization'] = _get_org_attrs(
-                self._server_config,
-                attrs['organization']['label'],
-            )
+            org = _get_org(self._server_config, attrs['organization']['label'])
+            attrs['organization'] = org.get_values()
         return super(ContentView, self).read(entity, attrs, ignore)
 
     def path(self, which=None):
@@ -1377,7 +1378,11 @@ class Domain(
 
 
 class Environment(
-        Entity, EntityCreateMixin, EntityDeleteMixin, EntityReadMixin):
+        Entity,
+        EntityCreateMixin,
+        EntityDeleteMixin,
+        EntityReadMixin,
+        EntitySearchMixin):
     """A representation of a Environment entity."""
 
     def __init__(self, server_config=None, **kwargs):
@@ -1492,7 +1497,11 @@ class ForemanTask(Entity, EntityReadMixin):
 
 
 class GPGKey(
-        Entity, EntityCreateMixin, EntityDeleteMixin, EntityReadMixin):
+        Entity,
+        EntityCreateMixin,
+        EntityDeleteMixin,
+        EntityReadMixin,
+        EntitySearchMixin):
     """A representation of a GPG Key entity."""
 
     def __init__(self, server_config=None, **kwargs):
@@ -1871,7 +1880,11 @@ class Interface(Entity):
 
 
 class LifecycleEnvironment(
-        Entity, EntityCreateMixin, EntityDeleteMixin, EntityReadMixin):
+        Entity,
+        EntityCreateMixin,
+        EntityDeleteMixin,
+        EntityReadMixin,
+        EntitySearchMixin):
     """A representation of a Lifecycle Environment entity."""
 
     def __init__(self, server_config=None, **kwargs):
@@ -1929,27 +1942,14 @@ class LifecycleEnvironment(
         super(LifecycleEnvironment, self).create_missing()
         if (self.name != 'Library' and  # pylint:disable=no-member
                 not hasattr(self, 'prior')):
-            response = client.get(
-                self.path('base'),
-                data={
-                    u'name': u'Library',
-                    # pylint:disable=no-member
-                    u'organization_id': self.organization.id,
-                },
-                **self._server_config.get_client_kwargs()
-            )
-            response.raise_for_status()
-            results = response.json()['results']
+            results = self.search({'organization'}, {u'name': u'Library'})
             if len(results) != 1:
                 raise APIResponseError(
-                    'Could not find the "Library" lifecycle environment for '
-                    'organization {0}. Search results: {1}'
+                    u'Could not find the "Library" lifecycle environment for '
+                    u'organization {0}. Search results: {1}'
                     .format(self.organization, results)  # pylint:disable=E1101
                 )
-            self.prior = LifecycleEnvironment(
-                self._server_config,
-                id=results[0]['id'],
-            )
+            self.prior = results[0]
 
 
 class Location(
@@ -1957,6 +1957,7 @@ class Location(
         EntityCreateMixin,
         EntityDeleteMixin,
         EntityReadMixin,
+        EntitySearchMixin,
         EntityUpdateMixin):
     """A representation of a Location entity."""
 
@@ -2243,6 +2244,7 @@ class Organization(
         EntityCreateMixin,
         EntityDeleteMixin,
         EntityReadMixin,
+        EntitySearchMixin,
         EntityUpdateMixin):
     """A representation of an Organization entity."""
 
@@ -2518,7 +2520,7 @@ class OverrideValue(Entity):
         super(OverrideValue, self).__init__(server_config, **kwargs)
 
 
-class Permission(Entity, EntityReadMixin):
+class Permission(Entity, EntityReadMixin, EntitySearchMixin):
     """A representation of a Permission entity."""
 
     def __init__(self, server_config=None, **kwargs):
@@ -2532,47 +2534,8 @@ class Permission(Entity, EntityReadMixin):
         }
         super(Permission, self).__init__(server_config, **kwargs)
 
-    def search(self, per_page=10000):
-        """Searches for permissions using the values for instance name and
-        resource_type
 
-        Example usage::
-
-            >>> from nailgun import entities
-            >>> entities.Permission(resource_type='User').search()
-            [
-                {'name': 'view_users', 'resource_type': 'User', 'id': 158},
-                {'name': 'create_users', 'resource_type': 'User', 'id': 159},
-                {'name': 'edit_users', 'resource_type': 'User', 'id': 160},
-                {'name': 'destroy_users', 'resource_type': 'User', 'id': 161},
-            ]
-            >>> entities.Permission(name='create_users').search()
-            [{'name': 'create_users', 'resource_type': 'User', 'id': 159}]
-
-        If both ``name`` and ``resource_type`` are provided, ``name`` is
-        ignored.
-
-        :param per_page: number of results per page to return
-        :returns: A list of matching permissions.
-
-        """
-        search_terms = {u'per_page': per_page}
-        if hasattr(self, 'name'):
-            search_terms[u'name'] = self.name  # pylint:disable=no-member
-        if hasattr(self, 'resource_type'):
-            # pylint:disable=no-member
-            search_terms[u'resource_type'] = self.resource_type
-
-        response = client.get(
-            self.path('base'),
-            data=search_terms,
-            **self._server_config.get_client_kwargs()
-        )
-        response.raise_for_status()
-        return response.json()['results']
-
-
-class Ping(Entity):
+class Ping(Entity, EntitySearchMixin):
     """A representation of a Ping entity."""
 
     def __init__(self, server_config=None, **kwargs):
@@ -2588,6 +2551,7 @@ class Product(
         EntityCreateMixin,
         EntityDeleteMixin,
         EntityReadMixin,
+        EntitySearchMixin,
         EntityUpdateMixin):
     """A representation of a Product entity."""
 
@@ -2646,10 +2610,8 @@ class Product(
                 parse('6.1')):
             if attrs is None:
                 attrs = self.read_json()
-            attrs['organization'] = _get_org_attrs(
-                self._server_config,
-                attrs['organization']['label'],
-            )
+            org = _get_org(self._server_config, attrs['organization']['label'])
+            attrs['organization'] = org.get_values()
         return super(Product, self).read(entity, attrs, ignore)
 
     def list_repositorysets(self, per_page=None):
@@ -3182,6 +3144,7 @@ class Role(
         EntityCreateMixin,
         EntityDeleteMixin,
         EntityReadMixin,
+        EntitySearchMixin,
         EntityUpdateMixin):
     """A representation of a Role entity."""
 
@@ -3698,6 +3661,7 @@ class User(
         EntityCreateMixin,
         EntityDeleteMixin,
         EntityReadMixin,
+        EntitySearchMixin,
         EntityUpdateMixin):
     """A representation of a User entity.
 
--- a/nailgun/entity_mixins.py
+++ b/nailgun/entity_mixins.py
@@ -19,6 +19,12 @@ else:  # pragma: no cover
     import _thread as thread  # pylint:disable=import-error
     import http.client as http_client  # pylint:disable=import-error
 
+# pylint:disable=too-many-lines
+# This module contains very extensive docstrings, so this module is easier to
+# understand than its size suggests. That said, it could be useful to split
+# each mixin in to a separate module. That would help to ensure that each
+# mixin stays independent.
+
 
 #: Default for ``poll_rate`` argument to
 #: :func:`nailgun.entity_mixins._poll_task`.
@@ -233,8 +239,8 @@ def _get_entity_id(field_name, attrs):
         return attrs[field_name_id]
     else:
         raise MissingValueError(
-            'Cannot find a value for the "{}" field. Searched for keys named '
-            '{}, but available keys are {}.'
+            'Cannot find a value for the "{0}" field. Searched for keys named '
+            '{1}, but available keys are {2}.'
             .format(field_name, (field_name, field_name_id), attrs.keys())
         )
 
@@ -266,8 +272,8 @@ def _get_entity_ids(field_name, attrs):
         return [entity['id'] for entity in attrs[plural_field_name]]
     else:
         raise MissingValueError(
-            'Cannot find a value for the "{}" field. Searched for keys named '
-            '{}, but available keys are {}.'
+            'Cannot find a value for the "{0}" field. Searched for keys named '
+            '{1}, but available keys are {2}.'
             .format(
                 field_name,
                 (field_name_ids, field_name, plural_field_name),
@@ -286,7 +292,7 @@ class NoSuchPathError(Exception):
 
 
 class NoSuchFieldError(Exception):
-    """Indicates that the assigned-to :class:`Entity` field does not exist."""
+    """Indicates that the referenced field does not exist."""
 
 
 class BadValueError(Exception):
@@ -708,8 +714,8 @@ class EntityCreateMixin(object):
         create
         └── create_json
             └── create_raw
-                ├── create_payload
-                └── create_missing
+                ├── create_missing
+                └── create_payload
 
     In short, here is what the methods do:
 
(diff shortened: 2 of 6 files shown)


nailgun/entities.py
```diff
@@ -56,9 +58,14 @@
...
 #     product.organization.id
 #
-# Unfortunately, these fields cannot simply be initialized with `None`, as the
-# server considers "nil" to be different from the absence of a value. This
-# inevitably means that instance attributes will be defined outside __init__.
+# Unfortunately, these fields cannot simply be initialized with `None`. These
+# have different effects:
+#
+#     product.description = None; product.update()
+#     del product.desciption; product.update()
```
sthirugn: description?

Ichimonji10: Products have a description field.

nailgun/entity_mixins.py
```diff
@@ -931,3 +937,325 @@ def update(self, fields=None):
...
+                        pass
+                else:
+                    try:
+                        attrs[field_name] = result[field_name]
+                    except KeyError:
+                        pass
+            normalized.append(attrs)
+        return normalized
+
+    def search(self, fields=None, query=None, filters=None):
```
Ichimonji10: This method contains an extensive docstring that demonstrates how to use the search mixin. (The examples are a little more verbose than necessary - I decided to opt for clarity over conciseness.) If you're just coming to the search mixin for the first time, this is a great place to start.

nailgun/entities.py
```diff
@@ -1970,27 +1983,14 @@ def create_missing(self):
...
-            )
-            response.raise_for_status()
-            results = response.json()['results']
+            results = self.search({'organization'}, {u'name': u'Library'})
             if len(results) != 1:
                 raise APIResponseError(
-                    'Could not find the "Library" lifecycle environment for '
-                    'organization {0}. Search results: {1}'
+                    u'Could not find the "Library" lifecycle environment for '
+                    u'organization {}. Search results: {}'
```
sthirugn: are we moving away from usage of using numbers ?

nailgun/entities.py
sthirugn: sorry I meant there is a typo

sthirugn:
ACK pending comments

nailgun/entities.py
Ichimonji10: Aha! Thanks for pointing this out.

nailgun/entities.py
Ichimonji10: The numbers are only necessary for python 2.6 and older. Nailgun has never supported Python 2.6 - it makes use of conveniences like dict comprehensions that aren't available in such versions of Python. I just don't see any particular reason to use numbers.

nailgun/entity_mixins.py
```diff
@@ -931,3 +937,325 @@ def update(self, fields=None):
...
+
+            Organization(name='foo').search(query={'per_page': 50})
+
+        In some cases, the server's search facilities may be insufficient, or
+        it may be inordinately difficult to craft a search query. In this case,
+        you can filter search results locally. For example, to ask the server
+        for a list of all organizations and then locally search through the
+        results for the organization named "foo"::
+
+            Organization().search(filters={'name': 'foo'})
```
Ichimonji10: The `filters` attribute is the only portion of this PR that I find to be really questionable:

* Someone who knows what they're doing can easily write a list comprehension that does everything `search_filter()` does and more (filtering on foreign key fields is not supported) quite trivially.
* I think it'd be more beneficial for us to focus our efforts on auto-generating good quality searches. That's a much tougher problem. Users shouldn't have to write queries all the time, which is something they have to do right now.
* I expect it'd be more useful to have something like an `expect` argument, where you can tell nailgun how many results you expect to be returned by any given query.

nailgun/entity_mixins.py
```diff
@@ -931,3 +937,325 @@ def update(self, fields=None):
...
+        1. Generate a search query. If ``fields`` is specified, all named
+           fields are included in the query. Otherwise, all populated fields
+           are included.
+        2. Merge ``query`` in to the generated search query.
+        3. Return the result.
+
+        The rules for generating a search query are simple, and can be
+        illustrated by example::
+
+            >>> SomeEntity(
```
omaciel: I'm confused by the example here. The docstring mentions that this is an example for generating a search query... should this have used `search_payload` instead?

nailgun/entity_mixins.py
```diff
@@ -931,3 +937,325 @@ def update(self, fields=None):
...
+        certain responses can only be parsed if the search query is known. If
+        that is the case, this method will be given a new ``payload`` argument,
+        where ``payload`` is what was sent to the server. This will only work
+        if ``search_raw`` and ``search`` are able to generate identical
+        payloads. Pending extensive use of :class:`EntitySearchMixin`:
+
+        * :meth:`search` may alter ``fields`` and ``query`` at will.
+        * :meth:`search_payload` may alter ``fields`` and ``query`` in an
+          idempotent manner.
+        * No other method should atler ``fields`` or ``query``.
```
omaciel: s/atler/alter

nailgun/entity_mixins.py
```diff
@@ -931,3 +937,325 @@ def update(self, fields=None):
...
+        Values on an entity are used to generate a search query, and the
+        ``fields`` argument can be used to specify which fields should be used
+        when generating a search query::
+
+            org = Organization(name='foo', description='bar')
+            results = org.search()  # Search by name and description.
+            results = org.search(fields={'name', 'description'})  # Same.
+            results = org.search(fields={'name'})  # Search by name.
+            results = org.search(fields={'description'})  # Search by desc
+            results = org.search(fields=set())  # Search for all organizations.
```
omaciel: The syntax here sort of goes against the 'pattern' used for the other examples. Imho, if I wanted to search for all organizations, I'd call `results = Organization().search()`. Using `fields=set()` is not as intuitive... to me :) I don't have a solution or recommendation per se, but if `results = Organization().search()` == `results = Organization().search(fields=set())` or `results = Organization().search()` == `results = org.search(fields=set())` I guess I'm ok with considering the use of `fields=set()` as a 'hidden feature' :)

nailgun/entity_mixins.py
```diff
@@ -931,3 +937,325 @@ def update(self, fields=None):
...
+        of 'foo'::
+
+            Organization().search(query={'search': u'name="foo"'})
+
+        The example above is rather pointless: it is easier and more concise to
+        use a generated query. But — and this is a **very** important "but" —
+        the manual search query is melded in to the generated query. This can
+        be used to great effect::
+
+            Organization(name='foo').search(query={'per_page': 50})
```
omaciel: Love it!

nailgun/entity_mixins.py
omaciel: In this case is the filter doing a `name like 'foo'` query? Whereas `Organization(name='foo').search()` would be a `name == 'foo'`?

Ichimonji10: `filters` performs exact matching. The highlighted code above has the same effect as this:

```python
orgs = entities.Organization().search()
orgs = [org for org in orgs if org.name == 'foo']
```

This:

```python
Organization(name='foo').search()
```

...passes a search term to the server and lets the server conduct the search. The latter is vastly more efficient.

Unfortunately, `search_payload` is very naive right now. Here's what it'll do:

```python
>>> entities.Organization(name='foo').search_payload()
{'name': 'foo'}
```

And unfortunately, that search will fail. It'll fail silently, in fact - Satellite will respond to this search by returning **every single organization**.

nailgun/entity_mixins.py
Ichimonji10: Yes. Good catch.

nailgun/entity_mixins.py
Ichimonji10: :+1:

nailgun/entities.py
Ichimonji10: :+1: Fixed.

nailgun/entity_mixins.py
Ichimonji10: As of right now, all of these statements hold true:

```python
Organization().search() == Organization().search(fields=set())
Organization().search() == Organization().search(set())
Organization().search() == org.search(fields=set())
Organization().search() == org.search(set())
```

I've done this in order to make `search()` similar to `update()`. `update()` has this signature:

```python
def update(self, fields=None):
```

By default, `update()` sends all fields:

```python
org = entities.Organization(id=1, description='new description').update()
```

But you can also choose which fields to send:

```python
org = entities.Organization(id=1).read()
org.description = 'new description'
org = org.update(['description'])
```

Similarly, `search()` uses all fields:

```python
orgs = entities.Organization(description='new description').search()
```

But you may want to choose which fields to search by:

```python
org = entities.Organization(id=1).read()
orgs = org.search({'name'})
assert org.name == orgs[0].name
```

Now, you absolutely are right in pointing out that it's unusual to ask for a `set()` as an argument. That's pretty new.

Thoughts?

nailgun/entity_mixins.py
Ichimonji10: By the way, the more I think about an `expect` argument, the more I want to add one in. Consider how cool this would be:

```python
    organization = Organization().search(
        query={'search': 'name="Systems Engineering"'},
        expect=1,
    )[0]
    location = Location().search(
        query={'search': 'name="Boston"'},
        expect=1,
    )[0]
    hostgroup = HostGroup().search(
        query={'search': 'name="RHEV 3.5"'},
        expect=1,
    )[0]
```

nailgun/entity_mixins.py
Ichimonji10: BTW, @elyezer has opined that it would be good to make `query` the first argument, simply because that's going to be a commonly used argument. Here's why I opted for `fields` as the first argument:

* Doing so makes `search()` similar to `update()`: the very first argument to both is `fields`.
* Doing so pushes people toward using the `fields` argument when possible. I'd like to focus our efforts on improving the `search_payload` method instead of just hacking around its naive implementation for ever and ever.

I can see the flip side, though. Thoughts?

nailgun/entity_mixins.py
omaciel: Love it!

nailgun/entity_mixins.py
omaciel: Totally onboard with keeping argument ordering consistently. Sadly I don't have any clever suggestions to propose, but I think that once we start using this code we will figure out what the pain points are, etc. So I'd vote to proceed with the current design and see where it will take us. Also, it will be great to hear from @oshtaier @abalakh @elyezer et all for the opinion.

omaciel:
ACK

nailgun/entities.py
```diff
@@ -1970,27 +1983,14 @@ def create_missing(self):
...
-                data={
-                    u'name': u'Library',
-                    # pylint:disable=no-member
-                    u'organization_id': self.organization.id,
-                },
-                **self._server_config.get_client_kwargs()
-            )
-            response.raise_for_status()
-            results = response.json()['results']
+            results = self.search({'organization'}, {u'name': u'Library'})
```
oshtaier: Could we use such statements here?:
```
results = self.search({u'organization_id': self.organization}, {u'name': u'Library'})
```
or
```
results = self.search({u'organization_id': self.organization.id}, {u'name': u'Library'})
```
I just don't understand difference between ```{'organization'}``` and ```{u'name': u'Library'}``` parameters representation

nailgun/entity_mixins.py
```diff
@@ -931,3 +937,325 @@ def update(self, fields=None):
...
+                payload[field] = value
+        payload.update(query)
+        return payload
+
+    def search_raw(self, fields=None, query=None):
+        """Search for entities.
+
+        Make an HTTP GET call to ``self.path('base')``. Return the response.
+
+        .. WARNING:: Subclasses should not fiddle with the ``fields`` or
```
oshtaier: Could you please clarify more that warning message?

nailgun/entity_mixins.py
```diff
@@ -931,3 +937,325 @@ def update(self, fields=None):
...
+        ``fields`` argument can be used to specify which fields should be used
+        when generating a search query::
+
+            org = Organization(name='foo', description='bar')
+            results = org.search()  # Search by name and description.
+            results = org.search(fields={'name', 'description'})  # Same.
+            results = org.search(fields={'name'})  # Search by name.
+            results = org.search(fields={'description'})  # Search by desc
+            results = org.search(fields=set())  # Search for all organizations.
+            results = org.search(fields={'title'})  # Error!
```
oshtaier: Here we have error, because in ```Organization(name='foo', description='bar')``` we did not specify ```title``` field value explicitly, right?

nailgun/entity_mixins.py
rplevka: :+1:

oshtaier: >Unfortunately, search_payload is very naive right now. Here's what it'll do:
```
>>> entities.Organization(name='foo').search_payload()
{'name': 'foo'}
```
>And unfortunately, that search will fail. It'll fail silently, in fact - Satellite will respond to this search by returning >every single organization.

Could you please provide some examples why ```search_payload``` is not so good at that moment?
And explain more text above. That part is not obvious for me as providing organization name should be sufficient to find particular organzation

nailgun/entity_mixins.py
omaciel: `Organization` doesn't have a `title` field

nailgun/entities.py
Ichimonji10:
```python
>>> set(['organization'])
{'organization'}
>>> {'organization'}
{'organization'}
>>> {'organization'} == set(['organization'])
True
```

nailgun/entity_mixins.py
Ichimonji10: @omaciel nailed it.

> Here we have error, because in `Organization(name='foo', description='bar')` we did not specify `title` field value explicitly, right?

Correct. You're telling nailgun to create a search using the "title" attribute, but the "title" attribute is not set.

> and at that moment it has `None` value?

Incorrect. No `title` attribute has been set at all.

```python
>>> from nailgun.entities import Organization
>>> org = Organization(name='foo', description='bar').
>>> hasattr(org, 'title')
False
>>> org.title
  File "<stdin>", line 1, in <module>
AttributeError: 'Organization' object has no attribute 'title'
```

Further example:

```python
>>> from nailgun.entities import Organization
>>> org = Organization(name='foo', description='bar')                                                     
>>> org.search_payload()
{'description': 'bar', 'name': 'foo'}
>>> org.search_payload({'name'})
{'name': 'foo'}
>>> org.search_payload({'title'})                                                                         
  File "<stdin>", line 1, in <module>
  File "/home/ichimonji10/code/nailgun/nailgun/entity_mixins.py", line 1012, in search_payload
    value = getattr(self, field)
AttributeError: 'Organization' object has no attribute 'title'
```

nailgun/entity_mixins.py
Ichimonji10: Let's say that the `User` class inherits from this mixin like so:

```python
class User(Entity, EntitySearchMixin):
    pass
```

In this case, `User` is a subclass of both `Entity` and `EntitySearchMixin`. That method may define a `search_raw` method:

```python
class User(Entity, EntitySearchMixin):
    def search_raw(self, fields=None, query=None):
        super(User, self).search_raw(fields, query)
```

So far, so good. But the subclass should *not* fiddle with either `fields` or `query`. This is bad:

```python
class User(Entity, EntitySearchMixin):
    def search_raw(self, fields=None, query=None):
        fields.add{'name'}
        query['per_page'] = 50
        super(User, self).search_raw(fields, query)
```

nailgun/entities.py
Ichimonji10: > Could we use such statements here?:
> ```
> results = self.search({u'organization_id': self.organization}, {u'name': u'Library'})
> ```
> or
> ```
> results = self.search({u'organization_id': self.organization.id}, {u'name': u'Library'})
> ```

No. The first parameter to `search()`, `fields`, is a set of field names. Do not pass a dict where a set is expected.

Ichimonji10: > I think about search functionality always in terms ```.search(file='something')```, so it is not so obvious for me what ```.search(file)``` should do

That's a fine thing to do if you want to write imperative code. But I'm doing my darndest to take advantage of the fact that Python is an object oriented language.

Ichimonji10: > I just don't understand difference between `{'organization'}` and `{u'name': u'Library'}` parameters representation

The first argument specifies which fields nailgun should use when automatically generating a search expression for you. The second argument is merged in to the first - it allows for customization after a search has been generated. Here's a simple example of how the queries are merged:

```python
>>> org = entities.Organization(name='foo', description='bar')                                            
>>> org.search_payload()
{'name': 'foo', 'description': 'bar'}
>>> org.search_payload({'name'})
{'name': 'foo'}
>>> org.search_payload({'name'}, {'per_page': 50})
{'per_page': 50, 'name': 'foo'}
```

I will fully admit that the search generator currently *sucks*. It's awfully naive. But I'm OK with that - I know that we have the tools in hand to make it better, and I want us to put some effort in to doing so rather than writing manual search queries for the rest of our days.

nailgun/entity_mixins.py
oshtaier: Organization actually has 'title' field
```
class Organization():
...
        self._fields = {
            'compute_resource': entity_fields.OneToManyField(
                AbstractComputeResource
            ),
            'config_template': entity_fields.OneToManyField(ConfigTemplate),
            'description': entity_fields.StringField(),
            'domain': entity_fields.OneToManyField(Domain),
            'environment': entity_fields.OneToManyField(Environment),
            'hostgroup': entity_fields.OneToManyField(HostGroup),
            'label': entity_fields.StringField(str_type='alpha'),
            'media': entity_fields.OneToManyField(Media),
            'name': entity_fields.StringField(required=True),
            'realm': entity_fields.OneToManyField(Realm),
            'smart_proxy': entity_fields.OneToManyField(SmartProxy),
            'subnet': entity_fields.OneToManyField(Subnet),
            'title': entity_fields.StringField(),
            'user': entity_fields.OneToManyField(User),
        }
```
or API call
```
"results": [{"id":5,"name":"𩀵𨡼𥲲𨐿䎘𦮱","title":"𩀵𨡼𥲲𨐿䎘𦮱","label":"547374fa-2a20-4c14-8ba2-9fa099e6a396","created_at":"2015-07-20T10:09:43Z", ........
```

So we can use only name and description in your example as search criteria, right?

nailgun/entity_mixins.py
Ichimonji10: @oshtaier This will return every single organization:

    GET /katello/api/v2/organizations
    {'name': 'foo'}

This will return all organizations with a name (equaling? like? containing?) foo:

    GET /katello/api/v2/organizations
    {'search': 'name=foo'}

This will return no organizations at all:

    GET /katello/api/v2/organizations
    {'search': 'name:foo'}

The search syntax is undefined, no errors are ever returned when an invalid search is submitted, and exactly which fields can be submitted in an outer hash versus an inner "search" hash varies on a per-URL basis.

So yes, providing a name is sufficient to find a particular organization. But how do you provide that name to the API? Not in a way that you might expect.

Ichimonji10: As an aside, the practice of passing parameters in the body of a GET request is horrible. It's absolutely horrible and violates HTTP semantics. Satellite should *NOT* support it. HTTP-aware network intermediaries are at liberty to drop the body of any GET request; caches and proxies will be confused by the practice; developers, sysadmins and network admins that know a thing or two about HTTP will cringe at this.

nailgun/entity_mixins.py
oshtaier: But we have similar idea for, let's say, for read method:
```
class Domain(Entity, EntityReadMixin):

        if attrs is None:
            attrs = self.read_json()
        attrs['domain_parameters_attributes'] = attrs.pop('parameters')
        return super(Domain, self).read(entity, attrs, ignore)
```
Here, we change parameter and send it to super class

Or it is completely different?

nailgun/entity_mixins.py
Ichimonji10: > Organization actually has 'title' field

Correct.

But merely having a field is useless when generating a search query. In order to generate a search query, some attributes must be present.

I don't understand your questions completely, but I may be able to answer with another example. Let's take the object you've shown:

```python
>>> cfg = ServerConfig('https://hp-ml310egen8-01.rhts.eng.bos.redhat.com', auth=('admin', 'changeme'), verify=False)
>>> cfg.save('oshtaier')
>>> cfg = ServerConfig.get('oshtaier')
>>> org = entities.Organization(cfg, id=5).read()
>>> org.search_payload().keys()
dict_keys(['title', 'config_template_ids', 'domain_ids', 'compute_resource_ids', 'hostgroup_ids', 'environment_ids', 'user_ids', 'smart_proxy_ids', 'id', 'subnet_ids', 'label', 'media_ids', 'description', 'name'])
```

See how all of these different keys are included in the search?

nailgun/entity_mixins.py
Ichimonji10: You can fiddle with those arguments at will in `read`. From `search_normalize`:

```
At this time, it is possible to parse all search results without
knowing what search query was used. However, it is possible that
certain responses can only be parsed if the search query is known. If
that is the case, this method will be given a new ``payload`` argument,
where ``payload`` is what was sent to the server. This will only work
if ``search_raw`` and ``search`` are able to generate identical
payloads. Pending extensive use of :class:`EntitySearchMixin`:

* :meth:`search` may alter ``fields`` and ``query`` at will.
* :meth:`search_payload` may alter ``fields`` and ``query`` in an
    idempotent manner.
* No other method should alter ``fields`` or ``query``.
```

nailgun/entity_mixins.py
JacobCallahan: That last point is exactly the sentiment I encountered while trying to find elegant solutions for this exact issue! Not sure if there is a better way for them to implement it, unless they create a new path or something.

nailgun/entities.py
JacobCallahan: :+1:

nailgun/entity_mixins.py
JacobCallahan: @oshtaier be careful. the line ``org = Organization(name='foo', description='bar')``` has only created an Organization locally. At this point, the Org only has those two fields set. This will remain the case until the Org is assigned a title or is created at the server. See below for some examples.

```
In [5]: org = entities.Organization(name='foo', description='bar')

In [6]: org
Out[6]: nailgun.entities.Organization(nailgun.config.ServerConfig(url=u'https://rhsm-qe-1.rhq.lab.eng.bos.redhat.com/', auth=(u'admin', u'changeme')), description='bar', name='foo')

In [7]: org.get_values()
Out[7]: {'description': 'bar', 'name': 'foo'}

In [8]: org.get_fields()
Out[8]: 
{'compute_resource': <nailgun.entity_fields.OneToManyField at 0x7ff314bf79d0>,
 'config_template': <nailgun.entity_fields.OneToManyField at 0x7ff314bf7a10>,
 'description': <nailgun.entity_fields.StringField at 0x7ff314bf7a50>,
 'domain': <nailgun.entity_fields.OneToManyField at 0x7ff314bf7a90>,
 'environment': <nailgun.entity_fields.OneToManyField at 0x7ff314bf7ad0>,
 'hostgroup': <nailgun.entity_fields.OneToManyField at 0x7ff314bf7b10>,
 'id': <nailgun.entity_fields.IntegerField at 0x7ff314bf7e10>,
 'label': <nailgun.entity_fields.StringField at 0x7ff314bf7b50>,
 'media': <nailgun.entity_fields.OneToManyField at 0x7ff314bf7b90>,
 'name': <nailgun.entity_fields.StringField at 0x7ff314bf7bd0>,
 'realm': <nailgun.entity_fields.OneToManyField at 0x7ff314bf7c10>,
 'smart_proxy': <nailgun.entity_fields.OneToManyField at 0x7ff314bf7c50>,
 'subnet': <nailgun.entity_fields.OneToManyField at 0x7ff314bf7c90>,
 'title': <nailgun.entity_fields.StringField at 0x7ff314bf7cd0>,
 'user': <nailgun.entity_fields.OneToManyField at 0x7ff314bf7d10>}

In [9]: org.title
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-9-5f07d6148295> in <module>()
----> 1 org.title

AttributeError: 'Organization' object has no attribute 'title'
```

nailgun/entities.py
JacobCallahan: Does this break compatibility with those that use 2.6 with robottelo?

nailgun/entity_mixins.py
Ichimonji10: @JacobCallahan Aye. It's a nut of a problem. I can think of some solutions, like adding a "filter" attribute to fields. But the best thing for me to do might be to just spend some time talking with the devs and better understanding their design.

nailgun/entities.py
Ichimonji10: @JacobCallahan You're right, this will break 2.6 compatibility. I'll revert this change.

JacobCallahan:
@Ichimonji10 I guess my single biggest concern is that the default results per page is 20. This could result in a number of failed searches, especially when dealing with automation. For example, over the lifecycle of the machines we currently run our automation on, there are Organizations with id's exceeding 530! Of course, a portion of these are deleted but most remain. For most non-automated users, they may never encounter this problem. But if we are running automation and are searching for an organization we just created, there is a high probability that the search will fail.

There are two main solutions to this issue. 
1. Hand it off to the person running the search. This would basically mean that we need to include a large page count in each search query. That is fine, but maybe not optimal.
2. Nailgun defaults to a per_page that is large enough to return a significant number of results. This helps to ensure our target is included in the results, but can quickly bog down a nailgun-based search. 

Of course, it would be optimal if the person performing the search includes a server-side filter. Just my thoughts.

JacobCallahan:
This has come a long way from our initial prototypes. Good job!
ACK, pending the few remaining comments and @oshtaier.

Ichimonji10:
@JacobCallahan 

Whether a search returns a lot of results depends on what we're keying our search on, right? For example, if I'm searching for something by name or label, then only one result should be returned - no matter how many organizations exist.

omaciel:
> 1. Hand it off to the person running the search. This would basically mean that we need to include a large page count in each search query. That is fine, but maybe not optimal.

@JacobCallahan I think that if someone writing a test specifically wants all entries for $ENTITY, it would be reasonable that he/she passes a filter for how many results to fetch (aka `per_page`)... but, perhaps the search method in a future version should return a generator instead of a list of results? That may work as data would be fetched lazily?

JacobCallahan:
@Ichimonji10 Yep, that is why I included that last line. However, if someone were to run something like this on our automation boxes, they would be sorely disappointed. 
```my_org = Organization().search(filters={'name': 'just_created'})```

@omaciel Yes, that is what I was thinking for my first solution option.

**Comment by Ichimonji10** [comment]
@Ichimonji10 Yep, that is why I included that last line. However, if someone were to run something like this on our automation boxes, they would be sorely disappointed. 
`my_org = Organization().search(filters={'name': 'just_created'})`

@JacobCallahan Believe it or not...

```python
>>> orgs = entities.Organization().search()
>>> len(orgs)
627
```

Ichimonji10:
@JacobCallahan Anyway, yes, a user would be disappointed. But I think that we can reasonably file a bug against the server for that issue, where the bug says "invalid searches are silently ignored."

oshtaier:
ACK

nailgun/entities.py
JacobCallahan: :+1:

Ichimonji10:
Hey all! I've taken your feedback in to account and tweaked the pull request. As it stands, I've made only very minor tweaks to the code itself. The majority of my changes have to do with documentation. I've added a short section in to the "examples" document and fiddled with many of the docstrings. There are three changes of particular note:

First, I've made changes to the "Examples" document.

Second, I've expanded the docstring for method `EntitySearchMixin.search_payload` and added several links pointing to it. It now gives more insight in to what the final search queries look li.e

Third, I'm now referencing `LifecycleEnvironment` in my examples instead of `Organization`. Why? Because this is an invalid search:

    GET /katello/api/v2/organizations
    {'name': 'foo', 'description': 'bar'}  # server ignores these params

But this is a valid search:

    GET /katello/api/v2/lifecycle_environments
    {'name': 'foo', 'organization_id': 1}  # server accepts these params

The default implementation of `search_payload` is pretty naive. But I at least want users to be able to copy and paste the examples in the documentation!

I'm currently running `make test-foreman-api-threaded` (Thanks @JacobCallahan!) as a final santiy check before opening this up for merging.

omaciel:
w00t!

Ichimonji10:
I've done a sanity check here: https://github.com/SatelliteQE/robottelo/pull/2610#issuecomment-124696030

I think this PR is open for merging.

oshtaier:
Re-ACK. Thanks @Ichimonji10 for you patience
:+1: For change from ```Organization``` entity to ```LifecycleEnvironment```

Ichimonji10:
:tada:

elyezer:
:+1:

Issue #151, opened 2015-07-20
Can't read() product, which has sync_plan assigned


abalakh:
When I try to `read()` a product, which has sync plan assigned like this:
```python
org = entities.Organization().create()
product = entities.Product(organization=org).create()
syncplan = entities.SyncPlan(organization=org).create()
syncplan.add_products(product.id)
product = product.read()
```
I receive following error:
```python
Error
Traceback (most recent call last):
  File "/home/andriy/Downloads/robottelo/tests/foreman/api/test_syncplan.py", line 454, in test_synchronize_sync_plan_with_one_product
    result = self.products[0].read()
  File "/home/andriy/Downloads/nailgun/nailgun/entities.py", line 2711, in read
    return super(Product, self).read(entity, attrs, ignore)
  File "/home/andriy/Downloads/nailgun/nailgun/entity_mixins.py", line 685, in read
    id=entity_id,
  File "/home/andriy/Downloads/nailgun/nailgun/entities.py", line 3433, in __init__
    _check_for_value('organization', kwargs)
  File "/home/andriy/Downloads/nailgun/nailgun/entities.py", line 137, in _check_for_value
    'A value must be provided for the "{0}" field.'.format(field_name)
TypeError: A value must be provided for the "organization" field.
```

abalakh:
Some thoughts about the defect:
When we are reading the entity, inside `EntityReadMixin.read()` we're collecting all its fields and entities, placing everything into one resulting entity and returning it.
Why sync_plan makes difference? According to [https://github.com/SatelliteQE/nailgun/blob/master/nailgun/entity_mixins.py#L683-L687](https://github.com/SatelliteQE/nailgun/blob/master/nailgun/entity_mixins.py#L683-L687):
```python
                    referenced_entity = field.entity(
                        self._server_config,
                        id=entity_id,
                    )
                setattr(entity, field_name, referenced_entity)
```
 all sub-entities are initialized by their type and id. It's ok for 99% cases, but for sync_plan we also need to specify organization to initialize it [https://github.com/SatelliteQE/nailgun/blob/master/nailgun/entities.py#L3425](https://github.com/SatelliteQE/nailgun/blob/master/nailgun/entities.py#L3425):
```python
 def __init__(self, server_config=None, **kwargs):
        _check_for_value('organization', kwargs)
...
        self._meta = {
            # pylint:disable=no-member
            'api_path': '{0}/sync_plans'.format(self.organization.path()),
        }
```
[...]

PR #158, opened 2015-07-23
Revert "Update SyncPlan entity"

Ichimonji10:
This reverts commit f0198c184d3efdd4f5179e2b13b2f1bc85031642.

The `POST /katello/api/v2/sync_plan` path is not available. As a result, setting
`self._meta['api_path']` to such makes it impossible to create sync plans via
the API.

### DIFF
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -32,7 +32,7 @@ from nailgun.entity_mixins import (
     EntityUpdateMixin,
     _poll_task,
 )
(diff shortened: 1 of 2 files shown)


abalakh:
ACK.
it's a pity it can't be used, as it could resolve #151 :disappointed: 
But yeah, not only create, also listing, adding and removing products is missing in this path too.

sthirugn:
ACK

PR #159, opened 2015-07-24
Remove redundant 'null' parameter from fields and entities

### LINKED ISSUE #135: Remove "null" from fields and entities
The base `Field` class has the following signature:

```python
    def __init__(
            self,
            required=False,
            choices=None,
            default=_SENTINEL,
            null=False):
```

The "required", "choices" and "default" attributes are all used when generating random data for testing purposes. However, the "null" attribute is not used at all, anywhere, to the best of my knowledge. This is silly. Why should a completely useless attribute be bumming around?

Drop the `null` attribute. Doing so will not affect anything right now, given that there is no usage of the field. It should be done before version 1.0 is released, as dropping the attribute is a backward-incomptable change. (Some client may reference the attribute.) In addition, the attribute can be added again after version 1.0 is released without breaking anything - only prompting a minor version number bump.

oshtaier:
Closes #135

```
git grep null
docs/Makefile:ifeq ($(shell which $(SPHINXBUILD) >/dev/null 2>&1; echo $$?), 1)
```

Executed random test from api to make sure that all fields populated properly and no exception raised:
```
nosetests tests/foreman/api/test_location.py -m test_create_location_with_description
.
Ran 1 test in 2.844s

OK
```

### DIFF
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -289,10 +289,7 @@ class Architecture(
     def __init__(self, server_config=None, **kwargs):
         self._fields = {
             'name': entity_fields.StringField(required=True),
-            'operatingsystem': entity_fields.OneToManyField(
-                OperatingSystem,
-                null=True,
-            ),
+            'operatingsystem': entity_fields.OneToManyField(OperatingSystem),
         }
         self._meta = {
             'api_path': 'api/v2/architectures',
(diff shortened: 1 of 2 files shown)


Ichimonji10:
ACK

Thank you!

elyezer:
ACK

PR #161, opened 2015-07-28
Only require unittest 2 for Python 3.3 and below

Ichimonji10:
Adjust the `requirements-dev.txt` file so that unittest2 is only installed if
Python 3.3 or below is being used. Adjust test modules to only import unittest2
if Python 3.3 or below is being used.

unittest2 is used on Python 3.3 and below so that the `subTest` context manager
can be used in unit tests. There is no need to use unittest2 on Python 3.4 and
above, as this feature has been added to the standard library.

Bonus: refactor out some unnecessary version-checking code.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,6 +1,6 @@
 # For `make test`
 mock
-unittest2
+unittest2 ; python_version < '3.4'
 
 # For `make lint`
 flake8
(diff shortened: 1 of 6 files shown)


Ichimonji10:
This change introduces yet more conditional imports in to NailGun. It's quite possible to refactor out the conditional imports, but there's some undesireable side-effects associated with doing so. See the last commit here: https://github.com/SatelliteQE/nailgun/compare/master...Ichimonji10:requirements2

sthirugn:
ACK

elyezer:
ACK

Issue #163, opened 2015-07-28
Set default version


Ichimonji10:
The `ServerConfig` class does not currently require that the "version" attribute be provided. This is problematic when an entity must make an assumption about what version the server is running. For example, entities have different sets of fields based on what version the server is running, so of course an assumption must be made - but what version should be assumed?

Update method `ServerConfig.__init__`. If no version is passed in when instantiating a `ServerConfig` object, then some version should be assumed. I would support assuming version '1!0' (epoch 1, version 0). This version will be greater than any version provided by Satellite, thus letting us assume that a user is running a development version of Satellite by default. This will work so long as Satellite retains its current versioning scheme.

PR #164, opened 2015-07-28
Standardize versioning logic

Ichimonji10:
Many entities need to make an assumption about what software version the server
is running. This is currently a PITA to do. The expressions that result are
idiosyncratic and ugly.

Related to #163.


sthirugn:
ACK

elyezer:
ACK

PR #168, opened 2015-08-05
Added discovered_host entity

sghai:
I would like to add tests around discovery feature. But for this, we need infrastructure along with provisioning where we can boot a unknown host via PXE and it gets IP from configured DHCP server on specific subnet on Foreman server. We don't have such infra for now though I really wanted to have one.

So, here I'm trying to discover a dummy  host via api to add basic tests. We have two api's to create discovered host.

```
POST /api/v2/discovered_hosts 
Create a discovered host
```
Unfortunately, this api doesn't work. I filed an issue in upstream after discussing with dev: http://projects.theforeman.org/issues/11277

The other api is to upload facts for discovered host and it creates the host as well. So I'm trying to create the dummy discovered host using this.

```
POST /api/v2/discovered_hosts/facts 
Upload facts for a host, creating the host if required
```

### DIFF
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -544,6 +544,76 @@ class AbstractComputeResource(
         }
 
 
+class DiscoveredHosts(
+        Entity,
+        EntityCreateMixin,
+        EntityDeleteMixin,
+        EntityReadMixin,
+        EntityUpdateMixin):
+    """A representation of a Foreman Discovered Host entity."""
+
+    def __init__(self, server_config=None, **kwargs):
+        self._fields = {
+            'name': entity_fields.StringField(required=True),
+            'ip': entity_fields.IPAddressField(required=True),
+            'mac': entity_fields.MACAddressField(required=True),
+        }
+        self._meta = {
+            'api_path': '/api/v2/discovered_hosts',
+            'server_modes': ('sat'),
+        }
+        super(DiscoveredHosts, self).__init__(server_config, **kwargs)
+
+    def path(self, which=None):
+        """Extend ``nailgun.entity_mixins.Entity.path``.
+
+        The format of the returned path depends on the value of ``which``:
+
+        facts
+            /discovered_hosts/facts
+
+        ``super`` is called otherwise.
+
+        """
+        if which == 'facts':
+            return '{0}/{1}'.format(
+                super(DiscoveredHosts, self).path(which='base'),
+                which
+            )
+        return super(DiscoveredHosts, self).path(which)
+
+    def create_payload(self):
+        """Wrap submitted data within an extra dict.
+
+        For more information, see `Bugzilla #1151220
+        <https://bugzilla.redhat.com/show_bug.cgi?id=1151220>`_.
+
+        """
+        return {
+            u'discovered_host': super(
+                DiscoveredHosts,
+                self
+            ).create_payload()
+        }
+
+    def upload_facts(self, payload):
+        """Helper to update facts for discovered host, and create the host.
+
+        :param payload: Parameters that are encoded to JSON and passed in
+            with the request.
+        :returns: The server's response, with all JSON decoded.
+        :raises: ``requests.exceptions.HTTPError`` If the server responds with
+            an HTTP 4XX or 5XX message.
+
+        """
+        response = client.post(
+            self.path('facts'),
+            payload,
+            **self._server_config.get_client_kwargs()
+        )
+        return _handle_response(response, self._server_config)
+
+
 class DockerComputeResource(AbstractComputeResource):
     """A representation of a Docker Compute Resource entity."""
 
--- a/tests/test_entities.py
+++ b/tests/test_entities.py
@@ -93,6 +93,7 @@ class InitTestCase(TestCase):
                 entities.ContentUpload,
                 entities.ContentView,
                 entities.ContentViewVersion,
+                entities.DiscoveredHosts,
                 entities.DockerComputeResource,
                 entities.DockerHubContainer,
                 entities.Domain,
@@ -197,6 +198,7 @@ class PathTestCase(TestCase):
                 (entities.ConfigTemplate, '/config_templates'),
                 (entities.ContentView, '/content_views'),
                 (entities.ContentViewVersion, '/content_view_versions'),
+                (entities.DiscoveredHosts, '/discovered_hosts'),
                 (entities.Organization, '/organizations'),
                 (entities.Product, '/products'),
                 (entities.RHCIDeployment, '/deployments'),
@@ -252,6 +254,7 @@ class PathTestCase(TestCase):
         for entity, which in (
                 (entities.ConfigTemplate, 'build_pxe_default'),
                 (entities.ConfigTemplate, 'revision'),
+                (entities.DiscoveredHosts, 'facts'),
         ):
             with self.subTest():
                 path = entity(self.cfg).path(which=which)
@@ -413,6 +416,7 @@ class CreatePayloadTestCase(TestCase):
                 entities.AbstractDockerContainer,
                 entities.Architecture,
                 entities.ConfigTemplate,
+                entities.DiscoveredHosts,
                 entities.Domain,
                 entities.Environment,
                 entities.Host,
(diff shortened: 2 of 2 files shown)


sghai:
The issue with this `POST /api/v2/discovered_hosts` api is:

```
In [6]: dhost1=DiscoveredHosts(
server_config,
name="testd",                                          
ip='192.168.110.220',
mac="52:54:00:2B:D1:DD"
).create_json()
/usr/lib/python2.7/site-packages/requests/packages/urllib3/util/ssl_.py:90: InsecurePlatformWarning: A true SSLContext object is not available. This prevents urllib3 from configuring SSL appropriately and may cause certain SSL connections to fail. For more information, see https://urllib3.readthedocs.org/en/latest/security.html#insecureplatformwarning.
  InsecurePlatformWarning
WARNING:nailgun.client:Received HTTP 422 response: {
  "error": {"id":null,"errors":{"ip":["has already been taken"],"discovery_attribute_set":["can't be blank"]},"full_messages":["IP address has already been taken","Discovery attribute set can't be blank"]}
}
```

sghai:
So I used other api to create dummy host and for that I added a helper function 'upload_facts'. It works well but I think that helper function can be improved.

```
In [5]: DiscoveredHosts().upload_facts({
'name': "testd",
'ipaddress': '192.168.110.220',
'discovery_bootif': '52:54:00:2B:D1:DD'
})
/usr/lib/python2.7/site-packages/requests/packages/urllib3/util/ssl_.py:90: InsecurePlatformWarning: A true SSLContext object is not available. This prevents urllib3 from configuring SSL appropriately and may cause certain SSL connections to fail. For more information, see https://urllib3.readthedocs.org/en/latest/security.html#insecureplatformwarning.
  InsecurePlatformWarning
Out[5]: 
{u'architecture_id': None,
 u'build': False,
 u'certname': None,
 u'comment': None,
 u'compute_profile_id': None,
 u'compute_resource_id': None,
 u'content_source_id': None,
 u'content_view_id': None,
 u'created_at': u'2015-08-05T06:11:16Z',
 u'discovery_rule_id': None,
 u'disk': None,
 u'domain_id': None,
 u'enabled': True,
 u'environment_id': None,
 u'grub_pass': u'',
 u'hostgroup_id': None,
 u'id': 22,
 u'image_file': None,
 u'image_id': None,
 u'installed_at': None,
 u'ip': u'192.168.110.220',
 u'last_compile': None,
 u'last_freshcheck': None,
 u'last_report': u'2015-08-05T06:11:16Z',
 u'lifecycle_environment_id': None,
 u'location_id': 2,
 u'mac': u'52:54:00:2b:d1:dd',
 u'managed': False,
 u'medium_id': None,
 u'model_id': None,
 u'name': u'mac5254002bd1dd',
 u'operatingsystem_id': None,
 u'organization_id': 1,
 u'otp': None,
 u'owner_id': None,
 u'owner_type': None,
 u'primary_interface': None,
 u'provision_method': None,
 u'ptable_id': None,
 u'puppet_ca_proxy_id': None,
 u'puppet_proxy_id': None,
 u'puppet_status': 0,
 u'realm_id': None,
 u'root_pass': None,
 u'serial': None,
 u'source_file_id': None,
 u'subnet_id': 2,
 u'updated_at': u'2015-08-05T06:11:16Z',
 u'use_image': None,
 u'uuid': None}
```

nailgun/entities.py
```diff
@@ -544,6 +544,58 @@ def create_payload(self):
...
+
+        :param params: Parameters that are encoded to JSON and passed in
+            with the request.
+        :returns: The server's response, with all JSON decoded.
+        :raises: ``requests.exceptions.HTTPError`` If the server responds with
+            an HTTP 4XX or 5XX message.
+
+        """
+        response = client.post(
+            self.path() + '/facts',
```
sghai: Is it correct way to form the path: `/api/v2/discovered_hosts/facts`

nailgun/entities.py
```diff
@@ -544,6 +544,58 @@ def create_payload(self):
...
+            {"facts": params},
```
sghai: Is it correct way to define hash containing diff facts ?

oshtaier: According to API documentation - I think it is pretty redundant statement. It will be fine to have just `params` and on Robottelo side: `discover_host.upload_facts({u'facts': 'fact_1'})`

nailgun/entities.py
oshtaier: Usually, we overload `path()` method and put here our custom paths. Please, refer to `class  Organization` or `class ContentView` for better understanding

nailgun/entities.py
```diff
@@ -544,6 +544,58 @@ def create_payload(self):
...
+
+        """
+        return {
+            u'discovered_host': super(
+                DiscoveredHosts,
+                self
+            ).create_payload()
+        }
+
+    def upload_facts(self, params):
```
oshtaier: nitpick: use `payload` word instead `params`. We had huge re-factor process to move to that word for *all* helper methods

nailgun/entities.py
sghai: Perfect..makes sense. thanks

nailgun/entities.py
sghai: yeah..agreed.

nailgun/entities.py
sghai: Yeah.. I tried earlier `path()` method but failed with below error:

```
/home/sghai/QE/nailgun/nailgun/entity_mixins.pyc in path(self, which)
    475         elif (which == 'self' or which is None) and hasattr(self, 'id'):
    476             return urljoin(base + '/', str(self.id))  # pylint:disable=E1101
--> 477         raise NoSuchPathError
    478 
    479     def get_fields(self):

NoSuchPathError: 
```
I'll give that another try..

sghai: Ok, Figured out. which should be set to `base`. doc++

oshtaier: Great!

oshtaier:
It will be nice if @Ichimonji10 review that PR and give his feedback too.

ACK pending unit tests. I know that can be painful and time consuming to add them, but we started to use coveralls and it is necessary to follow that standard. Of course, if that PR is urgent - everything is discussable.

nailgun/entities.py
```diff
@@ -544,6 +544,75 @@ def create_payload(self):
...
+        The format of the returned path depends on the value of ``which``:
+
+        facts
+            /discovered_hosts/facts
+
+        ``super`` is called otherwise.
+
+        """
+        if which == 'facts':
+            return '{0}/facts'.format(
```
abalakh: nitpick: change to 
```python
        if which == 'facts':
            return '{0}/{1}'.format(
                super(DiscoveredHosts, self).path(which='base'),
                which
            )
```

hardcoding `'{0}/facts'` path will work in this case, but if you'd like to add some more paths in future, you'll have a lot of `if ... return` statements, 1 per each path. 
in the proposed approach you'll need to update only condition, e.g. `if which == 'foo' or which == 'bar'` and everything will work fine without any extra code.

nailgun/entities.py
```diff
@@ -544,6 +544,75 @@ def create_payload(self):
...
+            u'discovered_host': super(
+                DiscoveredHosts,
+                self
+            ).create_payload()
+        }
+
+    def upload_facts(self, payload):
+        """Helper to update facts for discovered host, and create the host.
+
+        :param params: Parameters that are encoded to JSON and passed in
```
abalakh: `s/params/payload/`

nailgun/entities.py
sghai: :+1:

nailgun/entities.py
sghai: ah..missed that one. will fix now.

sghai:
Thanks Alex, Andrii for review comments. I'll see and try to add unit tests too. thanks for all comments.

abalakh:
ACK.
AFAIK, unit tests are not mandatory, but are very appreciated, so it would be great to at least try to add them :)
Feel free to ping me in case of any questions about them.

abalakh:
Re-ACK

sghai:
Thank you @oshtaier @abalakh for helping me to learn new stuff around coveralls. That was really helpful.

sthirugn:
ACK

sthirugn:
Good job.  Multiple ACKs here already. Merging this now to get this going.

Ichimonji10:
Please don't merge quite yet. Posting a diff now.

Ichimonji10:
Well damn.

PR #169, opened 2015-08-05
Update the `DiscoveredHost` entity

Ichimonji10:
* According to the API docs, updates should be wrapped in an extra hash.
* Simplify the test case for `upload_facts`.
* Rename `upload_facts` to `facts`, to match the underlying path.

### DIFF
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -590,13 +590,19 @@ class DiscoveredHosts(
 
         """
         return {
+            u'discovered_host': super(DiscoveredHosts, self).create_payload()
+        }
+
+    def update_payload(self, fields=None):
+        """Wrap submitted data within an extra dict."""
+        return {
             u'discovered_host': super(
                 DiscoveredHosts,
                 self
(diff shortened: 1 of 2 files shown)


sthirugn:
ACK

oshtaier:
ACK
@sghai check that PR. It is update for yours

Ichimonji10:
@oshtaier Thank you for the quick review.

sghai:
Thanks @Ichimonji10 for updates. I liked how you simplified `test_facts` with payload. Thank you.
 One thing would like to know:
- `update_payload` is to update the payload fields,  right ?

oshtaier:
We use `create_payload` to form proper parameters when we need, for example, wrap them like we do in current PR: (`u'discovered_host':`). Then we use these parameters to create our entity

`update_payload` method used for absolutely the same goal, but for update procedure (`EntityUpdateMixin`)

sghai:
That's what I wanted to know. Thanks @oshtaier .

PR #173, opened 2015-08-06
Fix a bug in method Subscription.upload

Ichimonji10:
An HTTP POST request should be made, not an HTTP PUT request. Without this fix,
the server responds to requests with an HTTP 404 error.


sthirugn:
ACK

omaciel:
ACK

sthirugn:
@omaciel do you have wifi in the beach? ;)

omaciel:
@sthirugn hehe I'm back home since late yesterday

PR #174, opened 2015-08-07
Make `Subscription.search` function

Ichimonji10:
Method `Subscription.search_raw` lays out the problem and chosen solution:

> The ``GET /katello/api/v2/subscriptions`` API call is not available.
> Instead, one of the following must be used:
>
> * ``GET /katello/api/v2/activation_keys/<id>/subscriptions``
> * ``GET /katello/api/v2/organizations/<id>/subscriptions``
> * ``GET /katello/api/v2/systems/<id>/subscriptions``
>
> Use the activation key path if ``self.activation_key`` is set, use the
> organization path if ``self.organization`` is set, use the system path
> if ``self.system`` is set, or raise an exception otherwise.

In order to make this work, several new paths and unit tests were also added.

### DIFF
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -31,6 +31,7 @@ from nailgun.entity_mixins import (
     EntityReadMixin,
     EntitySearchMixin,
     EntityUpdateMixin,
+    MissingValueError,
     _poll_task,
 )
 from packaging.version import Version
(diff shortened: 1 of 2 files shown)


oshtaier:
ACK

elyezer:
ACK

PR #175, opened 2015-08-12
Re-work helper methods, again

Ichimonji10:
There are two kinds of methods on entities:

* Methods inherited from a mixin. For example, `ActivationKey.create` is
  inherited from the `EntityCreateMixin`.
* Methods defined on that entity, such as `ActivationKey.add_subscriptions`.

The latter set of methods was recently overhauled to make them more uniform. The
methods are now named after the path they call, they are placed on the correct
entities, and their signatures are very similar. However, there are still
differences. Some methods provide default values for their arguments; some
require a second argument; some require only one argument but of a different
kind than others; and so on. This is bad for several reasons:

* It is hard to learn the various methods. A user has to look at the signature
  of each individual method and figure out exactly what arguments it requires.
  This significantly raises the conceptual weight of the methods.
* It is hard to design the various methods. A developer has to spend time
  examining the API docs for the method call to figure out what information the
  API requires, the developer has to reference a variety of other helper methods
  to see which implementation should be copied, and so on.
* Tests for the various methods are not quite the same. This means that there's
  a lot of copy-cat code in the unit test suite.

Design a method signature that addresses these issues. It is an approach that
was not discussed in SatelliteQE/nailgun#41. Here are some examples of method
calls before the change:

    # ActivationKey.add_subscriptions
    activation_key.add_subscriptions({
        'quantity': 1,
        'subscription_id': subs.id,
    })

    # Subscription.upload
    with open(manifests.clone(), 'rb') as manifest:
        sub.upload({'organization_id': org.id}, manifest)

And after the change:

    # ActivationKey.add_subscriptions
    activation_key.add_subscriptions(data={
        'quantity': 1,
        'subscription_id': subs.id,
    })

    # Subscription.upload
    with open(manifests.clone(), 'rb') as handle:
        sub.upload(data={'organization_id': org.id}, files={'content': handle})

The updated approach is more verbose, but it's utterly uniform. All methods have
a signature like so:

    def method_name(self, synchronous=True, **kwargs):

All of the `**kwargs` are passed directly to `nailgun.client` before a call is
made. As a result, clients using one of the helper methods can access the full
power of `nailgun.client`/`requests`!

### DIFF
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -2872,60 +2872,57 @@ class RepositorySet(
             'api_path': '{0}/repository_sets'.format(self.product.path()),
         }
 
-    def available_repositories(self):
+    def available_repositories(self, synchronous=True, **kwargs):
         """Lists available repositories for the repository set
 
-        :returns: Returns list of available repositories for the repository set
+        :param synchronous: What should happen if the server returns an HTTP
+            202 (accepted) status code? Wait for the task to complete if
+            ``True``. Immediately return the server's response otherwise.
+        :param kwargs: Arguments to pass to requests.
+        :returns: The server's response, with all JSON decoded.
+        :raises: ``requests.exceptions.HTTPError`` If the server responds with
+            an HTTP 4XX or 5XX message.
 
         """
-        response = client.get(
-            self.path('available_repositories'),
-            **self._server_config.get_client_kwargs()
-        )
-        return _handle_response(response, self._server_config)['results']
+        kwargs = kwargs.copy()  # shadow the passed-in kwargs
+        kwargs.update(self._server_config.get_client_kwargs())
+        response = client.get(self.path('available_repositories'), **kwargs)
+        return _handle_response(response, self._server_config, synchronous)
 
-    def enable(self, payload, synchronous=True):
+    def enable(self, synchronous=True, **kwargs):
         """Enables the RedHat Repository
 
         RedHat Repos needs to be enabled first, so that we can sync it.
 
-        :param payload: Parameters that are encoded to JSON and passed in
-            with the request. See the API documentation page for a list of
-            parameters and their descriptions.
         :param synchronous: What should happen if the server returns an HTTP
             202 (accepted) status code? Wait for the task to complete if
-            ``True``. Immediately return JSON response otherwise.
-        :returns: Returns information of the async task if an HTTP
-            202 response was received and synchronus set to ``True``.
-            Return JSON response otherwise.
+            ``True``. Immediately return the server's response otherwise.
+        :param kwargs: Arguments to pass to requests.
+        :returns: The server's response, with all JSON decoded.
+        :raises: ``requests.exceptions.HTTPError`` If the server responds with
+            an HTTP 4XX or 5XX message.
 
         """
-        response = client.put(
-            self.path('enable'),
-            payload,
-            **self._server_config.get_client_kwargs()
-        )
+        kwargs = kwargs.copy()  # shadow the passed-in kwargs
+        kwargs.update(self._server_config.get_client_kwargs())
+        response = client.put(self.path('enable'), **kwargs)
         return _handle_response(response, self._server_config, synchronous)
 
-    def disable(self, payload, synchronous=True):
+    def disable(self, synchronous=True, **kwargs):
         """Disables the RedHat Repository
 
-        :param payload: Parameters that are encoded to JSON and passed in
-            with the request. See the API documentation page for a list of
-            parameters and their descriptions.
         :param synchronous: What should happen if the server returns an HTTP
             202 (accepted) status code? Wait for the task to complete if
-            ``True``. Immediately return JSON response otherwise.
-        :returns: Returns information of the async task if an HTTP
-            202 response was received and synchronus set to ``True``.
-            Return JSON response otherwise.
+            ``True``. Immediately return the server's response otherwise.
+        :param kwargs: Arguments to pass to requests.
+        :returns: The server's response, with all JSON decoded.
+        :raises: ``requests.exceptions.HTTPError`` If the server responds with
+            an HTTP 4XX or 5XX message.
 
         """
-        response = client.put(
-            self.path('disable'),
-            payload,
-            **self._server_config.get_client_kwargs()
-        )
+        kwargs = kwargs.copy()  # shadow the passed-in kwargs
+        kwargs.update(self._server_config.get_client_kwargs())
+        response = client.put(self.path('disable'), **kwargs)
         return _handle_response(response, self._server_config, synchronous)
 
     def path(self, which=None):
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -2575,14 +2575,22 @@ class Product(
             attrs['organization'] = org.get_values()
         return super(Product, self).read(entity, attrs, ignore)
 
-    def sync(self):
-        """Synchronize :class:`repositories <Repository>` in this product."""
-        response = client.post(
-            self.path('sync'),
-            {},
-            **self._server_config.get_client_kwargs()
-        )
-        return _handle_response(response, self._server_config)
+    def sync(self, synchronous=True, **kwargs):
+        """Synchronize :class:`repositories <Repository>` in this product.
+
+        :param synchronous: What should happen if the server returns an HTTP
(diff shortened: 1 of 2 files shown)


Ichimonji10:
The corresponding Robottelo PR is at SatelliteQE/robottelo#2677.

Ichimonji10:
I feel silly submitting *yet another* pull request for reworking methods. But I'm doing this because I *really* want to finish the "Beta" milestone and because I *really* want to get it right.

This change does produce more verbose client code. But the code isn't much more verbose than what we already have, and it should make working with NailGun muuuuuch easier. Every single helper method (with one exception that I'm having trouble finding) has a completely identical method signature, which rocks. Also, clients have access to the full feature set of Requests, such as doing things like setting headers. And I feel confident that this approach will work for all helper methods that we might add in the future (pending massive changes in Satellite's API).

nailgun/entities.py
```diff
@@ -267,41 +267,39 @@ def path(self, which=None):
...
             an HTTP 4XX or 5XX message.
 
         """
-        response = client.put(
-            self.path('add_subscriptions'),
-            payload,
-            **self._server_config.get_client_kwargs()
-        )
-        return _handle_response(response, self._server_config)
+        kwargs = kwargs.copy()  # shadow the passed-in kwargs
```
sthirugn: What is the purpose of copy?

sthirugn:
+1 for standardizing and adding synchronous option to many functions.

sthirugn:
Do not merge this PR until the bug fix errata (scheduled for tomorrow) goes out

nailgun/entities.py
oshtaier: Interested in that too. Please, explain

nailgun/entities.py
```diff
@@ -267,41 +267,39 @@ def path(self, which=None):
             )
         return super(ActivationKey, self).path(which)
 
-    def add_subscriptions(self, payload):
+    def add_subscriptions(self, synchronous=True, **kwargs):
```
oshtaier: Why do we add `synchronous=True` to all helper functions? As far as I remember we need that parameter to manage asynchronous tasks.
Do we have so serious changes in the application logic that all of these operations become asynchronous?

oshtaier:
Don't like idea to change approaches so frequently, but we have to as any future fix will give us too much headache. (after 'Beta').
In my opinion with current **kwargs idea - it becomes even harder to determine what things should be passed to the functions. I mean here logical part, not the technical one which becomes simple as ever it can be.

Thanks @Ichimonji10 

ACK

Ichimonji10:
> Don't like idea to change approaches so frequently, but we have to as any future fix will give us too much headache. (after 'Beta').

I completely agree. I dislike changing approaches. But I want to get things right *now*.

> In my opinion with current **kwargs idea - it becomes even harder to determine what things should be passed to the functions. I mean here logical part, not the technical one which becomes simple as ever it can be.

Again, I completely agree. I'm making a trade-off between technical and logical convenience. I don't want to commit to the technical task of maintaining argument lists and correctly packing up data for all of these methods. And there's a *huge* technical benefit to using `**kwargs`: you can now use every single [`requests.request`](http://docs.python-requests.org/en/latest/api/#requests.request) argument! But the flip side is that users now have to read the Satellite API docs and correctly pack up arguments.

> Why did we add `synchronous=True` to all helper functions? As far as I remember we need that parameter to manage asynchronous tasks.
> Do we have so serious changes in the application logic that all of these operations become asynchronous?

The server won't return an HTTP 202 status code in response to all requests, and the `synchronous` argument has no effect when a non-202 status code is returned. The `synchronous` argument only has an effect if the server returns an HTTP 202 response. From the docs:

        :param synchronous: What should happen if the server returns an HTTP
            202 (accepted) status code? Wait for the task to complete if
            ``True``. Immediately return the server's response otherwise.

Think about it his way: the server may respond to any request with an HTTP 500 status code, and when it does, we react by raising an exception.Similarly, the server may respond to any request with an HTTP 202 status code, and when it does, we will now respond by polling the referenced task. You can control the former with Python's exception handling mechanisms, and you can control the latter with the `synchronous` argument.

oshtaier:
If we adding that parameter to each method, so can't we move that parameter on some higher level and remove it from signatures at all?

Ichimonji10:
> If we adding that parameter to each method, so can't we move that parameter on some higher level and remove it from signatures at all?

There's not much in the way of layering for the helper methods. They're much simpler than the mixins. Let's say you call `ActivationKey.add_subscriptions`. When you do, this is (roughly) the call stack:

    nailgun.entities.ActivationKey.add_subscriptions (assemble request arguments)
        nailgun.client.put (perform logging, encode JSON)
            requests.put (make actual request)
        nailgun.entities._handle_response (check for HTTP 202, 4XX, 5XX status codes, decode JSON)

I think it's really useful to automatically check for 202 status codes, much like we check for HTTP 4XX and 5XX status codes. So how do we handle them?

As an aside, we can't rely on requests to provide some mechanism for handling 202 status codes. I don't think Requests provides such functionality, and even if it did, Satellite doesn't provide a full URL for monitoring tasks. As a result, the right place to handle 202 status codes is in NailGun.

oshtaier:
ohh, I am totally agree with that.
I meant much primitive thing. According to the code, I am not sure why we need things like:
```
def add_subscriptions(self, synchronous=True, **kwargs):
def content_override(self, synchronous=True, **kwargs):
def facts(self, synchronous=True, **kwargs):
.....
....
```
and
```
if synchronous is True and response.status_code == ACCEPTED:
```
and not
```
def add_subscriptions(self, **kwargs):
def content_override(self, **kwargs):
def facts(self, **kwargs):
......
return _handle_response(response, self._server_config)
....
if response.status_code == ACCEPTED:
```

I don't see situation where we have ```synchronous=False```

nailgun/entities.py
Ichimonji10: Let's say the user wants to call `AbstractDockerContainer.logs` (or
`DockerHubContainer.logs`) on several containers. The client could put together
a request that looks like this:

```python
lost_list = [
    container.logs(data={'tail': 50, 'stdout': True, 'stderr': False})
    for container in containers
]
```

This works just fine. Another way to write the above is this:

```python
data = {'tail': 50, 'stdout': True, 'stderr': False}
lost_list = [container.logs(data=data) for container in containers]
```

In this latter form, we depend upon `data` being the same for every single
request. Right? It would be pretty weird if we passed `data` in to a method and
the method altered `data`. In fact, if the method altered `data`, that would
break the latter way of writing our code!

`kwargs.copy()` prevents `data` from being altered. Check this out:

```python
>>> def logs(data):
...     data.update({'tail': 1000})
... 
>>> data = {'tail': 50, 'stdout': True, 'stderr': False}
>>> logs(data)
>>> data
{'tail': 1000, 'stderr': False, 'stdout': True}
```

Uh oh! Using `kwargs.copy()` avoid this issue:

```python
>>> def logs(data):
...     data = data.copy()  # shadow the passed-in argument
...     data.update({'tail': 1000})
... 
>>> data = {'tail': 50, 'stdout': True, 'stderr': False}
>>> logs(data)
>>> data
{'tail': 50, 'stderr': False, 'stdout': True}
```

oshtaier: :+1: Thanks

Ichimonji10:
> I don't see situation where we have `synchronous=False`

Let's imagine that there's a user who wants to make a Satellite start syncing a repository. The user wants to script this, and they prefer to use the API over the CLI (maybe they don't have SSH access to the server?). The user could write a script like this:

```python
#!/usr/bin/env python
from nailgun.entities import Repository
Repository(name='...').search()[0].sync()
```

Great, right? Oh, no, that doesn't work, because the script will hang in perpetuity waiting for the server to respond. They need to write something like this:

```python
#!/usr/bin/env python
from nailgun.entities import Repository
Repository(name='...').search()[0].sync(synchronous=False)
```

oshtaier:
Sounds great

Ichimonji10:
Thanks for reviewing my PRs so promptly and asking good questions.

oshtaier:
ACK once more

nailgun/entities.py
sthirugn: nicely done!

sthirugn:
ACK

elyezer:
ACK

PR #180, opened 2015-08-17
Added more details to Docker container creation procedure

oshtaier:
Provided information should help people to get familiar with the code more quickly.

### DIFF
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -895,7 +895,8 @@ class AbstractDockerContainer(
                 str_type='latin1',
             ),
             'compute_resource': entity_fields.OneToOneField(
-                AbstractComputeResource
+                AbstractComputeResource,
+                required=True,
             ),
             'cpu_set': entity_fields.StringField(),
             'cpu_shares': entity_fields.StringField(),
@@ -998,7 +999,13 @@ class AbstractDockerContainer(
 
 
 class DockerHubContainer(AbstractDockerContainer):
-    """A docker container that comes from Docker Hub."""
+    """A docker container that comes from Docker Hub.
+
+    .. WARNING:: The ``repository_name`` field references an image repository
+        on the `Docker Hub <https://hub.docker.com/>`, not a locally created
+        :class:`nailgun.entities.Repository`.
+
+    """
 
     def __init__(self, server_config=None, **kwargs):
         self._fields = {


nailgun/entities.py
```diff
@@ -949,7 +951,12 @@ def logs(self, synchronous=True, **kwargs):
 
 
 class DockerHubContainer(AbstractDockerContainer):
-    """A docker container that comes from Docker Hub."""
+    """A docker container that comes from Docker Hub.
+    .. WARNING:: 'repository_name' field here refer to the name of the
+    image repository from your docker hub resource and doesn't have any
+    reference to repositories created in the application itself
```
Ichimonji10: The `WARNING` directive acts like the `param` and `returns` directive. It should be separated from surrounding text and successive lines indented, or else Sphinx doesn't parse it correctly:

screenshot_2015-08-17_09-37-53

I've fiddled with the wording below in an attempt to make it flow more nicely.

```
    """A docker container that comes from Docker Hub.

    .. WARNING:: The ``repository_name`` field references an image repository
        on the `Docker Hub <https://hub.docker.com/>`_, not a locally created
        :class:`nailgun.entities.Repository`.

    """
```

Ichimonji10:
ACK pending comment.

oshtaier:
@Ichimonji10 Thanks!

Ichimonji10:
Thank *you*. ACK.

nailgun/entities.py
```diff
@@ -844,9 +844,11 @@ def __init__(self, server_config=None, **kwargs):
             'command': entity_fields.StringField(
                 required=True,
                 str_type='latin1',
+                default='top',
```
elyezer: I vote for not adding this because there are some images that does not provide top like default centos.

elyezer: Actually we should talk more about this. As default image is busybox then the top command will be available. But regarding docker the command should be passed explicitly.

Ichimonji10: This is an interesting addition. NailGun users won't encounter this unless they explicitly ask for fake data to be generated. And should Ontic development ever move forward, the fake data generation may be pulled out of NailGun entirely. Consequently, I'm willing to tailor the fake data generation in a way that suits Robottelo's needs.

oshtaier: 1) Actually, I am performing all development using CentOS default installation image and do nothing extra to have 'top' command, but I have it.
2) We can use another command that any Linux package has. For example 'ls' (as far as I know).
3) I like to have that statement, because that minimize code from robottelo side and anyway you will provide some value for that field, so it doesn't matter - you have failure on Nailgun or Robottelo side

But I am fine with any solution, just let me know your decision, because I have PR in Robottelo blocked by this one

elyezer: The only concern is that top is not common and there are some images that does have it. Which will cause an error.

We have a place in robottelo to add some default values [1]. To me there seems to be a better place to define these default values.

If @oshtaier and @Ichimonji10 thinks this is a good addition, I will not block it. I'm only not completely sure about this being on NailGun.

[1] https://github.com/SatelliteQE/robottelo/blob/master/robottelo/common/helpers.py#L375

Ichimonji10: @elyezer has a good point. In the other cases where we provide default values, those defaults work 100% of the time. How about [this](https://github.com/SatelliteQE/robottelo/pull/2692#issuecomment-132203235) solution?

oshtaier:
Updated PR according to above discussion and moved 'command' logic on Robottelo side. Thanks @elyezer and @Ichimonji10  for suggestions.

Ichimonji10:
ACK

elyezer:
ACK

PR #184, opened 2015-08-24
Rename DiscoveryRule field

Ichimonji10:
Rename the `search` field to `search_` in order to avoid a name conflict between
the field and method names.

### DIFF
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -650,20 +650,22 @@ class DiscoveryRule(
         EntityReadMixin,
         EntitySearchMixin,
         EntityUpdateMixin):
-    """A representation of a Foreman Discovery Rule entity."""
+    """A representation of a Foreman Discovery Rule entity.
+
+    .. NOTE:: The ``search_`` field is named as such due to a naming conflict
+        with :meth:`nailgun.entity_mixins.Entity.path`.
+
+    """
 
     def __init__(self, server_config=None, **kwargs):
         self._fields = {
-            'name': entity_fields.StringField(required=True),
-            'search': entity_fields.StringField(required=True),
-            'hostgroup': entity_fields.OneToOneField(
-                HostGroup,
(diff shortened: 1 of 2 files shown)


nailgun/entities.py
```diff
@@ -650,20 +650,22 @@ class DiscoveryRule(
...
-    """A representation of a Foreman Discovery Rule entity."""
+    """A representation of a Foreman Discovery Rule entity.
+
+    .. NOTE:: The ``search_`` field is named as such due to a naming conflict
+        with :meth:`nailgun.entity_mixins.Entity.path`.
+
+    """
 
     def __init__(self, server_config=None, **kwargs):
         self._fields = {
```
Ichimonji10: I've ordered these fields alphabetically, so this change looks bigger than it actaully is.

nailgun/entities.py
```diff
@@ -677,19 +679,27 @@ def create_payload(self):
...
+        payload = super(DiscoveryRule, self).create_payload()
+        if 'search_' in payload:
+            payload['search'] = payload.pop('search_')
+        return {u'discovery_rule': payload}
+
+    def read(self, entity=None, attrs=None, ignore=None):
+        """Rename ``search`` to ``search_``."""
+        if attrs is None:
+            attrs = self.read_json()
+        attrs['search_'] = attrs.pop('search')
```
omaciel: Is it safe to assume that we will always get ``attrs['search']``? Do we need to check if it exists first?

omaciel:
ACK pending question.

nailgun/entities.py
Ichimonji10: Yes, it's safe to assume that the server will always return the "search" attribute. The `read()` method assumes that the server returns all attributes.

sghai:
ACK

elyezer:
:+1:

PR #186, opened 2015-08-26
Work around BZ 1257255

Ichimonji10:
According to the bug:

> An HTTP GET /api/v2/discovery_rules/:id returns a hash of information about a
> discovery rule. The "max_count" key is missing from this hash.

Work around this issue. When `DiscoveryRule.read` is called, submit a no-op PUT
request and fetch the `max_count` attribute from the response. Test the change.
See: https://bugzilla.redhat.com/show_bug.cgi?id=1257255

Test results from Robottelo:

    $ nosetests tests/foreman/api/test_multiple_paths.py -m DiscoveryRule
    ..........S
    Ran 11 tests in 14.982s

    OK (SKIP=1)
    $ nosetests tests/foreman/api/test_discoveryrule.py
    ..........
    Ran 10 tests in 10.777s

    OK

### DIFF
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -689,10 +689,27 @@ class DiscoveryRule(
         return {u'discovery_rule': payload}
 
     def read(self, entity=None, attrs=None, ignore=None):
-        """Rename ``search`` to ``search_``."""
+        """Work around a bug. Rename ``search`` to ``search_``.
+
+        For more information on the bug, see `Bugzilla #1257255
+        <https://bugzilla.redhat.com/show_bug.cgi?id=1257255>`_.
+
+        """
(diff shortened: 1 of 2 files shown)


Ichimonji10:
There are three failing API tests. This fixes one of the three failing tests. SatelliteQE/robottelo#2715 fixes the other two.

oshtaier:
one more workaround :(
ACK

Ichimonji10:
> one more workaround :(

And a slightly longer `entities.py`. :/ But that's part of what makes NailGun useful. And anyway, I'm confident that `entities.py` can be split up at some point.

elyezer:
ACK

PR #187, opened 2015-08-27
Adding new 'registry' entity

oshtaier:
After recent fix for https://bugzilla.redhat.com/show_bug.cgi?id=1206679 defect, it become possible to add one more entity to API

### DIFF
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -2769,6 +2769,44 @@ class Realm(
         ).read()
 
 
+class Registry(
+        Entity,
+        EntityCreateMixin,
+        EntityDeleteMixin,
+        EntityReadMixin,
+        EntityUpdateMixin):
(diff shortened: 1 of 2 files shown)


elyezer:
Does this entity works? An sample run would be good.

ACK

Ichimonji10:
I think #2716 shows that this works. My only concern is the `read()` method. Does it work correctly? I'm wondering because of the presence of the `password` attribute. Satellite doesn't usually return sensitive info like that.

oshtaier:
I actually have PR pending in Robottelo right now that have a lot of tests proving that entity works properly

oshtaier:
Yes, it is returning password :) I verified that fact explicitly

Ichimonji10:
Aha! You're ahead of me. Thank you. ACK

Issue #189, opened 2015-08-27
URLField.gen_value doesn't return especially unique values


Ichimonji10:
Method ``URLField.gen_value`](http://nailgun.readthedocs.org/en/latest/api/nailgun.entity_fields.html#nailgun.entity_fields.URLField) frequently generates identical values. As a result of this issue, several work-arounds have been added to module `nailgun.entities`. One example is at [`Media.create_missing`](http://nailgun.readthedocs.org/en/latest/api/nailgun.entities.html#nailgun.entities.Media.create_missing):

```python
    def create_missing(self):
        """Give the ``path_`` instance attribute a value if it is unset.

        By default, :meth:`nailgun.entity_fields.URLField.gen_value` does not
        return especially unique values. This is problematic, as all media must
        have a unique path.

        """
        if not hasattr(self, 'path_'):
            self.path_ = gen_url(subdomain=gen_alpha())
        return super(Media, self).create_missing()
```

Let's fix `gen_url` and drop the work-arounds from module `nailgun.entities`.

PR #190, opened 2015-08-28
Make 'nailgun.entity_fields.URLField.gen_value' method always generates unique value

oshtaier:
+Removing corresponding workaround
+Removing redundant unit tests

### DIFF
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -22,7 +22,7 @@ workings of entity classes.
 
 """
 from datetime import datetime
-from fauxfactory import gen_alpha, gen_alphanumeric, gen_url
+from fauxfactory import gen_alphanumeric
 from nailgun import client, entity_fields
 from nailgun.entity_mixins import (
     Entity,
@@ -1496,9 +1496,9 @@ class Domain(
     def create_missing(self):
         """Customize the process of auto-generating instance attributes.
 
-        By default, :meth:`nailgun.entity_fields.URLField.gen_value` does not
-        return especially unique values. This is problematic, as all domain
-        names must be unique.
+        By default, :meth:`nailgun.entity_fields.StringField.gen_value` can
+        produce strings in both lower and upper cases, but domain name should
+        be always in lower case due logical reason.
 
         """
         if not hasattr(self, 'name'):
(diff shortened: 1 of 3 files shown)


nailgun/entities.py
```diff
@@ -1496,9 +1496,9 @@ def __init__(self, server_config=None, **kwargs):
     def create_missing(self):
         """Customize the process of auto-generating instance attributes.
 
-        By default, :meth:`nailgun.entity_fields.URLField.gen_value` does not
-        return especially unique values. This is problematic, as all domain
-        names must be unique.
+        By default, :meth:`nailgun.entity_fields.StringField.gen_value` can
+        produce strings in both lower and upper cases, but domain name should
+        be always in lower case due logical reason.
```
Ichimonji10: Thanks for fixing this. I must've copied and pasted. Blech!

Ichimonji10:
ACK.

Fixes #189.

abalakh:
ACK

nailgun/entity_fields.py
```diff
@@ -265,4 +266,4 @@ class URLField(StringField):
 
     def gen_value(self):
         """Return a value suitable for a :class:`URLField`."""
-        return gen_url()
+        return gen_url(subdomain=gen_alpha())
```
rplevka: 9gag and 4chan fans might get upset about this solution..how about this?
```python
return gen_url(subdomain=gen_alphanumeric())
```

also, how about including a `.` to support multiple-domain format (like `foo.bar.org`)?
of course it must not be at the beginning or at the end of the string

oshtaier: We do not restrict or introduce any limitation to possible URL, so if you like to use particular URL in your specific test case - just do it, but for generic scenario (for create_missing() method) I prefer to use most basic generator method

elyezer:
:+1:

nailgun/entity_fields.py
Ichimonji10: @rplevka You're right. It'd be good to generate more "interesting" data. I didn't both pushing for the change, though, because this PR really just works around issues in the underlying fauxfactory library. An even better solution is to make the `gen_url` function produce a greater variety of URLs and obey [RFC 2606](https://www.rfc-editor.org/rfc/rfc2606.txt).

omaciel: > [...] this PR really just works around issues in the underlying fauxfactory library.

What are they? Is there an issue filed to track them?

Ichimonji10: @omaciel There's two issues:

* `gen_url`, `gen_email` and possibly others generate real values that might be in use in the wild. For example, `gen_url` might spit out [test.com](https://www.test.com/).
* `gen_url` doesn't produce especially random values. That's why we're calling `gen_url(subdomain=gen_alpha())` instead of `gen_url()` here in NailGun.

omaciel/fauxfactory#26 is describes how `gen_email` produces "real" values. There aren't issues for the other methods that produce "real" values. There's not an issue for `gen_url`'s lack of randomness.

oshtaier: @omaciel from my observation `gen_url` method give us not so many variations of inputs and such data become similar pretty fast. Let's say, it generate:
`test.com`, `example.com`, `test.biz`, `example.biz` and few more, so even for 10 tests there is chance that we will have duplicated entry. Using proposed approach we decrease such chances in hundreds times

PR #199, opened 2015-09-03
[DO NOT MERGE] Updated entities for incremental update

sthirugn:


omaciel:
Can haz tests? :)

nailgun/entities.py
```diff
@@ -1124,6 +1128,23 @@ def promote(self, synchronous=True, **kwargs):
         response = client.post(self.path('promote'), **kwargs)
         return _handle_response(response, self._server_config, synchronous)
 
+    def inc_update(self, synchronous=True, **kwargs):
```
Ichimonji10: Can you rename this to `incremental_update`? I'm trying to keep method names identical to the underlying path called. I'm going to add a decorator at some point that'll formalize this convention.

nailgun/entities.py
```diff
@@ -1105,6 +1105,10 @@ def path(self, which=None):
             return '{0}/promote'.format(
                 super(ContentViewVersion, self).path(which='self')
             )
+        if which == 'incremental_update':
```
Ichimonji10: Please mention this new path in the docstring.

nailgun/entities.py
```diff
@@ -1598,12 +1619,24 @@ def create_payload(self):
         return {u'environment': super(Environment, self).create_payload()}
 
 
-class Errata(Entity):
+class Errata(Entity,
```
Ichimonji10: All on one line.

nailgun/entities.py
Ichimonji10: Please add some test coverage for this method. You can just copy and paste the relevant bits from https://github.com/SatelliteQE/nailgun/compare/master...Ichimonji10:inc-update

nailgun/entities.py
```diff
@@ -1105,6 +1105,10 @@ def path(self, which=None):
             return '{0}/promote'.format(
                 super(ContentViewVersion, self).path(which='self')
             )
+        if which == 'incremental_update':
+            return '{0}/incremental_update'.format(
+                super(ContentViewVersion, self).path(which=None)
+            )
```
abalakh: How about
```python
        if which in ('promote', 'incremental_update'):
            return '{0}/{1}'.format(
                super(ContentViewVersion, self).path(which='self'),
                which
            )
```
?

nailgun/entities.py
abalakh: Please, rename it to `def incremental_update` to match its path.

nailgun/entities.py
Ichimonji10: That won't work correctly. It'll produce a path like so:

    /katello/api/v2/content_view_versions/<id>/incremental_update

Which is incorrect. The current implementation will sometimes produce that path, too. `which` should be "base":

```python
super(ContentViewVersion, self).path(which='self'),
```

sthirugn:
Rebase got messed up, fixing it

sthirugn:
@Ichimonji10 my changes are done, please cherry pick

sthirugn:
this pr is tested with robottelo tests, the pr for robottelo will follow soon.

Ichimonji10:
Cherry-picked on to #200.

sthirugn:
The commit in this pr is cherrypicked to https://github.com/SatelliteQE/nailgun/pull/200.  Closing this pr.

PR #204, opened 2015-09-07
Updated entities with some fields and with update mixin

sghai:
Updated 3 entities:
- `Environment` with update mixin
- `media` with update mixin
- `SmartProxy` with update mixin and `location` and `organization` fields

PR #206, opened 2015-09-09
Updated entities with update mixin

sghai:
This PR includes similar code changes as in https://github.com/SatelliteQE/nailgun/pull/204

This updates 3 entities:
- `Environment` with update mixin
- `Media` with update mixin
- `SmartProxy` with update mixin and location and organization fields


sghai:
@Ichimonji10: I've updated the bz id for Media entity. Please review it again. thanks

nailgun/entities.py
```diff
@@ -2334,6 +2355,23 @@ def read(self, entity=None, attrs=None, ignore=None):
         attrs['path_'] = attrs.pop('path')
         return super(Media, self).read(entity, attrs, ignore)
 
+    def update(self, fields=None):
+        """Fetch a complete set of attributes for this entity.
+
+        Beware of `Bugzilla #1261047
+        <https://bugzilla.redhat.com/show_bug.cgi?id=1261047>`_:
```
Ichimonji10: Thanks for filing this bug. Much appreciated.

nailgun/entities.py
```diff
@@ -3330,13 +3368,19 @@ def update_payload(self, fields=None):
...
+        EntitySearchMixin,
+        EntityUpdateMixin):
     """A representation of a Smart Proxy entity."""
 
     def __init__(self, server_config=None, **kwargs):
         self._fields = {
             'name': entity_fields.StringField(required=True),
             'url': entity_fields.URLField(required=True),
+            'location': entity_fields.OneToManyField(Location),
+            'organization': entity_fields.OneToManyField(Organization),
```
Ichimonji10: Good choice. These should both be `OneToManyField`s.

This is a little bit confusing, because `GET /api/smart_proxies` only lets you search by one location or organization at a time:

    GET /api/smart_proxies
        List all smart proxies
        location_id: Scope by locations
        organization_id: Scope by organizations

However, the POST and PUT calls let you work with multiple locations or organizations. For example:

    POST /api/smart_proxies
        smart_proxy[location_ids]: REPLACE locations with given ids
        smart_proxy[organization_ids]: REPLACE organizations with given ids
        ...

The take-away is that the `SmartProxy` entity really does have one-to-many relationships with organizations and locations, and the GET call is just limited.

Ichimonji10:
Looks great. I assume that the manual results shown in #204 still hold true here too. ACK.

nailgun/entities.py
sghai: Ah.. thanks for sharing this. So far I was just trying `PUT` call.

sghai:
Yeah.. results are same and still applicable as shown in #204. thanks

elyezer:
ACK

omaciel:
:100:

PR #212, opened 2015-09-11
Fix typos in documentation

Ichimonji10:
One of the sample scripts in the documentation is missing a call to the
`create()` method. Add this missing method call.

One of the paragraphs mentions a "hass". This is a typo of "hash". Change the
word to "dict", which will likely be more familiar term to Python users.


JacobCallahan:
ACK

omaciel:
ACK

elyezer:
:+1:

PR #223, opened 2015-09-23
Make `_poll_task` more robust

### LINKED ISSUE #222: Method _poll_task needs to check if task has started running
Several of our automated tests have show that synchronization and publishing tasks are being polled only once and immediately failing with a `TaskFailedError` exception.

The problem seems to be related to how `_poll_task` only polls a task if its state is `running`. If the state is **not** `running` then we check for the final result, and if `result != 'success' we raise the exception.

Turns out that we have some tasks triggered by the automation that sit on the `planned` state for a while and since `'planned' != 'running'` we get out of the loop and stop polling right after the first check.

Based on the [source code](https://github.com/Dynflow/dynflow/blob/e9c02b60b26e4a8636c85b5d51192d468430d2f7/lib/dynflow/execution_plan.rb#L23) from *TheForeman*, a task can go through the following state transitions:

```ruby
    def self.state_transitions
      @state_transitions ||= { pending:  [:stopped, :scheduled, :planning],
                               scheduled: [:planning, :stopped],
                               planning: [:planned, :stopped],
                               planned:  [:running],
                               running:  [:paused, :stopped],
                               paused:   [:running],
                               stopped:  [] }
    end
```

So I believe that we need to make sure that when polling a task the following states should be observed before either exiting the loop or timing out: `scheduled`, `planning`, `planned` and `running`

An example of a failed test shows (snip):

```json
{"id":"97caeb65-60f1-44df-b482-1d244bb3412d",
"label":"Actions::Katello::Repository::Sync",
"pending":true,
"username":"admin",
"started_at":"2015-09-21T00:56:32Z",
"ended_at":null,
"state":"planned",
"result":"pending",
"progress":0.0,
"input":{"repository":{
"id":9,
"name":"쒲10𣊈쾹10𣊈𩶴麣10𣊈𩶴𨃉䧠惋10𣊈𩶴𨃉𧵷10𣊈𩶴𨃉𧵷𪥚10𣊈𩶴𨃉𧵷𪥚𪚑쾚낕",
"label":"4b6fa137-86c7-427b-b56f-d6078d03c69a"},
"product":{"id":3,"name":"10𣊈𩶴𨃉𧵷𪥚𪚑𩙉䧊10𣊈𩶴𨃉𧵷𪥚𪚑𩙉𩏞",
"label":"7eb07dfd-3dde-4c0d-92a5-ee10619929cb","cp_id":"1442796989475"}
```

Ichimonji10:
Fix #222:

> Several of our automated tests have show that synchronization and publishing
> tasks are being polled only once and immediately failing with a
> `TaskFailedError` exception.
>
> […]
>
> Turns out that we have some tasks triggered by the automation that sit on the
> `planned` state for a while and since `'planned' != 'running'` we get out of
> the loop and stop polling right after the first check.

`_poll_task` aborts polling if a task is anything other than the "running"
state. This is bad logic, as a task may proceed through the following states:

* paused
* pending
* planned
* planning
* running
* scheduled
* stopped

Fiddle with `_poll_task`:

* Abort polling if a task enters the "paused" or "stopped" state.
* Continue polling if a task is in any other state, such as "planning",
  "scheduled" or "running".

### DIFF
--- a/nailgun/entity_mixins.py
+++ b/nailgun/entity_mixins.py
@@ -82,34 +82,31 @@ def _poll_task(task_id, server_config, poll_rate=None, timeout=None):
     timer = threading.Timer(timeout, raise_task_timeout)
 
     # Poll until the task finishes. The timeout prevents an infinite loop.
+    path = '{0}/foreman_tasks/api/tasks/{1}'.format(server_config.url, task_id)
     try:
         timer.start()
-        path = '{0}/foreman_tasks/api/tasks/{1}'.format(
-            server_config.url,
-            task_id
-        )
         while True:
             response = client.get(path, **server_config.get_client_kwargs())
             response.raise_for_status()
             task_info = response.json()
-            if task_info['state'] != 'running':
+            if task_info['state'] in ('paused', 'stopped'):
                 break
             time.sleep(poll_rate)
     except KeyboardInterrupt:
         # raise_task_timeout will raise a KeyboardInterrupt when the timeout
         # expires. Catch the exception and raise TaskTimedOutError
-        raise TaskTimedOutError('Timed out polling task {0}'.format(task_id))
+        raise TaskTimedOutError(
+            'Timed out polling task {0}. Task information: {1}'
+            .format(task_id, task_info)
+        )
     finally:
         timer.cancel()
 
     # Check for task success or failure.
     if task_info['result'] != 'success':
         raise TaskFailedError(
-            'Task {0} completed with result {1}. Error message(s): {2}'.format(
-                task_id,
-                task_info['result'],
-                task_info['humanized']['errors']
-            )
+            'Task {0} did not succeed. Task information: {1}'
+            .format(task_id, task_info)
         )
     return task_info
 
--- a/tests/test_entity_mixins.py
+++ b/tests/test_entity_mixins.py
@@ -211,27 +211,46 @@ class MakeEntitiesFromIdsTestCase(TestCase):
 class PollTaskTestCase(TestCase):
     """Tests for :func:`nailgun.entity_mixins._poll_task`."""
 
-    def test__poll_task_failed_task(self):
-        """Test what happens when an asynchronous task completes and fails.
+    def setUp(self):
+        """Create a bogus server configuration object."""
+        self.cfg = config.ServerConfig('bogus url')
+
+    def test__poll_task_failure(self):
+        """What happens when a foreman task completes but does not succeed?
 
         Assert that a :class:`nailgun.entity_mixins.TaskFailedError` exception
         is raised.
 
         """
-        get_return = mock.Mock()
-        get_return.json.return_value = {
-            'state': 'not running',
-            'result': 'not success',
-            'humanized': {'errors': 'bogus error'},
-        }
-        with mock.patch.object(client, 'get') as client_get:
-            client_get.return_value = get_return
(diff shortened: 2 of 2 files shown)


nailgun/entity_mixins.py
```diff
@@ -82,34 +82,31 @@ def raise_task_timeout():
     timer = threading.Timer(timeout, raise_task_timeout)
 
     # Poll until the task finishes. The timeout prevents an infinite loop.
+    path = '{0}/foreman_tasks/api/tasks/{1}'.format(server_config.url, task_id)
     try:
         timer.start()
-        path = '{0}/foreman_tasks/api/tasks/{1}'.format(
-            server_config.url,
-            task_id
-        )
```
Ichimonji10: I've moved this out of the `try` block because I see no reason for it to be inside.

nailgun/entity_mixins.py
```diff
@@ -82,34 +82,31 @@ def raise_task_timeout():
...
         while True:
             response = client.get(path, **server_config.get_client_kwargs())
             response.raise_for_status()
             task_info = response.json()
-            if task_info['state'] != 'running':
+            if task_info['state'] in ('paused', 'stopped'):
```
Ichimonji10: This is the important bit.

nailgun/entity_mixins.py
```diff
@@ -82,34 +82,31 @@ def raise_task_timeout():
...
                 break
             time.sleep(poll_rate)
     except KeyboardInterrupt:
         # raise_task_timeout will raise a KeyboardInterrupt when the timeout
         # expires. Catch the exception and raise TaskTimedOutError
-        raise TaskTimedOutError('Timed out polling task {0}'.format(task_id))
+        raise TaskTimedOutError(
+            'Timed out polling task {0}. Task information: {1}'
+            .format(task_id, task_info)
+        )
```
Ichimonji10: I figure debugging may be easier if we include the entire `task_info` dict. This'll let us answer questions like "what state was the task in when polling timed out?"

nailgun/entity_mixins.py
```diff
@@ -82,34 +82,31 @@ def raise_task_timeout():
...
     # Check for task success or failure.
     if task_info['result'] != 'success':
         raise TaskFailedError(
-            'Task {0} completed with result {1}. Error message(s): {2}'.format(
-                task_id,
-                task_info['result'],
-                task_info['humanized']['errors']
-            )
+            'Task {0} did not succeed. Task information: {1}'
+            .format(task_id, task_info)
```
Ichimonji10: I cannot count on `['humanized']['errors']` being present, as the task may be in a state such as "paused". Easier and possibly more useful to just dump the entire dict of task information.

tests/test_entity_mixins.py
```diff
@@ -211,27 +211,46 @@ def test_pass_in_both(self):
...
+                with mock.patch.object(client, 'get') as get:
+                    get.return_value.json.return_value = {
+                        'state': state,
+                        'result': 'not success'
+                    }
+                    with self.assertRaises(entity_mixins.TaskFailedError):
+                        # pylint:disable=protected-access
+                        entity_mixins._poll_task(gen_integer(), self.cfg)
+
+    def test__poll_task_success(self):
```
Ichimonji10: I added this second test case just for kicks. Expands test coverage by one line.

sthirugn:
ACK

omaciel:
ACK

PR #227, opened 2015-10-07
[DONOTMERGE] Use only alpha characters for Host password field.

omaciel:
Our automation has shown that using UTF-8 characters for the password
field for a Host can be problematic. Defaulting to 'alpha' by default.

### DIFF
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -2022,7 +2022,10 @@ class Host(  # pylint:disable=too-many-instance-attributes
             'puppet_class': entity_fields.OneToManyField(PuppetClass),
             'puppet_proxy': entity_fields.OneToOneField(SmartProxy),
             'realm': entity_fields.OneToOneField(Realm),
-            'root_pass': entity_fields.StringField(length=(8, 30)),
+            'root_pass': entity_fields.StringField(
+                str_type='alpha',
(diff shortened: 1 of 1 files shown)


Ichimonji10:
Looks good, but I'm wondering what sort of error is raised by Satellite. Is the limitation intentional or a bug?

ACK pending question.

sthirugn:
ACK pending open question

nailgun/entities.py
```diff
@@ -2022,7 +2022,10 @@ def __init__(self, server_config=None, **kwargs):
             'puppet_class': entity_fields.OneToManyField(PuppetClass),
             'puppet_proxy': entity_fields.OneToOneField(SmartProxy),
             'realm': entity_fields.OneToOneField(Realm),
-            'root_pass': entity_fields.StringField(length=(8, 30)),
+            'root_pass': entity_fields.StringField(
+                str_type='alpha',
```
sthirugn: why alpha specifically, no alphanumeric allowed too?

omaciel:
May not have to do this after all...

omaciel:
I decided to close this as it may have been a red herring

Ichimonji10:
Thanks for following up on this, @omaciel.

PR #230, opened 2015-10-27
Add support for organization debug certificate

oshtaier:
Adding one more helper for organization entity debug certificate and also put verification for response content type as we can receive different contents from server

### DIFF
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -100,7 +100,9 @@ def _handle_response(response, server_config, synchronous=False):
     2. Poll the server for a foreman task to complete if an HTTP 202 (accepted)
        status code is returned and ``synchronous is True``.
     3. Immediately return if an HTTP "NO CONTENT" response is received.
-    4. Return the server's response, with all JSON decoded.
+    4. Determine what type of the content returned from server. Depending on
+       the type method should return server's response, with all JSON decoded
+       or just response content itself.
 
     :param response: A response object as returned by one of the functions in
         :mod:`nailgun.client` or the requests library.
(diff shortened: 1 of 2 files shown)


nailgun/entities.py
```diff
@@ -100,7 +100,9 @@ def _handle_response(response, server_config, synchronous=False):
     2. Poll the server for a foreman task to complete if an HTTP 202 (accepted)
        status code is returned and ``synchronous is True``.
     3. Immediately return if an HTTP "NO CONTENT" response is received.
-    4. Return the server's response, with all JSON decoded.
+    4. Determine what type of the content returned from server. Depending on
+    the type method should return server's response, with all JSON decoded or just
+    response content itself.
```
Ichimonji10: Try:

```
    4. Determine what type of the content returned from server. Depending on
       the type method should return server's response, with all JSON decoded or
       just response content itself.
```

nailgun/entities.py
```diff
@@ -113,7 +115,10 @@ def _handle_response(response, server_config, synchronous=False):
         return ForemanTask(server_config, id=response.json()['id']).poll()
     if response.status_code == NO_CONTENT:
         return
-    return response.json()
+    if response.headers.get('content-type', '').lower() == 'application/json':
+        return response.json()
+    else:
+        return response.content
```

Ichimonji10:
Looks sane. Have you successfully used the method?

oshtaier:
Yes, I got that certificate as a string and can operate with it in any way I like

nailgun/entities.py
oshtaier: Thanks :)

Ichimonji10:
ACK

Nice job, @oshtaier.

abalakh:
ACK

PR #238, opened 2015-11-17
Update list of OS families

oshtaier:
- Update list of possible OS families according to UI:
image

- Remove duplication of code for Media entity as it has same list of operation systems:
image

Verified that change works properly using `test_update_with_new_os` test from `api/test_partitiontable.py`

Corresponding update for robottelo part will be provided soon (there are no impact by any of that change, so we are free to merge current PR whenever we like)

### DIFF
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -72,10 +72,12 @@ else:  # pragma: no cover
 _FAKE_YUM_REPO = 'http://inecas.fedorapeople.org/fakerepos/zoo3/'
 _OPERATING_SYSTEMS = (
     'AIX',
+    'Altlinux',
     'Archlinux',
     'Debian',
     'Freebsd',
     'Gentoo',
+    'Junos'
     'Redhat',
     'Solaris',
     'Suse',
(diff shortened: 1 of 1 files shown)


abalakh:
ACK

elyezer:
@oshtaier do you know if there is an API path to get that information?

ACK

oshtaier:
In case of exception - I get next response:
```
Received HTTP 422 response: {
  "error": {"id":null,"errors":{"os_family":["must be one of [ AIX, Altlinux, Archlinux, Debian, Freebsd, Gentoo, Junos, Redhat, Solaris, Suse, Windows ]"]},"full_messages":["Operating system family must be one of [ AIX, Altlinux, Archlinux, Debian, Freebsd, Gentoo, Junos, Redhat, Solaris, Suse, Windows ]"]}
}
```
so I guess that we have that list somewhere, but I do not see direct request to the server in apidoc that will suffice our needs. Anyway, that fact can be investigated further

Ichimonji10:
:+1:

elyezer:
@oshtaier thanks for the information. We can take a look on it later.

PR #244, opened 2015-11-24
Override update_payload for operating system entity

oshtaier:
When it is not mandatory to wrap every attribute for update procedure, there is still some (like `family`) that need it, so overriding `update_payload` method like we usually did in similar situations


Ichimonji10:
Looks good. Have you used the updated code?

oshtaier:
Of course :) I am going to submit PR with a lot of test cases for OS functionality tomorrow

Ichimonji10:
ACK

abalakh:
ACK

PR #251, opened 2016-01-18
Drop unneeded workaround

### LINKED ISSUE #250: Remove workaround for ElasticSearch issue when fetching an entity
Our automation has shown that when deleting an activation key we attempt to fetch the deleted activation key several times despite the `404` being returned. Speaking to @Ichimonji10 he pointed me to [Redmine #4638](http://projects.theforeman.org/issues/4638) and this section of [code](https://github.com/SatelliteQE/nailgun/blob/master/nailgun/entities.py#L231-L234).

Since the Redmine issue has been marked as resolved I think that we may remove the workaround now. At least, I think that upon deletion we should not attempt the workaround since I do expect the entity to be gone from the system.

Ichimonji10:
Katello issue 4638 [1] has been resolved. As a result, the workaround for that
bug in `ActivationKey.create_raw` is no longer needed. It is possible to create
an activation key without this workaround in place:

```python
>>> from nailgun import config, entities
>>> cfg = config.ServerConfig(…)
>>> act_key = entities.ActivationKey(cfg).create(create_missing=True)
>>> act_key.id
204
```

Fix #250.

[1] http://projects.theforeman.org/issues/4638

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -52,10 +52,6 @@ the challenges developers face:
   experienced developers may find the API hard to work with.
 * The Satellite 6 API is not consistent in its implementation. For example, see
   the "Payload Generation" section of `this blog post`_.
-* The Satellite 6 API contains bugs. For example, `Foreman bug #4638`_ describes
-  how it is impossible to definitively know whether a given activation key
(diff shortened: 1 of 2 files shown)


omaciel:
APT

sthirugn:
ACK

PR #259, opened 2016-01-27
Accept instance attributes in Host.create_missing()

abalakh:
As rules for determining what attributes are requred for host creation are complex and depend on attributes passed and their values, nailgun currently allows to create a bogus managed host only in case no host attributes were passed. 
This is problematic when, for example, user wants to create a host with a custom name, or for already existing organization. For custom name user is forced to do the following trick:
```python
host = entities.Host()
host.create_missing()
host.name = name
host = host.create(create_missing=False)
```
However, for already existing organization (or domain, architecture, etc.) there's no easy workaround as all the dependecies, created inside `create_missing()` will still use a new organization even if we change the host's organization.

The purpose of PR is to make the nailgun accept host attributes, which are required for bogust host creation. This will allow us to create bogust hosts with custom names, organizations, domains, etc. If some custom entities like domain/architecture/operatingsystem/media etc were passsed, `create_missing()` will make sure all the entities are properly linked and will add the missing associations if needed.
Previously mentioned example with custom name will change to trivial case like this:
```python
host = entities.Host(name='foo').create()
```
Custom organization/location could be used as usual:
```python
host = entities.Host(organization=org).create()
```
And for custom entities you can pass, for example, custom os and be sure it will be associated with host's medium, no matter if that medium was created by `create_missing()` or was already existing and passed to host as an attribute.

### DIFF
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -88,10 +88,6 @@ class APIResponseError(Exception):
     """Indicates an error if response returns unexpected result."""
 
 
-class HostCreateMissingError(Exception):
-    """Indicates that ``Host.create_missing`` was unable to execute."""
-
-
 def _handle_response(response, server_config, synchronous=False):
     """Handle a server's response in a typical fashion.
 
@@ -2049,11 +2045,8 @@ class Host(  # pylint:disable=too-many-instance-attributes
         attributes should be filled in are mildly complex, and it is hard to
         know which scenario a user is aiming for.
 
-        Populate the values necessary to create a bogus managed host. However,
-        _only_ do so if no instance attributes are present. Raise an exception
-        if any instance attributes are present. Assuming that this method
-        executes in full, the resultant dependency graph will look, in part,
-        like this::
+        Populate the values necessary to create a bogus managed host. The
+        resultant dependency graph will look, in part, like this::
 
                  .-> medium --------.
                  |-> architecture <-V-.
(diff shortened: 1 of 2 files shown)


abalakh:
This isn't a breaking change as in such implementation `create_missing()` will accept only those instance attributes, which are used in our bogust managed host creation flow and will raise an exception as before in case anything unexpected was passed (assuming if user passed some "extra" attributes, he uses different creation flow and needs different dependencies).

However, I'm thinking about complete removal of attributes check - there're a lot of flags for host which are harmless and could be used in our flow without any regression. For example, `enabled` flag - right now to set `enabled=False` user needs to do the following:
```python
host = entities.Host()
host.create_missing()
host.enabled = False
host = host.create(create_missing=False)
```
But this flag is not affecting our creation flow so it could be whitelisted to simplify it to `host = entities.Host(enabled=False).create()`. And there's a lot of such "harmless" attributes like host_parameters, ip, root_pass etc.
And even if something from different creation flows was passed (like host collection) - we still won't introduce any regression, as `create_missing()` is raising an exception in such cases, meaning the tests with such flows are definitely creating the host with `create_missing` attribute set to `False`.
Please share your thoughts on whether that attributes check should remain or can be dropped.

abalakh:
`[WIP]` tag as
- this change still needs to be approved
- open question regarding attributes check
- if previous 2 points get resolved, need to add some unit tests not to decrease the coverage

Ichimonji10:
First off - nice job laying out the problem at hand and possible solutions, @abalakh! The context you've given has really helped me sink back into NailGun and think about this problem.

The code you've written up looks nice. It's long, but that's as expected. The rules for creating a host *are* complicated.

The question that `Host.create_missing()` has to contend with is "how can I generate missing attributes without clobbering what the user has provided?" My answer to that question was "don't even try to generate attributes if the user has given any," and that's why I added in the `HostCreateMissingError` exception. You've done some good work and provided a different answer: "let's be smart about generating attributes if the user has already given some (or raise a `HostCreateMissingError` if the user has given certain attributes)."

With that in mind, I do have one substantial suggestion: let's completely drop the `HostCreateMissingError`. That would make this method easier to reason about. It would change the answer to "let's be smart about generating attributes if the user has already given some (end of story)."

nailgun/entities.py
```diff
@@ -2034,6 +2034,7 @@ def __init__(self, server_config=None, **kwargs):
         self._meta = {'api_path': 'api/v2/hosts', 'server_modes': ('sat')}
         super(Host, self).__init__(server_config, **kwargs)
 
+    # pylint:disable=no-member,too-many-branches,too-many-statements
```
Ichimonji10: Can you move this statement down into the body of the method?

```python
def create_missing(self):
    """docstring"""
    # pylint:disable=no-member,too-many-branches,too-many-statements
```

Or, if that doesn't work, try this?

```python
def create_missing(self): # pylint:disable=no-member,too-many-branches,too-many-statements
    """docstring"""
```

As is, this pylint directive applies not only to this method, but every other method that is or might be added to the `Host` class.

abalakh: :+1:

abalakh:
@Ichimonji10 Agree with you.
I've updated the PR, completely removed `HostCreateMissingError` and added unit test. PR is ready for review :)

Ichimonji10:
Nice job. ACK

omaciel:
ACK

Issue #265, opened 2016-03-14
Update attributes for HostCollection entity


omaciel:
New version for the **Host Collection** API shows that some attributes have changed:

```shell
2016-03-11 18:31:28 - nailgun.client - DEBUG - Making HTTP POST request to https://<EDITED>/katello/api/v2/host_collections with options {'verify': False, 'auth': ('admin', 'changeme'), 'headers': {'content-type': 'application/json'}} and data {"organization_id": 888, "name": "\ubb8d\ud87e\udd5d\ud862\udd9c"}.
2016-03-11 18:31:28 - nailgun.client - DEBUG - Received HTTP 200 response:   {"name":"뮍𥪧𨦜","organization_id":888,"max_hosts":null,"description":null,"total_hosts":0,"unlimited_hosts":true,"created_at":"2016-03-11 23:31:28 UTC","updated_at":"2016-03-11 23:31:28 UTC","id":14,"permissions":{"deletable":true,"editable":true}}
```

Namely, ``max_content_hosts`` is now ``max_hosts`` and ``unlimited_content_hosts`` is ``unlimited_hosts``.

When fixing this issue, it will be important to consider branching strategies to support 6.2 and 6.1 APIs.

abalakh:
Addressed in SatelliteQE/robottelo#3344 and #266, closing this.

PR #266, opened 2016-03-14
Updating attributes for 'Host Collection' entity.

omaciel:
The API has changed for 'Host Collection' and some attributes have
been renamed, namely `max_content_hosts` has changed to `max_hosts`
and `unlimited_content_host` to `unlimited_host`.

### DIFF
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -1853,15 +1853,23 @@ class HostCollection(
     def __init__(self, server_config=None, **kwargs):
         self._fields = {
             'description': entity_fields.StringField(),
-            'max_content_hosts': entity_fields.IntegerField(),
+            'max_hosts': entity_fields.IntegerField(),
             'name': entity_fields.StringField(required=True),
             'organization': entity_fields.OneToOneField(
                 Organization,
                 required=True,
             ),
             'system': entity_fields.OneToManyField(System),
-            'unlimited_content_hosts': entity_fields.BooleanField(),
+            'unlimited_hosts': entity_fields.BooleanField(),
         }
+        # The following attributes have been renamed with Satellite
+        # 6.2, so we revert them to their old values if we have an
+        # older version of Satellite
+        if _get_version(server_config) < Version('6.2'):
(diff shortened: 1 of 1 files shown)


omaciel:
This should resolve #265

abalakh:
I would suggest solving it in the opposite way - `max_hosts` & `unlimited_host` in entity fields and check for sat version <6.2 where fields are updated to `max_content_hosts` & `unlimited_content_host`.
My point is about full support of the latest version and adding workarounds for outdated things rather than support outdated fields and add a workaround for latest supported sat version :)

omaciel:
@abalakh done :)

oshtaier:
ACK

nailgun/entities.py
```diff
@@ -1853,15 +1853,23 @@ class HostCollection(
...
                 required=True,
             ),
             'system': entity_fields.OneToManyField(System),
-            'unlimited_content_hosts': entity_fields.BooleanField(),
+            'unlimited_hosts': entity_fields.BooleanField(),
         }
+        # The following attributes have been renamed with Satellite
+        # 6.2, so we revert them to their old values if we have an
+        # older version of Satellite
+        if _get_version(server_config) <= Version('6.1'):
```
abalakh:
```python
>>> from packaging.version import parse
>>> parse('6.1.7') <= parse('6.1')
False
>>> parse('6.1.7') < parse('6.2')
True
```
`6.1.7` is obviously bigger than `6.1` so the check won't work, please, change this line to
```python
        if _get_version(server_config) < Version('6.2'):
```

abalakh:
Pending comment

nailgun/entities.py
omaciel: Nice catch!

omaciel:
@abalakh comment addressed

abalakh:
In ideal world we should also add a unit test to validate the changes and not to decrease the coverage, but this thing is blocking us and the test can be added later.
ACK

sthirugn:
perhaps a new issue needs to be added to add test coverage?

Issue #270, opened 2016-04-06
Deprecate the use of `entities.System` in favor of `entities.Host`


omaciel:
As per this [issue](http://projects.theforeman.org/issues/13904) it seems that we should be using `entities.Host` whenever we used to use `entities.System`. This needs to be fully investigated.

omaciel:
We should also update `Robottelo` existing tests to make sure that it uses `entities.Hosts`.

JacobCallahan:
specifically the targeted api paths are below. the systems path still exists, but will hopefully be removed in a later version.

```
"host_subscriptions": {
      "List a host's subscriptions": "/api/hosts/:host_id/subscriptions",
      "Trigger an auto-attach of subscriptions": "/api/hosts/:host_id/subscriptions/auto_attach",
      "List subscription events for the host": "/api/hosts/:host_id/subscriptions/events",
      "Add a subscription to a host": "/api/hosts/:host_id/subscriptions/add_subscriptions",
      "Set content overrides for the host": "/api/hosts/:host_id/subscriptions/content_override",
      "Get content and overrides for the host": "/api/hosts/:host_id/subscriptions/product_content"
    },
```

oshtaier:
@JacobCallahan 
It doesn't seem that API is still exist
HTTPError: 500 Server Error: Internal Server Error for url: https://server/katello/api/v2/systems

and basically https://github.com/Katello/katello/pull/5831/files shows that System API is going to be removed completely

JacobCallahan:
@oshtaier oh nice it looks like they finally removed the path. Unfortunately, it is still listed in the GET to /api/v2/

elyezer:
@JacobCallahan don't believe on that docs, since it says docker registries are there and they are not. It raise a 404 when trying to access any of the registry paths.

JacobCallahan:
@elyezer haha ok :). I hope there is already a bug open for that.

elyezer:
@JacobCallahan yes, there is.

abalakh:
robottelo part is done in SatelliteQE/robottelo#3384 and SatelliteQE/robottelo#3391.
Regarding nailgun - 99% is done in #276 and #277 , 1 more small fix is required though:
* Un-ignore host attribute for HostCollection entity as soon as BZ1325989 is fixed (simply drop `HostCollection.read()`)

So... Waiting for BZ1325989 so we can finally close this issue :)

abalakh:
[...]

PR #276, opened 2016-04-12
Deprecate the use of `entities.System` in favor of `entities.Host`

abalakh:
Part of #270

### DIFF
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -1858,13 +1858,13 @@ class HostCollection(
     def __init__(self, server_config=None, **kwargs):
         self._fields = {
             'description': entity_fields.StringField(),
+            'host': entity_fields.OneToManyField(Host),
             'max_hosts': entity_fields.IntegerField(),
             'name': entity_fields.StringField(required=True),
             'organization': entity_fields.OneToOneField(
                 Organization,
                 required=True,
             ),
-            'system': entity_fields.OneToManyField(System),
             'unlimited_hosts': entity_fields.BooleanField(),
         }
         # The following attributes have been renamed with Satellite
@@ -1874,6 +1874,8 @@ class HostCollection(
             self._fields['max_content_hosts'] = self._fields.pop('max_hosts')
             self._fields['unlimited_content_hosts'] = self._fields.pop(
                 'unlimited_hosts')
+            self._fields['system'] = entity_fields.OneToManyField(System)
+            self._fields.pop('host')
 
         self._meta = {
             'api_path': 'katello/api/v2/host_collections',
@@ -1887,6 +1889,18 @@ class HostCollection(
         if 'system_ids' in payload:
             payload['system_uuids'] = payload.pop('system_ids')
         return payload
+
+    def read(self, entity=None, attrs=None, ignore=None):
+        """Ignore 'host' field as it is not returned by the server.
+
+        For more information, see `Bugzilla #1325989
+        <https://bugzilla.redhat.com/show_bug.cgi?id=1325989>`_.
+
+        """
+        if ignore is None:
+            ignore = set()
+        ignore.add('host')
+        return super(HostCollection, self).read(entity, attrs, ignore)
 
     def update_payload(self, fields=None):
         """Rename ``system_ids`` to ``system_uuids``."""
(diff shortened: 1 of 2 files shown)


abalakh:
`[DONOTMERGE]` tag as i'd like to add some unit tests not to reduce coverage. Will work on them when receive all the comments/acks regarding this PR.

nailgun/entities.py
```diff
@@ -1888,6 +1890,18 @@ def create_payload(self):
...
         return payload
 
+    def read(self, entity=None, attrs=None, ignore=None):
+        """Ignore 'host' field as it is not returned by the server.
+
+        For more information, see `Bugzilla #1325989
+        <https://bugzilla.redhat.com/show_bug.cgi?id=1325989>`_.
+
+        """
+        if ignore is None:
```
elyezer: Would not be required to copy the host attr?

nailgun/entities.py
```diff
@@ -2000,6 +2014,53 @@ def update_payload(self, fields=None):
         return {u'hostgroup': super(HostGroup, self).update_payload(fields)}
 
 
+class HostPackage(Entity):
+    """A representation of a Host Package entity."""
+
+    def __init__(self, server_config=None, **kwargs):
+        if _get_version(server_config) < Version('6.2'):
+            raise NotImplementedError(
```
elyezer: I saw this and it makes me wonder if would be good to add a deprecation warning when the user access the system entity on Satellite >= 6.2?

elyezer:
The changes looks good to me. ACK

nailgun/entities.py
abalakh: You mean [this](https://github.com/SatelliteQE/nailgun/pull/276/files/0f714442b656ac44bbb5da476bd8d407b35a0df1#diff-3822710a997c3f2f7c3e93e884af69caR4183) warning? :)

elyezer: That one! I just did not see that! Good job

nailgun/entities.py
abalakh: I don't need host attrs here, i'm just passing received attrs to super with no changes. Purpose of this `read()` is to add 1 field to ignore and that's it. Anyway, that's a temporary solution - mentioned BZ should be fixed pretty soon and whole `read()` will be dropped.

abalakh:
OK, covered 12 of 13 missed lines. The only thing that wasn't covered - `HostCollection.read()`. I skipped it on purpose - that `read()` is temporary workaround which will be dropped really soon.

abalakh:
Hm... On the other hand, it takes no extra efforts to cover it and drop corresponding test lines together with `HostCollection.read()` when time comes. Will update the PR in a moment

elyezer:
@abalakh :+1:

oshtaier:
So what was final decision about `System` and `cvs`?
```
system = entities.System(
    content_view=content_view,
    environment=lc_env,
    organization=org,
).create()
```

abalakh:
No final decision here, that's why i put a comment that this PR is a `Part of #270`, not `Closes #270` :) Planned to work on that right after this PR, but i just took a brief look as you reminded and it seems as a simple API bug about missing `content_view_id` attribute for me: we had this ability for system path, we have it via WebUI for hosts too, but for some reason content views are not accepted via API for hosts. So I'll submit a bug and we'll see how it's going.

abalakh:
To make it more clear:
To completely switch from System entity to Host entity we need to:
Nailgun:
* This PR, which deprecates system, adds host attribute where it's needed, etc, etc,
* Un-ignore host attribute for HostCollection entity as soon as BZ1325989 is fixed
* Add content_view field to Host entity as soon as it's added to Satellite

Robottelo:
* Fix HostCollection tests (already done, corresponding PR is in place)
* Fix 2 tests with System entity and content views specified as soon as BZ### is fixed
* Fix robottelo's multiple_path tests
* Fix API_PATHS in test_api_smoke.py

Basically, the biggest ones are the first for nailgun and robottelo, which are already done, all the rest are simple 1-3 lines fixes which are very easy to do, but we need corresponding bugs to be closed first.

oshtaier:
ACK

PR #281, opened 2016-05-02
Let pylint use all CPU cores

Ichimonji10:
The `lint` make targets calls flake8 and pylint. By default, pylint uses
one CPU core. Update the make file and tell pylint to use all available
CPU cores. On the author's anemic quad-core system, `git clean -dfx &&
time make lint` reports the following times (in seconds) without this
patch:

1. 31
2. 31
3. 31
4. 31
5. 31

…and with this patch:

1. 19
2. 19
3. 18
4. 19
5. 19

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,5 @@
 TEST_OPTIONS=-m unittest discover --start-directory tests --top-level-directory .
+CPU_COUNT=$(shell python -c "from multiprocessing import cpu_count; print(cpu_count())")
 
 help:
 	@echo "Please use \`make <target>' where <target> is one of:"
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,4 +1,5 @@
 TEST_OPTIONS=-m unittest discover --start-directory tests --top-level-directory .
+CPU_COUNT=$(shell python -c "from multiprocessing import cpu_count; print(cpu_count())")
```
elyezer: This will fail if the user have Python 2.x installed.

elyezer:
ACK, pending comment.

Makefile
Ichimonji10: Hunh?

```
$ python2 --version
Python 2.7.11
$ python2 -c "from multiprocessing import cpu_count; print(cpu_count())"
4
```

elyezer: Oo so there is no need to import the print function from __future__?

Interesting :100:

Ichimonji10: Ahhh, I see what your concern is. :smile: Interestingly, this is a case where Python 2 is forward compatible with Python 3, but not the reverse.

sthirugn:
ACK

PR #289, opened 2016-05-19
Setting default type and length for org names

sthirugn:
It is becoming increasingly difficult to troubleshoot satellite after robottelo runs because org names are created in non-english characters which is not so user friendly to read.  Also these create some complex rendering issues in satellite UI.

In today's auto standup meeting, the team agreed to create default standard english names for satellite orgs. This PR fixes the same by restricting the org name's type and length if they are not explicitly specified.


sthirugn:
```
In [3]: from nailgun import entities
In [4]: org = entities.Organization().create()
...
...
In [5]: org.name
Out[5]: u'NMyoeJad'
```

sthirugn:
@rochacbruno

sthirugn:
Addresses part of https://github.com/SatelliteQE/robottelo/issues/2521

sghai:
Thanks to who all are agreed for this change. It was much needed change since a long time and I'm sure this will make life easy to fix some of UI issues.

sghai:
Ideally we should create all entities with standard english or readable names (with limited length). And covers other non-english names in separate tests. This would also be a helpful step in debugging.

rochacbruno:
completely agree with that!

ACK

PR #291, opened 2016-05-20
Introduces standard default english names for entities

### LINKED ISSUE #290: Fix default names for nailgun created entities
Sometimes nailgun created entities with default names - create rendering issues in Satellite.  One can argue that this needs to be fixed in robottelo.  But robottelo also has data validation tests for each entity separately.

Nailgun needs to be fixed to provide standard english names if the caller did not provide a name for an entity.  It is the responsibility of robottelo to test entities with different character set names and boundary values.
dockerrepositorytestcase-test_positive_create_repos_using_multiple_products-screenshot-2016-05-19_11_49_28

sthirugn:
- The default names will be used only when the caller did not provide the entity
name
- Fixes #290, https://github.com/SatelliteQE/robottelo/issues/2521

### DIFF
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -207,7 +207,11 @@ class ActivationKey(
             'environment': entity_fields.OneToOneField(LifecycleEnvironment),
             'host_collection': entity_fields.OneToManyField(HostCollection),
             'max_hosts': entity_fields.IntegerField(),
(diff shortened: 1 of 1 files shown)


sthirugn:
Thanks for @elyezer and @rplevka for a technical discussion in arriving at the default length range of 6 to 12.

rplevka:
this looks great,
ACK

PR #292, opened 2016-05-23
Introducing signals in nailgun 0.27.0

rochacbruno:
- Add blinker as an optional requirement, if not provided signals will be ignored unless there is a handler trying to connect, in that case it will raise an exception informing that blinker needs to be installed.
- Add documentation for signals module
- Add test_signals to test only that all signals are properly emitted

```python
from nailgun import entities, signals

@signals.pre_create.connect
def pre_create_handler(sender, create_missing):
    if isinstance(sender, entities.Organization):
        print("Org will be created, lets assign its domain")
        sender.domain = "http://example.com"

@signals.post_create.connect
def post_create_handler(sender, entity):
    if isinstance(entity, entities.Organization):
        print("Org {e.id} was created with domain {e.domain}!!!".format(e=entity))


>>> org = Organization(cfg, id=1).create()
>>> 'Org will be created, lets assign its domain' 
>>> 'Org 1 was created with domain http://example.com!!!'
```


tests/test_signals.py
```diff
@@ -0,0 +1,155 @@
...
+if version_info < (3, 4):
+    from unittest2 import TestCase
+else:
+    from unittest import TestCase
+
+ORG_DATA = {"id": 1, "domain": "http://example.com"}
+
+
+class FakeResponse(object):
+    '''Faking here cause we only need to test signals'''
```
elyezer: We use `"` for docstring, can you please follow the code style? Same applies for other docstrings.

tests/test_signals.py
```diff
@@ -0,0 +1,155 @@
...
+from sys import version_info
+from mock import patch
+from nailgun import entities, config, signals, client
+
+if version_info < (3, 4):
+    from unittest2 import TestCase
+else:
+    from unittest import TestCase
+
+ORG_DATA = {"id": 1, "domain": "http://example.com"}
```
elyezer: We use single quote for string in general, please update the code.

tests/test_signals.py
elyezer: Why not use mock for the response?

elyezer:
It looks good to me. I really like the signals approach specially by making them optional.

requirements-dev.txt
```diff
@@ -19,3 +19,6 @@ twine
 
 # For code coverage
 coveralls
+
+# For signals
+blinker
```
elyezer: :+1:

tests/test_signals.py
rochacbruno: Done!

tests/test_signals.py
rochacbruno: Done!
Do you have an example of using mock for response making status_code returning 200 and json returning some valid data?

elyezer:
ACK, thank you @rochacbruno

nailgun/entity_mixins.py
```diff
@@ -839,7 +847,10 @@ def create(self, create_missing=None):
             on the current object.
 
         """
-        return self.read(attrs=self.create_json(create_missing))
+        signals.pre_create.send(self, create_missing=create_missing)
+        entity = self.read(attrs=self.create_json(create_missing))
+        signals.post_create.send(self, entity=entity)
+        return entity
```
sthirugn: Does Read action really need a signal?

nailgun/entity_mixins.py
```diff
@@ -1245,6 +1261,9 @@ def search(self, fields=None, query=None, filters=None):
         ]
         if filters is not None:
             entities = self.search_filter(entities, filters)
+
+        signals.post_search.send(self, entities=entities, fields=fields,
+                                 query=query, filters=filters)
```
sthirugn: Do you think of any scenario in which signals will be useful for a search operation?

sthirugn:
ACK pending questions

nailgun/entity_mixins.py
rochacbruno: This is not a **read** this is **create** take a look https://github.com/rochacbruno/nailgun/blob/add_signals/nailgun/entity_mixins.py#L824-L850 explanation on docstring (looks like **.read** actually creates with that attrs create_missing.

nailgun/entity_mixins.py
rochacbruno: I imagine a scenario where one would like to apply a filter to every search in batch,

```python
@signals.pre_search.connect
def add_filters_to_every_search(sender, fields, query, filters):
    filters.append(#filters_here)
```

With that any **.search** performed will automatically get that filters added.

The post_search may be useful to record logs, statistics etc

There is no overhead on emitting signals, but it looks like uncommon scenarios, if agreed I can remove the search signals.

```

sthirugn: np, we can leave it if harmless

sthirugn:
and pypi is now refreshed with version 0.27.0

PR #295, opened 2016-05-27
Using a decorator to emit signals "DRY"

rochacbruno:
I was annoyed by the amount of **.send()** calls we added and thinking on follow the DRY concept I turned all the **.send()** calls in to a decorator.

Now whenever we want to emit a signal we just need to decorate it.

> ~~That decorator can be turned in to an opensource python library itself as a blinker plugin, as I did not found any plugin or blinker feature that does that out of the box~~  https://github.com/SatelliteQE/blinker_herald

Instead of

```python
def create(self, ...):
     signals.pre_create.send(....)
     ...
     signals.post_create.send(...)
     return ...
```

Now we just need to to do

```python
@signals.emit()
def create(self, ....):
    ...
```

No need to change anything in method or function and the **pre** and **post** signals will be emitted.

> Checked the overhead and no significant impact added, actually it is taking less time now, maybe because decorator is pre-compiled in signals.py module and there is only one **send** call there.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,7 @@ docs-clean:
 	@cd docs; $(MAKE) clean
 
 lint:
-	flake8 .
+	flake8 --ignore=E731 .
 	pylint -j $(CPU_COUNT) --reports=n --disable=I --ignore-imports=y \
 		nailgun tests setup.py docs/conf.py
 	pylint -j $(CPU_COUNT) --reports=n --disable=I --ignore-imports=y --disable=similarities \
--- a/docs/signals.rst
+++ b/docs/signals.rst
@@ -15,6 +15,9 @@ Signals
 Overview
 --------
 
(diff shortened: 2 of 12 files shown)


omaciel:
Looks great @rochacbruno and a huge 👍 for the idea of turning this into a separate lib/module. Let me know what you want to call it and I can create a new repo for you :)

rochacbruno:
@omaciel I created a package called **blinker_herald** ready to push, waiting for the repo.

rochacbruno:
This is ready to merge as all the test_cases on https://github.com/SatelliteQE/robottelo/pull/3492 executed with success and also all the nailgun tests passed.

Ichimonji10:
Very cool, @rochacbruno. :+1:

sthirugn:
ACK

sthirugn:
Merging it now for the next run.

PR #301, opened 2016-06-22
Add DockerRegistryContainer entity

abalakh:
```python
>>> entities.DockerRegistryContainer(
...     compute_resource=215,
...     organization=[1784],
...     registry=38,
...     repository_name='rhel',
... ).create()
nailgun.entities.DockerRegistryContainer(nailgun.config.ServerConfig(url=u'https://example.com', verify=False, auth=(u'admin', u'changeme')), tty=None, attach_stdin=None, command=None, tag=None, registry=nailgun.entities.Registry(nailgun.config.ServerConfig(url=u'https://example.com', verify=False, auth=(u'admin', u'changeme')), id=38), id=36, name=u'hopeful_babbage', cpu_set=None, repository_name=u'rhel', attach_stderr=None, compute_resource=nailgun.entities.AbstractComputeResource(nailgun.config.ServerConfig(url=u'https://example.com', verify=False, auth=(u'admin', u'changeme')), id=215), entrypoint=None, location=[], memory=None, organization=[], attach_stdout=None, cpu_shares=None)
```

### DIFF
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -1041,10 +1041,7 @@ class AbstractDockerContainer(
             'attach_stderr': entity_fields.BooleanField(),
             'attach_stdin': entity_fields.BooleanField(),
             'attach_stdout': entity_fields.BooleanField(),
-            'command': entity_fields.StringField(
-                required=True,
-                str_type='latin1',
-            ),
+            'command': entity_fields.StringField(required=True, default='top'),
             'compute_resource': entity_fields.OneToOneField(
                 AbstractComputeResource,
                 required=True,
@@ -1169,6 +1166,24 @@ class DockerHubContainer(AbstractDockerContainer):
             'tag': entity_fields.StringField(required=True, default='latest'),
         }
         super(DockerHubContainer, self).__init__(server_config, **kwargs)
+
+
+class DockerRegistryContainer(AbstractDockerContainer):
+    """A docker container that comes from custom external registry.
+
+    .. WARNING:: The ``repository_name`` field references an image repository
+        on the custom registry, not a locally created
+        :class:`nailgun.entities.Repository`.
+
+    """
+
+    def __init__(self, server_config=None, **kwargs):
+        self._fields = {
+            'registry': entity_fields.OneToOneField(Registry, required=True),
+            'repository_name': entity_fields.StringField(required=True),
+            'tag': entity_fields.StringField(required=True, default='latest'),
+        }
+        super(DockerRegistryContainer, self).__init__(server_config, **kwargs)
 
 
 class ContentUpload(Entity):
--- a/tests/test_entities.py
+++ b/tests/test_entities.py
@@ -101,6 +101,7 @@ class InitTestCase(TestCase):
                 entities.DiscoveryRule,
                 entities.DockerComputeResource,
                 entities.DockerHubContainer,
+                entities.DockerRegistryContainer,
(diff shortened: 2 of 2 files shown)


nailgun/entities.py
```diff
@@ -1042,6 +1042,7 @@ def __init__(self, server_config=None, **kwargs):
             'attach_stdin': entity_fields.BooleanField(),
             'attach_stdout': entity_fields.BooleanField(),
             'command': entity_fields.StringField(
+                default='top',
```
abalakh: Bonus. at least 95% of our tests don't care what command will be executed in container and are focused on different things, so i can't see the reason to manually specify exact command each time, considering cli factory is using default value for command too.

nailgun/entities.py
```diff
@@ -1042,6 +1042,7 @@ def __init__(self, server_config=None, **kwargs):
...
                 required=True,
                 str_type='latin1',
```
abalakh: btw i have questions to this line - is it supposed to be `latin1`? server is returning 500 error each time when random latin1 string is sent as a command, and basically all our tests are defining custom command string (probably because of that).

oshtaier: Does server return 500 error for all other types when we generate random string of these types?

abalakh: Tried it out with 'utf8' - same result. But not for 'numeric', for example, it is working ok. So i guess it's working ok with default alphanumeric characters and is failing on special ones.
The thing is - if you pass some random command like '012a4754955fa33cdadd2109fd2f836ae46a15f010d2617ad0356b48b826d614' - creation will still fail with 500 error and following reason:
```
2016-06-23 13:42:08 - nailgun.client - WARNING - Received HTTP 500 response: {
  "error": {"message":"Cannot start container 012a4754955fa33cdadd2109fd2f836ae46a15f010d2617ad0356b48b826d614: [8] System error: exec: \"162817867131753\": executable file not found in $PATH\n"}
}
```
This basically means there's no point in generating random value for this field by default, we should use predefined one in `default=` parameter, as otherwise nothing will be created. Therefore, i'll drop this `str_type=` line.

oshtaier:
ACK

nailgun/entities.py
elyezer: When you say this you are looking just for Robottelo. Since this is something on Robottelo I think we can't assume that on NailGun for all users.

Also depending on the image top may not be available and if the user forget to fill the command it will use top automatically and will not enforce the user on providing a command.

I am ok on setting this default because almost images have the top, I remember having issues with some images though. I just left some thoughts related to the NailGun users, if others agree on merging with the default value I am ok.

elyezer:
ACK, left https://github.com/SatelliteQE/nailgun/pull/301#discussion_r68234598 as a thinking comment, feel free to merge if you are ok if the default value.

nailgun/entities.py
abalakh: Previously nailgun was sending just random string which would fail in 100% cases when command is not specified, so imo it's still a better option (and if user haven't specified some field it basically means he doesn't care what value will be provided and lets nailgun decide how to fill it, that's why we have `create_missing`). But i'm ok with defining any other command if you remember more widespread ones (pwd? cd? echo?)

elyezer: Using top is ok. Thanks for giving more context.

Let's merge and if needed in the future we update to a list of some commands to NailGun choose.

PR #306, opened 2016-08-18
Added 'download_policy' to `Repository` entities.

omaciel:
Adding newer `download_policy` field to the `Repository` entity. As per
official API documentation:

      download policy for yum repos (either 'immediate', 'on_demand', or
'background')

      Validations:

	Must be one of: immediate, on_demand, background.

### DIFF
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -3461,6 +3461,11 @@ class Repository(
             'docker_upstream_name': entity_fields.StringField(
                 default='busybox'
             ),
+            'download_policy': entity_fields.StringField(
+                choices=('background', 'immediate', 'on_demand'),
+                default='immediate',
+                required=True,
+            ),
             'gpg_key': entity_fields.OneToOneField(GPGKey),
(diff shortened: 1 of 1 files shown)


JacobCallahan:
ACK. Looks good! I propose that we change the default to on_demand after lazy sync drops in z-stream.

rochacbruno:
ACK

rochacbruno:
@JacobCallahan do we need to open an issue to not forget about changing the default to `on_demand`?

nailgun/entities.py
```diff
@@ -3461,6 +3461,11 @@ def __init__(self, server_config=None, **kwargs):
             'docker_upstream_name': entity_fields.StringField(
                 default='busybox'
             ),
+            'download_policy': entity_fields.StringField(
+                choices=('background', 'immediate', 'on_demand'),
+                default='immediate',
+                required=True,
```
sghai: This should be applicable for yum only repos.  May be we should set `required` flag only if `content_type == `yum``.

PR #309, opened 2016-08-19
Fixed download_policy required flag only if content_type='yum'

sghai:
The required flag for `download_policy` should only be set for `yum` type repos but not for others.

### DIFF
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -3464,7 +3464,6 @@ class Repository(
             'download_policy': entity_fields.StringField(
                 choices=('background', 'immediate', 'on_demand'),
                 default='immediate',
-                required=True,
             ),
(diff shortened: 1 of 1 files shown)


rplevka:
ACK
thanks for a prompt fix

omaciel:
ACK

omaciel:
By the way, it is interesting that the API docs don't mention that only `yum` supports `download_policy` 👎

rplevka:
@omaciel 
hmm, here's what i can see in apidoc - POST repository section on katello-nightly:
```
download_policy
optional 	

download policy for yum repos (either 'immediate', 'on_demand', or 'background')

Validations:

    Must be one of: immediate, on_demand, background.
```

- it is mentioning that it is applicable to yum repos

omaciel:
@rplevka I take it all back :)

PR #315, opened 2016-09-14
Add Smart Variables helpers

svtkachenko:
```python
hostgroup = entities.HostGroup().create()
hostgroup.add_puppetclass(data={'puppetclass_id': puppet.id})
hostgroup.list_smart_variables()

016-09-14 10:32:04 - nailgun.client - DEBUG - Making HTTP GET request to https://example.com/api/v2/hostgroups/14/smart_variables with options {'verify': False, 'auth': ('admin', 'changeme'), 'headers': {'content-type': 'application/json'}} and no data.
2016-09-14 10:32:05 - nailgun.client - DEBUG - Received HTTP 200 response: {
  "total": 190,
  "subtotal": 190,
  "page": 1,
  "per_page": 20,
  "search": null,
  "sort": {
    "by": null,
    "order": null
  },
  "results": [{"description":null,"parameter_type":null,"default_value":"example","hidden_value?":false,"hidden_value":"*****","validator_type":"list","validator_rule":"test, example, 30","override_value_order":"fqdn\nhostgroup\nos\ndomain","override_values_count":1,"merge_overrides":false,"merge_default":false,"avoid_duplicates":false,"puppetclass_id":2,"puppetclass_name":"ntp","created_at":"2016-09-13 16:08:30 UTC","updated_at":"2016-09-13 16:08:31 UTC","variable":"숖","id":180},
]
}
```

### DIFF
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -2196,11 +2196,15 @@ class HostGroup(
             /api/hostgroups/:hostgroup_id/puppetclass_ids
         smart_class_parameters
             /api/hostgroups/:hostgroup_id/smart_class_parameters
+        smart_class_variables
+            /api/hostgroups/:hostgroup_id/smart_variables
 
         Otherwise, call ``super``.
 
         """
-        if which in ('puppetclass_ids', 'smart_class_parameters'):
+        if which in ('puppetclass_ids',
+                     'smart_class_parameters',
+                     'smart_variables'):
             return '{0}/{1}'.format(
                 super(HostGroup, self).path(which='self'),
                 which
@@ -2209,6 +2213,9 @@ class HostGroup(
 
     def add_puppetclass(self, synchronous=True, **kwargs):
         """Add a Puppet class to host
+
+        Here is an example of how to use this method::
+            hostgroup.add_puppetclass(data={'puppetclass_id': puppet.id})
 
         :param synchronous: What should happen if the server returns an HTTP
             202 (accepted) status code? Wait for the task to complete if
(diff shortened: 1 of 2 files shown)


nailgun/entities.py
```diff
@@ -2210,6 +2212,9 @@ def path(self, which=None):
     def add_puppetclass(self, synchronous=True, **kwargs):
         """Add a Puppet class to host
 
+        Here is an example of how to use this method::
+            hostgroup.add_puppetclass(data={'puppetclass_id': puppet.id})
+
```
sghai: +1 to example

sghai:
ACK

nailgun/entities.py
```diff
@@ -2200,7 +2200,9 @@ def path(self, which=None):
         Otherwise, call ``super``.
 
         """
-        if which in ('puppetclass_ids', 'smart_class_parameters'):
+        if which in ('puppetclass_ids',
```
oshtaier: update the docstring

nailgun/entities.py
```diff
@@ -2559,7 +2581,7 @@ def path(self, which=None):
         Otherwise, call ``super``.
 
         """
-        if which in ('smart_class_parameters',):
+        if which in ('smart_class_parameters', 'smart_variables'):
```
oshtaier: update the docstring

oshtaier:
ACK pending comments

nailgun/entities.py
omaciel: Since this method is very specific to adding a puppet class, shouldn't we make it easier to use and instead of having a `**kwargs` argument, use a `class_id` argument? Then the method can pass `data={'puppetclass_id': class_id}` explicitly and using this method would be simpler: `hostgroup.add_puppetclass(my_puppet_class.id)`.

oshtaier: it is basically our standard signature that we have to use for all helpers

oshtaier: We even had a PR to make all helpers to use it

Ichimonji10: I think a benefit of using `**kwargs` is that katello can add/change/remove expected parameters, and the signature of this method doesn't need to be updated to reflect the change in katello's expectations. (Mind you, there's already plenty of tight coupling between nailgun and katello's API. But there's no need to add more coupling than necessary, right?)

svtkachenko:
Docstring updated.

oshtaier:
ACK

PR #317, opened 2016-09-28
Add puppet class helpers

svtkachenko:
```python
hostgroup.delete_puppetclass(data={'puppetclass_id': puppet.id})
DEBUG - Making HTTP DELETE request to https://example.com/api/v2/puppetclasses/429 with options {'verify': False, 'auth': ('admin', 'changeme'), 'headers': {'content-type': 'application/json'}} and no data.
DEBUG - Received HTTP 200 response: {"id":429,"name":"firewall","created_at":"2016-09-28T15:47:40.960Z","updated_at":"2016-09-28T15:47:40.960Z"}
```

### DIFF
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -2260,9 +2260,11 @@ class HostGroup(
         Otherwise, call ``super``.
 
         """
-        if which in ('puppetclass_ids',
-                     'smart_class_parameters',
-                     'smart_variables'):
+        if which in (
+                'puppetclass_ids',
+                'smart_class_parameters',
+                'smart_variables',
+        ):
             return '{0}/{1}'.format(
                 super(HostGroup, self).path(which='self'),
                 which
@@ -2270,7 +2272,7 @@ class HostGroup(
         return super(HostGroup, self).path(which)
 
     def add_puppetclass(self, synchronous=True, **kwargs):
-        """Add a Puppet class to host
+        """Add a Puppet class to host group
 
         Here is an example of how to use this method::
             hostgroup.add_puppetclass(data={'puppetclass_id': puppet.id})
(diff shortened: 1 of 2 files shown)


nailgun/entities.py
```diff
@@ -2289,6 +2289,33 @@ def add_puppetclass(self, synchronous=True, **kwargs):
...
+        :returns: The server's response, with all JSON decoded.
+        :raises: ``requests.exceptions.HTTPError`` If the server responds with
+            an HTTP 4XX or 5XX message.
+
+        """
+        kwargs = kwargs.copy()
+        kwargs.update(self._server_config.get_client_kwargs())
+        path = "{0}/{1}".format(
+            self.path('puppetclass_ids'),
+            kwargs['data'].pop('ppuppetclass_id')
```
omaciel: s/ppuppetclass_id/puppetclass_id?

svtkachenko: Indeed.

tests/test_entities.py
```diff
@@ -1313,6 +1316,7 @@ def setUpClass(cls):
                 entities.Organization(**generic).download_debug_certificate,
                 'get'
             ),
+            (entities.Host(**generic).add_puppetclass, 'post'),
```
omaciel: Do we need to add `delete_puppetclass` here as well?

omaciel:
@svtkachenko great job! I have added 2 comments which you will want to address before we get this merged.

nailgun/entities.py
```diff
@@ -2289,6 +2289,33 @@ def add_puppetclass(self, synchronous=True, **kwargs):
...
+        :returns: The server's response, with all JSON decoded.
+        :raises: ``requests.exceptions.HTTPError`` If the server responds with
+            an HTTP 4XX or 5XX message.
+
+        """
+        kwargs = kwargs.copy()
+        kwargs.update(self._server_config.get_client_kwargs())
+        path = "{0}/{1}".format(
+            self.path('puppetclass_ids'),
+            kwargs['data'].pop('puppetclass_id')
```

A heads up to the other reviewers: the call to `pop()` here is intentional. Satellite expects a puppetclass ID as part of the request path, but not as part of the request body. Another approach would be to change the signature of this method to `delete_puppetclass(self, puppetclass_id, synchronous=True, *kwargs)`. That would work, but it would make this method different from nearly all the other custom methods, which accept arguments of `(self, synchronous=True, **kwargs)`.

Ichimonji10:
This looks nice. I like the change, but will defer to current users like @rplevka, @rochacbruno, etc.

tests/test_entities.py
omaciel: @Ichimonji10 any suggestion for how we may be able to handle this?

Ichimonji10: `GenericTestCase` is designed for methods that don't require any extra parameters. For full coverage, a new test case would need to be written.

> Maybe it requires separate test but it would be full copy of test_generic with single line changed :(

This test method would need slightly more than one line of changes. The `self.assertEqual(client_request.call_args[1], kwargs)` assertion would need to be modified or dropped, as would the line that sets `kwargs`.

svtkachenko:
Update: missing tests added, typo fixed.

omaciel:
@rplevka @rochacbruno mind reviewing it? LGTM

rochacbruno:
ACK! LGTM

@rplevka it is up to you

rplevka:
@svtkachenko, perhaps sanity tests for `add_puppetclass` would be nice to have?
not a blocker.

ACK pending question

svtkachenko:
@rplevka It is covered in `test_generic`.

rplevka:
ACK

PR #321, opened 2016-10-03
Add Package entity

ehelms:


nailgun/entities.py
```diff
@@ -3811,6 +3811,32 @@ def list_smart_variables(self, synchronous=True, **kwargs):
         return _handle_response(response, self._server_config, synchronous)
 
 
+class Package(Entity, EntityReadMixin, EntitySearchMixin):
+    """A representation of a Package entity."""
+
+    def __init__(self, server_config=None, **kwargs):
+        self._fields = {
+            'arch': entity_fields.StringField(),
+            'checksum': entity_fields.StringField(),
```
omaciel: Should we add the valid choices for this field?

nailgun/entities.py
```diff
@@ -3811,6 +3811,32 @@ def list_smart_variables(self, synchronous=True, **kwargs):
...
+            'description': entity_fields.StringField(),
+            'epoch': entity_fields.StringField(),
+            'filename': entity_fields.StringField(),
+            'name': entity_fields.StringField(
```
omaciel: So only `name` is a required field?

nailgun/entities.py
```diff
@@ -3811,6 +3811,32 @@ def list_smart_variables(self, synchronous=True, **kwargs):
...
+                str_type='alpha',
+                length=(6, 12)
```
svtkachenko: I think `lenth` is used when creating entity, for read-only entities there is no use in this.

nailgun/entities.py
ehelms: What are the valid choices? Checksums are giant hex strings.

nailgun/entities.py
ehelms: Actually, I don't think you can lookup by name only search by it, so I guess I should remove this? I added fields based on what the API returns for a package vs. the parameters you can pass TO the API. Which should this set of attributes be?

omaciel: `name` would be populated for you automatically on `create`, so I was wondering what are the minimal attributes one must specify in order to create a package?

ehelms:
One does not simply create a package. They are read only, since they come
from sync or upload to a repo.

PR #347, opened 2016-12-23
Add mixins to Image entity

svtkachenko:


oshtaier:
Add unit tests for mixins

svtkachenko:
@oshtaier Looking into existing tests I think that they are used for entities that override methods from mixins.

oshtaier:
I would check these two places:
https://github.com/SatelliteQE/nailgun/blob/master/tests/test_entities.py#L482
https://github.com/SatelliteQE/nailgun/blob/master/tests/test_entities.py#L1142

svtkachenko:
Yeap, checked that before writing previous comment.
`CreateTestCase`: the list of added entities is much smaller than entire list of entities, all added entities except `ConfigGroup` have overridden `create` or `create_payload` method.
`UpdateTestCase`: the list of added entities is much smaller than entire list of entities, all added entities have overridden `update` or `update_payload` method.

It is not so difficult to add another entity to the tests thanks to its good design, however bloated code is prone to errors so better keep it compacted.
Also @coveralls says that `Coverage remained the same`.

PR #350, opened 2017-01-03
Implemented Entity __eq__ method comparing all fields

### LINKED ISSUE #335: Make entities __eq__ comparable
As soon as #326 got merged, having proper json generation we should implement the `__eq__` method in `Entity` mixin making entities fully comparable by equality of its fields.

Example:

```python
>>> Host(id=1) == Host(id=1)
True
```

The `__eq__` should serialize the entity to JSON, ensure it is `read` again from server and then compare all fields and values, only if all `_fields` has same values it should return True.

That helps with unit tests

renzon:
close #335

### DIFF
--- a/nailgun/entity_mixins.py
+++ b/nailgun/entity_mixins.py
@@ -555,7 +555,7 @@ class Entity(object):
                 value = values[field_name]
                 if value is None:
                     json_dct[field_name] = None
-                if isinstance(field, OneToOneField):
+                elif isinstance(field, OneToOneField):
                     json_dct[field_name] = value.to_json_dict()
                 elif isinstance(field, OneToManyField):
                     json_dct[field_name] = [
--- a/nailgun/entity_mixins.py
+++ b/nailgun/entity_mixins.py
@@ -83,6 +83,7 @@ def _poll_task(task_id, server_config, poll_rate=None, timeout=None):
     def raise_task_timeout():  # pragma: no cover
         """Raise a KeyboardInterrupt exception in the main thread."""
         thread.interrupt_main()
+
     timer = threading.Timer(timeout, raise_task_timeout)
 
     # Poll until the task finishes. The timeout prevents an infinite loop.
@@ -551,16 +552,29 @@ class Entity(object):
         json_dct = {}
         for field_name, field in fields.items():
             if field_name in values:
+                value = values[field_name]
+                if value is None:
+                    json_dct[field_name] = None
                 if isinstance(field, OneToOneField):
-                    json_dct[field_name] = values[field_name].to_json_dict()
+                    json_dct[field_name] = value.to_json_dict()
                 elif isinstance(field, OneToManyField):
                     json_dct[field_name] = [
-                        entity.to_json_dict() for entity in values[field_name]
+                        entity.to_json_dict() for entity in value
                     ]
                 else:
-                    json_dct[field_name] = to_json_serializable(
-                        values[field_name])
+                    json_dct[field_name] = to_json_serializable(value)
         return json_dct
+
+    def __eq__(self, other):
+        """Compare two entities based on their properties. Even nested
+        objects are considered for equality
+
+        :param other: entity to compare self to
+        :return: boolean indicating if entities are equal or not
+        """
+        if other is None:
+            return False
+        return self.to_json_dict() == other.to_json_dict()
 
 
 class EntityDeleteMixin(object):
--- a/nailgun/entity_mixins.py
+++ b/nailgun/entity_mixins.py
@@ -555,6 +555,10 @@ class Entity(object):
                 value = values[field_name]
                 if value is None:
                     json_dct[field_name] = None
+                    # This conditions is needed because some times you get
+                    # None on an OneToOneField what lead to an error
+                    # on bellow condition, e.g., calling value.to_json_dict()
+                    # when value is None
                 elif isinstance(field, OneToOneField):
                     json_dct[field_name] = value.to_json_dict()
                 elif isinstance(field, OneToManyField):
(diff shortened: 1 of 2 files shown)


nailgun/entity_mixins.py
```diff
@@ -551,17 +552,30 @@ def to_json_dict(self):
         json_dct = {}
         for field_name, field in fields.items():
             if field_name in values:
+                value = values[field_name]
+                if value is None:
+                    json_dct[field_name] = None
```
rochacbruno: coveralls says this line misses test coverage

nailgun/entity_mixins.py
```diff
@@ -551,17 +552,30 @@ def to_json_dict(self):
...
 
+    def __eq__(self, other):
+        """Compare two entities based on their properties. Even nested
+        objects are considered for equality
+
+        :param other: entity to compare self to
+        :return: boolean indicating if entities are equal or not
+        """
+        if other is None:
+            return False
```
rochacbruno: This line also missing test coverage

renzon:
@rochacbruno coverage improved.

renzon:
For the one who is going to merge: please use "Squash and Merge" Once I fixed code on different commit.

nailgun/entity_mixins.py
```diff
@@ -551,17 +552,30 @@ def to_json_dict(self):
...
+                    json_dct[field_name] = value.to_json_dict()
                 elif isinstance(field, OneToManyField):
                     json_dct[field_name] = [
-                        entity.to_json_dict() for entity in values[field_name]
+                        entity.to_json_dict() for entity in value
                     ]
                 else:
-                    json_dct[field_name] = to_json_serializable(
-                        values[field_name])
+                    json_dct[field_name] = to_json_serializable(value)
```
elyezer: I did the following test:

```python
In [1]: from nailgun.entity_mixins import to_json_serializable

In [2]: to_json_serializable(None)

In [3]: print(to_json_serializable(None))
None
```

There is no need to treat the value is None since the else clause will already take care of it.

tests/test_entity_mixins.py
```diff
@@ -344,6 +364,70 @@ def test_bad_value_error(self):
...
             SampleEntityTwo(self.cfg, one_to_many=1)
 
+    def test_eq_none(self):
+        """Test method ``nailgun.entity_mixins.Entity.__eq__`` against None
+
+        Assert that ``__eq__`` returns False when compared to None.
+
+        """
+        alice = SampleEntity(self.cfg, id=1, name='Alice')
+        self.assertNotEquals(alice, None)
```
elyezer: Use `assertNotEqual` (without the `s` at the end) since `assertNotEquals` is deprecated and will produce a warning:

```
In [7]: t.assertNotEquals
Out[7]: <bound method TestCase._deprecate.<locals>.deprecated_func of <__main__.Test testMethod=runTest>>

In [9]: t.assertNotEquals??
Signature: t.assertNotEquals(*args, **kwargs)
Source:   
        def deprecated_func(*args, **kwargs):
            warnings.warn(
                'Please use {0} instead.'.format(original_func.__name__),
                DeprecationWarning, 2)
            return original_func(*args, **kwargs)
File:      /usr/lib64/python3.5/unittest/case.py
Type:      method
```

elyezer: Same applies for `assertEquals` which should be `assertEqual`

nailgun/entity_mixins.py
renzon: At first I was doing it. But discovered a case during test on robottelo where an OneToOne is None and. 'json_dct[field_name] = value.to_json_dict()' lead to error because to_json_dict() is been called on a None value.

renzon: Finnaly found the case I mentioned above:

```python
hostgroup = entities.HostGroup(
            location=[self.loc],
            organization=[self.org],
        ).create()
        hostgroup.to_json_dict()
```
Stack trace:

```console
    def to_json_dict(self):
        """Create a dct with Entity properties for json encoding.
            It can be overridden by subclasses for each standard serialization
            doesn't work. By default it call _to_json_dict on OneToOne fields
            and build a list calling the same method on each object on OneToMany
            fields.
    
            :return: dct
            """
        fields, values = self.get_fields(), self.get_values()
        json_dct = {}
        for field_name, field in fields.items():
            if field_name in values:
                if isinstance(field, OneToOneField):
>                   json_dct[field_name] = values[field_name].to_json_dict()
E                   AttributeError: 'NoneType' object has no attribute 'to_json_dict'
```

elyezer: I see, do you mind adding a comment on that if clause? It may happen that someone else realize that the `json_dct[field_name] = to_json_serializable(value)` works for None and is not aware of that specific case.

renzon: @elyezer: done!

tests/test_entity_mixins.py
```diff
@@ -344,6 +364,70 @@ def test_bad_value_error(self):
...
             SampleEntityTwo(self.cfg, one_to_many=1)
 
+    def test_eq_none(self):
+        """Test method ``nailgun.entity_mixins.Entity.__eq__`` against None
+
+        Assert that ``__eq__`` returns False when compared to None.
+
+        """
+        alice = SampleEntity(self.cfg, id=1, name='Alice')
+        self.assertFalse(alice == None)
```
rochacbruno: Comparison to None should be `alice is None` instead of `==`

elyezer: Good catch, better to use `self.assertIsNone(alice)`

renzon: The point is that I want __eq__ to be called with None ;)

renzon:
@rochacbruno take a look on the change

PR #353, opened 2017-01-16
[6.2.z] Implemented Entity __eq__ method comparing all fields (#350)

renzon:
(cherry picked from commit ecf91be)

### DIFF
--- a/nailgun/entity_mixins.py
+++ b/nailgun/entity_mixins.py
@@ -1,10 +1,14 @@
 # -*- encoding: utf-8 -*-
 """Defines a set of mixins that provide tools for interacting with entities."""
+import json as std_json
 from collections import Iterable
+from datetime import date, datetime
+
 from fauxfactory import gen_choice
 from inflection import pluralize
 from nailgun import client, config, signals
-from nailgun.entity_fields import IntegerField, OneToManyField, OneToOneField
+from nailgun.entity_fields import (
+    IntegerField, OneToManyField, OneToOneField, ListField)
 import threading
 import time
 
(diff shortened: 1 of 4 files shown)


renzon:
Cherry picking to be able to use on z streams branches. Worth mentioning that there is no 6.1.z branch on nailgun which would refer to 6.1.z satellite z stream.

nailgun/entity_mixins.py
```diff
@@ -191,6 +196,15 @@ def _payload(fields, values):
                 values[field_name + '_ids'] = [
                     entity.id for entity in values.pop(field_name)
                 ]
+            elif isinstance(field, ListField):
+                def parse(obj):
```
rochacbruno: Coveralls says this inner function is not covered, can you write a simple test case that covers this elif, or if does not make sense to test include a `... ListField): # pragma: no cover` comment here

nailgun/entity_mixins.py
```diff
@@ -1304,3 +1374,26 @@ def search_filter(entities, filters):
...
+
+    :param obj: entity or any json serializable object
+
+    :return: serializable object
+
+    """
+    if isinstance(obj, Entity):
+        return obj.to_json_dict()
+
+    if isinstance(obj, dict):
```
rochacbruno: All this cases below are not covered by tests, can you include simple test cases for each of those data types?

nailgun/entity_mixins.py
```diff
@@ -510,6 +523,63 @@ def __repr__(self):
...
+        ...         'id': 1,
+        ...         'name': 'Nailgun Org',
+        ...     }
+        >>> org = entities.Organization(config.ServerConfig('foo'), \*\*kwargs)
+        >>> org.to_json()
+        '{"id": 1, "name": "Nailgun Org"}'
+
+        :return: str
+        """
+        return std_json.dumps(self.to_json_dict())
```
rochacbruno: This return is also not being covered by test shoult it?

nailgun/entity_mixins.py
renzon: Strange, it seems cherry pick not worked as I expected. Going to check the test file because it was covered.

renzon:
@rochacbruno fixed test coverage.

rochacbruno:
Coveralls is still mad, but looks like a configuration problem, or some areas we miss a `pragma: no cover` comment

PR #358, opened 2017-01-19
Implemented Entity __eq__ method comparing all fields (#350)

renzon:
(cherry picked from commit ecf91be)

(cherry picked from commit f231e8c)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,6 @@ sudo: false
 language: python
 python:
     - 2.7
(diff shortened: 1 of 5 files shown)


renzon:
I am dropping python 3.3 as well because new sphinx is not compatible with it.

renzon:
Coverage not related to __eq__ and build failing for 3.3 due sphinx needing python >=3.4.

PR #359, opened 2017-01-20
Increase timeout for manifest delete/refresh

abalakh:
Default value for timeout is 5 minutes. Subscription upload takes more
time and as an exclusion has timeout of 15 minutes.
Automation results uncovered manifest delete/refresh commands take
pretty much comparable with upload amount of time (latest run shows
they took ~7 mins), which causes failures.
Updating timeout for manifest delete/refresh to match upload, hopefully
this will help us to get rid of constant subscription test failures.

Test results:
```python
py.test tests/foreman/api/test_subscription.py -k 'delete or refresh'
============================= test session starts ==============================
platform linux2 -- Python 2.7.10, pytest-3.0.3, py-1.4.31, pluggy-0.4.0
rootdir: /home/andrii/workspace/robottelo, inifile: 
plugins: xdist-1.13.1, repeat-0.2
collected 4 items 

tests/foreman/api/test_subscription.py ..

============================== 2 tests deselected ==============================
=================== 2 passed, 2 deselected in 78.82 seconds ====================
```

PR #360, opened 2017-01-20
[Cherry-pick][6.2.z] Increase timeout for manifest delete/refresh (#359)

abalakh:
Default value for timeout is 5 minutes. Subscription upload takes more
time and as an exclusion has timeout of 15 minutes.
Automation results uncovered manifest delete/refresh commands take
pretty much comparable with upload amount of time (latest run shows
they took ~7 mins), which causes failures.
Updating timeout for manifest delete/refresh to match upload, hopefully
this will help us to get rid of constant subscription test failures.

### DIFF
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -4568,7 +4568,12 @@ class Subscription(
             self._org_path('delete_manifest', kwargs['data']),
             **kwargs
         )
-        return _handle_response(response, self._server_config, synchronous)
+        return _handle_response(
+            response,
+            self._server_config,
+            synchronous,
+            timeout=900,
+        )
 
     def manifest_history(self, synchronous=True, **kwargs):
         """Obtain manifest history for subscriptions.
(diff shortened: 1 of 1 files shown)


rplevka:
are we really going to workaround issues like this?

abalakh:
@rplevka no better ideas, unfortunately. We filled the BZ already (actually, you did it 😄 ) pointing to bad performance, until the fix is there - which way are we gonna test manifests? Right now subscription upload takes around 7 minutes, so with 5 mins timeout we'll get a failure. For the functionality which is actually working, just requires some extra time. And the BZ is there. Any better ideas?

rplevka:
@abalakh sorry for non-constructive comment, I wanted to return to it once i finish me other task and add more info..

here's an idea on how we could proceed:
https://github.com/SatelliteQE/robottelo/issues/4028

I was just kind of busy with the rest of the failures so i couldn't dig into this more..

abalakh:
@rplevka good idea, whenever it's ready it will probably be a better workaround. Anyway imo it should not block this PR:
a) SatelliteQE/robottelo#4028 is currently an idea, without implementation, without tests (maybe 5 mins won't be enough even for a smaller repo?)
b) this PR doesn't introduce a new workaround, it applies already existing one to some missing places
c) it's basically a cherry-pick, to make 6.2.z consistent with master branch

rplevka:
ACK

PR #365, opened 2017-01-24
Add ProvisioningTemplate.

svtkachenko:
`Config Template` entity now is deprecated and it is just an alias for `Provisioning Template`, so rename should be enough.
Probably same change should be made for 6.2 as well.

### DIFF
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -1022,7 +1022,7 @@ class ConfigTemplate(
         """
         kwargs = kwargs.copy()  # shadow the passed-in kwargs
         kwargs.update(self._server_config.get_client_kwargs())
-        response = client.get(self.path('build_pxe_default'), **kwargs)
+        response = client.post(self.path('build_pxe_default'), **kwargs)
         return _handle_response(response, self._server_config, synchronous)
 
     def clone(self, synchronous=True, **kwargs):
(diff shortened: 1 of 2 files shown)


svtkachenko:
Looks like after all `Provisioning Template` and `Config Template` does not work at the same way, see [1]
So adding another entity instead of renaming existing one.

[1] https://bugzilla.redhat.com/show_bug.cgi?id=1416040

rochacbruno:
@svtkachenko you wrote:

> Probably same change should be made for 6.3 as well.

But you mean 6.2 ?

svtkachenko:
@rochacbruno Of course you're right, it was a typo.

rochacbruno:
@oshtaier can you review this one?

oshtaier:
Could you describe the exact difference between Config Template and Provisioning Template entities difference?

oshtaier:
Otherwise code looks good

svtkachenko:
@oshtaier In theory Provisioning Template is a new name for Config Template and ConfigTemplate should be deprecated, -- they have same parameters and methods, even help/doc for ConfigTemplate uses word `ProvisioningTemplate`. However, in practice there are problems with ProvisioningTemplate, at least one method `build_pxe_default` does not work, see [1] for details. At first I've just renamed ConfigTemplate class but then decided to add another class as broken functionality in ProvisioningTemplate brakes some tests.

[1] https://bugzilla.redhat.com/show_bug.cgi?id=1416040

oshtaier:
I don't really want to delay that PR anymore, but all that sounds as predictions. Please, create an issue for Config Template removal to not forget about all these facts

svtkachenko:
Deprecation of `ConfigTemplate` was confirmed by devs. so no predictions, but indeed issue is needed.

svtkachenko:
Looks like a problem with `build_pxe_default` it not a bug for Satellite but rather bug in nailgun.

svtkachenko:
Fixed bug for `build_pxe_default` method. Previously it used `GET` method within ConfigTemplate, however that was undocumented feature and didn't work within ProvisioningTempalte, `POST` should be used instead.

svtkachenko:
Issue created, bug fixed. PR finally can be merged :)

PR #369, opened 2017-01-30
Provide default name length & type for ComputeResource entities

abalakh:
Default name length for `AbstractComputeResource` (and subentities like LibvirtComputeResource) was not set, which can lead to entities with one character in name being created, which in rare cases (like latest automation run 😄 ) can attempt to create 2 entities with the same name (the longer name is, the less is the chance of the same name generation).
Also removing `cjk` from the tuple of acceptable string types as it creates issues with readability and basically increases efforts for debugging.
This PR should have been implemented as a part of #290

Issue #368, opened 2017-01-30
Add default name length and type for AbstractComputeResource


abalakh:
#290 haven't covered `AbstractComputeResource` entity (and all child subentities too), that's why we still can receive compute resources with 1 character in name or `cjk` characters.

abalakh:
Covered in #369, #370 and #371.

PR #371, opened 2017-01-30
[6.1.z] [Cherry-pick] Provide default name length & type for ComputeResource entities (#369)

abalakh:
Part of #368


rplevka:
ACK

PR #383, opened 2017-02-27
Update Image and Host entities

svtkachenko:
Cherry-picked from cherry-pick #356 :D
But seriously, I've found that Image implementation was wrong while cherry-picking it to 6.2.z. Time to have those changes in master.


renzon:
@pondrejk I am not the nitpick you mentioned is fixed. Can you check it and merge?

PR #388, opened 2017-03-13
Unique fields ignored on __eq__

### LINKED ISSUE #362: entities compare should ignore unique fields
When comparing 2 entities

```python
entity2 = entity1.clone()
entity2 == entity1
False
```

It will always return False, when entity has unique fields as `id` and others.

So we need to find some way in `__eq__` to remove that unique fields when comparing, or provide helper method for that.

renzon:
close #362

### DIFF
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -228,6 +228,7 @@ class ActivationKey(
                 required=True,
                 str_type='alpha',
                 length=(6, 12),
+                unique=True
             ),
             'organization': entity_fields.OneToOneField(
                 Organization,
@@ -357,6 +358,7 @@ class Architecture(
                 required=True,
                 str_type='alpha',
                 length=(6, 12),
+                unique=True
             ),
             'operatingsystem': entity_fields.OneToManyField(OperatingSystem),
         }
@@ -410,6 +412,7 @@ class AuthSourceLDAP(
                 required=True,
                 str_type='alpha',
                 length=(1, 60),
+                unique=True
             ),
             'onthefly_register': entity_fields.BooleanField(),
             'port': entity_fields.IntegerField(),
@@ -485,6 +488,7 @@ class Bookmark(
                 required=True,
                 str_type='alpha',
                 length=(6, 12),
+                unique=True
             ),
             'public': entity_fields.BooleanField(),
             'query': entity_fields.StringField(required=True),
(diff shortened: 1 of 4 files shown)


abalakh:
I don't get it, guys. Looking at example in issue
```python
entity2 = entity1.clone()
entity2 == entity1
False
```
this behavior seems 100% valid for me - cloned entity is a different entity, with different id, name and maybe something else. I was thinking `__eq__` for entities is used to understand whether some expected_entity and actual_entity are pointing to the same object in satellite. What's the use case for this?

renzon:
@abalakh 

AFAIK there are 2 use case for this feature:

1) You save a entity during some procedure and search for each for latter assertion. So in this case you are looking for identity. In this case you can do assertEqual(e1.id, e2.id) checking for id or even assertEqual(e1.to_json(), e2.id.to_json()) which will validate the entire properties.

2) You clone the entity and in that process unique, like id and name, fields will be different but all other will be eq other properties.

So that is what this implementation is addressing.

Right now __eq__ is working exactly the way you described. So for the same use cases:

1) You can only do an assertEqual(e1, e2)
2) But for cloned propeties you must exclude manually, all the time, unique fields. Ex: https://github.com/renzon/robottelo/blob/b44a53f7a2674ad97660fde3e9541b5e2adba11e/tests/foreman/api/test_contentview.py#L385-L385

So with that in mind, what you @abalakh think?

abalakh:
Frankly, on the one hand it sounds like totally different use cases for me - first one is used to ensure it the same entity, second one - that other entity have same field values, on the other hand - well, that's how `==` works in Python, unlike `is` it can point to different object with the same value. So i think this is fine, at first i thought more about `is` implementation which would be wrong.
It just looks pretty weird from user perspective, when entity A == entity B returns True, when those are definitely different entities and not clones :) Btw, one potential false positive generator is to use __eq__ to ensure entity B is a clone of entity A - we can accidentally think that totally different entity C (which just happens to have same entities assigned/other fields equal) is a clone of entity A.

renzon:
@abalakh: if you think more appropriate we could revert eq with the old behavior and create an compare_non_unique_fields or some method with better name.

rochacbruno:
@renzon if it is the case to revert then new method can be `entity.is_clone_of(other_entity)` we can discuss that deeper in the mtg.

PR #391, opened 2017-03-21
Reverted Entity __eq__ comparison on all fields, created compare to m…

### LINKED ISSUE #390: Make __eq__ compare all properties and create method "compare" to allow more granular comparisons
related to #388

renzon:
…ore granular comparisons

close #390

### DIFF
--- a/nailgun/entity_mixins.py
+++ b/nailgun/entity_mixins.py
@@ -588,15 +588,35 @@ class Entity(object):
         :param other: entity to compare self to
         :return: boolean indicating if entities are equal or not
         """
-        if other is None:
+        if not isinstance(other, type(self)):
             return False
+        return self.to_json_dict() == other.to_json_dict()
 
-        def filter_unique_fields(_, field):
-            """Filter functio for unique fields"""
-            return not field.unique
+    def compare(self, other, filter_fcn=None):
+        """Returns True if properties can be compared in terms of eq.
+        Entity's Fields can be filtered accordingly to 'filter_fcn'.
+        This callable receives field's name as first parameter and field itself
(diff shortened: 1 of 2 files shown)


renzon:
@abalakh @rochacbruno I think this make comparison more explicit. I called the method `compare` but let me know if you guys have a better name. As you may know, naming is the most difficult on Software Engineering ;)

nailgun/entity_mixins.py
```diff
@@ -590,13 +590,33 @@ def __eq__(self, other):
...
+        unique will not be compared by default. 'id' and 'name' are examples of
+        unique fields commonly ignored. Check Entities fields for fields marked
+        with 'unique=True'
+
+
+        :param other: entity to compare
+        :param filter_fcn: callable
+        :return: boolean
+        """
+        if other is None:
```
ldjebran: think we should add also the case where not isinstance(other, type(self))

```python 
if other is None or not isinstance(other, type(self)):
    return False
```
what do you think?

ldjebran: this comment apply to  __eq__ also, as we should have an exception that say that no attr  to_json_dic, instead directly return False

rochacbruno:
@renzon needs Cpick to 6.2.z ?

renzon:
@rochacbruno I am not sure, once it is not used by know. But if we start to use it, Cpick would be nice to made Cpicking on robottelo easier. What you think?

rochacbruno:
@renzon it is better to Cpick now, because in future it may end in errors if robottelo tests are implemented and then Cpicked.

PR #393, opened 2017-03-22
Update Host.create_missing with additional reads.

svtkachenko:
Inspired by @abalakh comment in https://github.com/SatelliteQE/robottelo/pull/4458#issuecomment-287330692

`Host`'s `create_missing` helper tried to access to some fields inside nested entities without checking whether these fields actually present. And they don't when passing `id` to `Host` instead of `Entity` object obtained by `read`, so additional read is necessary. 

Before:
```python
In [4]: entities.Host(config, environment=1).create(create_missing=True)
AttributeError: 'Environment' object has no attribute 'location'
In [5]: entities.Host(config, domain=1).create(create_missing=True)
AttributeError: 'Domain' object has no attribute 'location'
In [6]: entities.Host(config, operatingsystem=1).create(create_missing=True)
AttributeError: 'OperatingSystem' object has no attribute 'architecture'
```
After:
```python
In [5]: entities.Host(config, environment=1).create(create_missing=True)
Out[5]: nailgun.entities.Host( environment=nailgun.entities.Environment(nailgun.config.ServerConfig(..., id=1), ...)

In [6]: entities.Host(config, domain=1).create(create_missing=True)
Out[6]: nailgun.entities.Host(domain=nailgun.entities.Domain(nailgun.config.ServerConfig(..., id=1), ...)

In [7]: entities.Host(config, operatingsystem=1).create(create_missing=True)
Out[7]: nailgun.entities.Host(operatingsystem=nailgun.entities.OperatingSystem(nailgun.config.ServerConfig(..., id=1), ...)

In [8]: entities.Host(config, medium=1).create(create_missing=True)
Out[8]: nailgun.entities.Host(medium=nailgun.entities.Media(..., id=1), ...)
```

### DIFF
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -2856,6 +2856,9 @@ class Host(  # pylint:disable=too-many-instance-attributes
                  |-> domain
                  '-> environment
 
+        If nested entities were passed by `id` (i.e. entity was only
+        initialized and not read, and therefore contains only `id` field)
+        perform additional read request.
         """
         # pylint:disable=no-member,too-many-branches,too-many-statements
         super(Host, self).create_missing()
@@ -2874,6 +2877,8 @@ class Host(  # pylint:disable=too-many-instance-attributes
                 organization=[self.organization],
             ).create(True)
         else:
+            if not hasattr(self.domain, 'organization'):
+                self.domain = self.domain.read()
             if self.location.id not in [
                     loc.id for loc in self.domain.location]:
                 self.domain.location.append(self.location)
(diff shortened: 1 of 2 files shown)


oshtaier:
To be honest, I see nothing wrong to do additional reads unconditionally

nailgun/entities.py
```diff
@@ -2702,6 +2702,8 @@ def create_missing(self):
                  |-> domain
                  '-> environment
 
+        If nested entities contain only `id` (usually it means that there is no
+        `organization` field) perform additional read request.
```
abalakh: Didn't get your point in brackets. `(usually it means that entity was only initialized and not read)` maybe?

abalakh:
@oshtaier but there's no point in it in most cases (we do usually pass entities instead of ids as it's preferable way), those extra 4 GET requests will be applied to all existing host tests, and following workaround is actually needed for 3 tests in total introduced by @svtkachenko last PR. I'd vote to keep these reads in conditions.

PR #397, opened 2017-04-11
Generate unicast MAC for MACAddressField.

svtkachenko:
Fix for SatelliteQE/robottelo#4543.


sghai:
I think we need to fix the mac generation from Fauxfactory. But as per test results. looks like, the given fix works as expected.

sghai:
ACK

PR #411, opened 2017-05-28
Update SmartClassParameters and SmartVariables entities

svtkachenko:
Changes:
* `use_puppet_default` now is deprecated, `omit` should be used
* `hidden_value` now is used only to set boolean flag and `hidden_value?` is used to read it back, so `hidden_value` is missing in POST/GET response and should be ignored. Actual value now can be read from `default_value` instead of `hidden_value`: `u'*****'` when using GET without parameters and actual value when using GET with `show_hidden` parameter.

```
In [9]: entities.SmartVariable(server_config, puppetclass=10, variable='123', default_value='123', hidden_value=True).create()
Out[9]: nailgun.entities.SmartVariable(merge_overrides=False, override_value_order=u'fqdn\nhostgroup\nos\ndomain', puppetclass=nailgun.entities.PuppetClass(id=10), override_values=[], validator_type=None, variable=u'123', id=3072, description=None, default_value=u'*****', validator_rule=None, avoid_duplicates=False, merge_default=False, variable_type=None, hidden_value?=True)

In [11]: entities.SmartVariable(server_config, id=3072).read()
Out[11]: nailgun.entities.SmartVariable(merge_overrides=False, override_value_order=u'fqdn\nhostgroup\nos\ndomain', puppetclass=nailgun.entities.PuppetClass(id=10), override_values=[], validator_type=None, variable=u'123', id=3072, description=None, default_value=u'*****', validator_rule=None, avoid_duplicates=False, merge_default=False, variable_type=None, hidden_value?=True)

In [13]: sv.default_value
Out[13]: u'*****'

In [15]: getattr(sv, 'hidden_value?')
Out[15]: True
```
```

### DIFF
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -4020,6 +4020,7 @@ class OverrideValue(
                 SmartClassParameters),
             'smart_variable': entity_fields.OneToOneField(SmartVariable),
             'use_puppet_default': entity_fields.BooleanField(),
+            'omit': entity_fields.BooleanField(),
         }
         super(OverrideValue, self).__init__(server_config, **kwargs)
         # Create an override value for a specific smart class parameter
@@ -5229,6 +5230,7 @@ class SmartClassParameters(
             'merge_overrides': entity_fields.BooleanField(),
             'merge_default': entity_fields.BooleanField(),
             'avoid_duplicates': entity_fields.BooleanField(),
+            'override_value_order': entity_fields.StringField(),
             'override_values': entity_fields.DictField()
         }
         self._meta = {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: nailgun/entities.py [hunk 1]
```diff
@@ -4020,6 +4020,7 @@ def __init__(self, server_config=None, **kwargs):
                 SmartClassParameters),
             'smart_variable': entity_fields.OneToOneField(SmartVariable),
             'use_puppet_default': entity_fields.BooleanField(),
+            'omit': entity_fields.BooleanField(),
```
oshtaier: so `use_puppet_default` is not used anymore?

oshtaier:
Do you plan to fix codebase for robotello as well?

nailgun/entities.py
svtkachenko: It is marked as deprecated but still supported.

svtkachenko:
It is already in progress: https://github.com/SatelliteQE/robottelo/pull/4747
* `override_value_order` logic updated
* `hidden_value` is updated for CLI part, however API requires changes in nailgun's basic `read` method as currently it does not send parameters with GET request.
* `omit` should follow `use_puppet_default` logic but I want to perform additional testing of this parameters before updating all tests.

PR #417, opened 2017-06-28
Add docker support

JacobCallahan:
This change will allow us to make automated docker builds of nailgun.
Since nailgun is run from a python interpreter, the default behavior is
to run an interpreter upon initialization. From there, a user can
interact with nailgun as usual.
Alternatively, someone developing code for nailgun can make changes
locally, then build their own images using the Dockerfile.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,11 @@
+FROM python:3
+MAINTAINER https://github.com/SatelliteQE
+
+RUN mkdir nailgun
+COPY / /root/nailgun
+RUN cd /root/nailgun && python3 setup.py install
+
+ENV HOME /root/nailgun
+WORKDIR /root/nailgun
+
+CMD ["python"]


Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM python:3
```
JacobCallahan: With this tag, Docker will always pull the latest stable python version. This is currently 3.6.1. If desired, we can specify a version like 2.7 or 3.5.

JacobCallahan:
If this is something you guys really want, we can add this same Dockerfile to each branch and setup automated build for all of those as well.

JacobCallahan:
And of course a user can write a script in nailgun, mount it to a container and execute.

```
$ docker run -t -v $(pwd)/nailtest.py:/root/nailtest.py:Z nailgun python /root/nailtest.py
Created organization tester.
{'compute_resource': [],
 'config_template': [nailgun.entities.ConfigTemplate(id=1),
                     ...
                     nailgun.entities.ConfigTemplate(id=2)],
 'default_content_view': nailgun.entities.ContentView(id=10),
 'description': None,
 'domain': [],
 'environment': [],
 'hostgroup': [],
 'id': 10,
 'label': 'tester',
 'library': nailgun.entities.LifecycleEnvironment(id=10),
 'media': [],
 'name': 'tester',
 'provisioning_template': [nailgun.entities.ProvisioningTemplate(id=1),
                           ...
                           nailgun.entities.ProvisioningTemplate(id=2)],
 'smart_proxy': [nailgun.entities.SmartProxy(id=1)],
 'subnet': [],
 'title': 'tester',
 'user': []}
Deleted organization. Exiting...
```

PR #422, opened 2017-07-05
Check for r[field + '_id'] if not 'id' in r[field]

akofink:
This format of attrs currently breaks the `_get_entity_id` function:

Check out the example inputs to this function below (unnecessary data left out with '...').

```python
field_name = 'sync_plan'
attrs = {..., u'sync_plan_id': 10, ..., u'sync_plan': {u'name': u'Weekly RHEL Sync', u'interval': u'weekly', u'sync_date': u'2017-01-01 00:00:00 UTC', u'next_sync': None, u'description': None}, ...}
```

Note the absence of `attrs['sync_plan']['id']`

### DIFF
--- a/nailgun/entity_mixins.py
+++ b/nailgun/entity_mixins.py
@@ -245,9 +245,9 @@ def _get_entity_id(field_name, attrs):
     if field_name in attrs:
         if attrs[field_name] is None:
             return None
-        else:
+        elif 'id' in attrs[field_name]:
             return attrs[field_name]['id']
-    elif field_name_id in attrs:
+    if field_name_id in attrs:
         return attrs[field_name_id]
     else:
         raise MissingValueError(


abalakh:
@akofink shouldn't it be considered as API bug instead? It doesn't seem logical to me to place entity id outside entity dict.

akofink:
@abalakh We could "fix" this in the api, but since it's a one-to-one relationship, the related entity's ID is a column in the database on the main entity. I'd argue the API is more accurately representing the underlying data.

i.e.

```ruby
Katello::Product(id: integer, name: string, description: text, cp_id: string, multiplier: integer, provider_id: integer, created_at: datetime, updated_at: datetime, gpg_key_id: integer, sync_plan_id: integer, label: string, organization_id: integer)
```

Also, this behavior exists as early as Katello 3.0. On the other hand, since we're providing sync_plan details in the products#show response, I don't see why we would have left off the 'id' field. That *should* be there. I can submit an issue to Katello for that.

Even with the Katello API modification, we're still calling `['id']` without checking if it's actually there. That seems like an elementary oversight that should be fixed.

akofink:
https://github.com/Katello/katello/pull/6851

abalakh:
@akofink 
>  I don't see why we would have left off the 'id' field

That's basically our thoughts too.
I believe some time ago for such weird cases we've decided to introduce specific workarounds instead of changing global (_get_entity_id) logic, as it sounded more like a bug for us. If id-out-of-entity-dict is intended approach and will be used widely in katello, i think apart from updating _get_entity_id you should probably remove such workarounds, too. Not sure how many of them do we have, here's one example: https://github.com/SatelliteQE/nailgun/blob/master/nailgun/entities.py#L4231-L4239
But if it's not intended behavior i'd like to see it fixed in katello rather than workarounded in nailgun :)

akofink:
@abalakh Katello definitely has some issues with it's API. I just found that any OneToOne relationship with organization omits results['organization_id'] by default and only provides results['organization']['id']. #smh So, it seems you can't rely on anything; however, I can tell you how it *should* be:

A OneToOne field should absolutely always have the associated object's ID (i.e. sync_plan_id) at the top level, and there should be no guarantee the associated object's details are resolved and shown along with the top level object's response (i.e. results['sync_plan']['...']). If this were consistent throughout the app, `_get_entity_id` would always grab results['sync_plan_id'] and never need to grab results['sync_plan']['id'].

I don't recommend changing NailGun to how Katello *should* be and coding individual workarounds for inconsistencies, hoping they will eventually get fixed. Instead, we should make NailGun work with the current Katello. I think there are currently 4ish possible API responses for OneToOne relationships (let's take product->sync_plan for example):

```python
'sync_plan_id' in results and 'sync_plan' not in results # 1
'sync_plan_id' in results and 'id' not in results['sync_plan'] # 2
'sync_plan_id' in results and 'id' in results['sync_plan'] # 3
'sync_plan_id' not in results and 'id' in results['sync_plan'] # 4
'sync_plan_id' not in results and 'id' not in results['sync_plan'] # 5 (very unlikely)
```

We can safely assume wherever the ID is, it's correct, and if it's in 2 places, they match.

Currently, NailGun handles 1,3,4; this PR adds support for 2. Obviously, 5 would be a huge bug in Katello that would require immediate fix. :stuck_out_tongue_closed_eyes:

abalakh:
D'oh, disregard my comment about workarounds like https://github.com/SatelliteQE/nailgun/blob/master/nailgun/entities.py#L4231-L4239 , they're added for different purpose - to pass entities required for initialization.

So, according to your comment, katello doesn't follow one style, thus improving nailgun's response handling seems the best option.

akofink:
Who should we mention for merge?

abalakh:
@rochacbruno @ldjebran @svtkachenko @oshtaier bump

PR #436, opened 2017-08-21
add support for chunked content uploads

evgeni:
this allows uploading big files that cannot be uploaded in one go

### DIFF
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -23,6 +23,8 @@ workings of entity classes.
 """
 from datetime import datetime
 from sys import version_info
+import hashlib
+import os.path
 
 from fauxfactory import gen_alphanumeric
 from packaging.version import Version
@@ -41,8 +43,10 @@ from nailgun.entity_mixins import (
 
 if version_info.major == 2:  # pragma: no cover
     from httplib import ACCEPTED, NO_CONTENT  # pylint:disable=import-error
+    from urlparse import urljoin  # pylint:disable=import-error
 else:  # pragma: no cover
     from http.client import ACCEPTED, NO_CONTENT  # pylint:disable=import-error
+    from urllib.parse import urljoin  # pylint:disable=F0401,E0611
 
 # pylint:disable=too-many-lines
 # The size of this file is a direct reflection of the size of Satellite's API.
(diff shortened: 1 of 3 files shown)


ehelms:
ping @SatelliteQE/quality-engineers can someone review this please?

nailgun/entities.py
```diff
@@ -4827,6 +4836,144 @@ def upload_content(self, synchronous=True, **kwargs):
...
 
+    def content_upload_id(self, synchronous=True, **kwargs):
+        """Create an upload request
+
+        :param synchronous: What should happen if the server returns an HTTP
+            202 (accepted) status code? Wait for the task to complete if
+            ``True``. Immediately return the server's response otherwise.
+        :param kwargs: Arguments to pass to requests.
+        :returns: The server's response, with all JSON decoded.
+        :raises: ``requests.exceptions.HTTPError`` If the server responds with
```
omaciel: Don't you need to explicitly raise this exception by examining the response?

nailgun/entities.py
```diff
@@ -4827,6 +4836,144 @@ def upload_content(self, synchronous=True, **kwargs):
...
+        kwargs = kwargs.copy()  # shadow the passed-in kwargs
+        kwargs.update(self._server_config.get_client_kwargs())
+        response = client.post(self.path('content_uploads'), **kwargs)
+        json = _handle_response(response, self._server_config, synchronous)
+        return json
+
+    def content_upload_put(self, upload_id, filepath, **kwargs):
+        """Upload content chunked into a repository
+
+        Expects an upload_id and a path to a file
```
omaciel: Can you add the corresponding docstring tokens for the required arguments and their types?

nailgun/entities.py
```diff
@@ -4827,6 +4836,144 @@ def upload_content(self, synchronous=True, **kwargs):
...
+
+        :param kwargs: Arguments to pass to requests.
+        :returns: The server's response, with all JSON decoded.
```
omaciel: Where is the returned response object?

nailgun/entities.py
```diff
@@ -4827,6 +4836,144 @@ def upload_content(self, synchronous=True, **kwargs):
...
+                        'multipart': True}
+                client.put(content_upload_path, data, **kwargs)
+
+                offset += len(chunk)
+                chunk = contentfile.read(content_chunk_size)
+
+    def content_upload_delete(self, upload_id, **kwargs):
+        """Delete an (incomplete/imported) upload from a repository
+
+        It expects either a list of uploads or upload_ids (but not both).
```
omaciel: Please add docstring tokens for required parameters

nailgun/entities.py
```diff
@@ -4827,6 +4836,144 @@ def upload_content(self, synchronous=True, **kwargs):
...
+
+        :param kwargs: Arguments to pass to requests.
+        :returns: The server's response, with all JSON decoded.
```
omaciel: Explicit return needed?

omaciel:
Was there a reason why the existing mixings for `delete` and `put` were not used? Also, can we have tests?

nailgun/entities.py
rochacbruno: @omaciel you are right, it never raises for 4xx or 5xx, `requests` will raise only if address is not found
 `[Errno -5] No address associated with hostname`)

However as all our other methods are not raising I assume we do not want it to raise error but only logging the errors on logger output.

So we need either to remove all those `:raises:` from all methods here or make `nailgun.client` to raise the exceptions or capture explicitly

rplevka: As the guy who usually look into the test results trying to figure out, what went wrong, I vote for raising a proper exception.

evgeni:
You mean `EntityDeleteMixin` and `EntityUpdateMixin`? Those would operate on the `Repository`, not the upload inside the `Repository`. Also the `EntityUpdateMixin` does not support chunking as we need it (but that could be added, I guess).

Tests? Yes, absolutely, will do.

evgeni:
I realized there actually is a `ContentUpload` entity, which should be extended, instead. → WIP

nailgun/entities.py
evgeni: My expectation was that `_handle_response` does this for me as it calls `raise_for_status`?

Either way, I don't think this change would belong into this PR, right?

evgeni:
@SatelliteQE/quality-engineers this now has tests and uses the proper entity, please have another look, thanks!

nailgun/entities.py
rochacbruno: @evgeni yeah you are right! I missed the `raise_for_status` there https://github.com/SatelliteQE/nailgun/blob/0d186d822941be318bd637285fd0a75d901f097d/nailgun/entities.py#L113 So no change is needed and actually your PR ir right and don't need to raise explicitly here 
c/c @omaciel

ehelms:
It would also be quite helpful that then this gets merged, a new release is made.

rochacbruno:
@ehelms the maintainers are @omaciel @elyezer @sthirugn and @Ichimonji10 according to https://pypi.org/project/nailgun/ so only they have access to upload the release.
(which is something we can delegate to TraviCI to upload new releases when a new tagged commit passes)

elyezer:
I can make a release, just let me know when I should do it.

ehelms:
Now is good :)

rochacbruno:
One more review/merge needed before release

ehelms:
It's an ACK for me (forgot I was a maintainer on the git repo).

evgeni:
Thanks for merging!
Thanks for releasing!

:fireworks:

elyezer:
@evgeni thank **you** for the PR

PR #438, opened 2017-08-30
Added 'download_policy' field for Smart Proxy entity

abalakh:


oshtaier:
any examples?

abalakh:
Unfortunately, no test results as API doesn't return this field (BZ 1486609). However, it's updating correctly and functionality works as expected (see SatelliteQE/robottelo#5205 where the test relies on `dowload_policy` value).

PR #444, opened 2017-09-20
allow DELETE responses with a single blank

evgeni:
sometimes DELETE returns HTTP 200 with an empty response, but it's not
really empty, but contains only a blank. strip that blank and accept
this response as empty, not trying to parse JSON from it


ehelms:
@oshtaier @rochacbruno Could one of you look at this please?

evgeni:
@elyezer can we have a merge and a release with that bugfix please? :)

elyezer:
Sure thing, let me do that

PR #445, opened 2017-09-20
add relationship between Repository and Organization

evgeni:


akofink:
Looks good to me. Just needs a test for `read()` [here](https://github.com/SatelliteQE/nailgun/blob/master/tests/test_entities.py#L1653).

PR #447, opened 2017-09-27
Improve release.sh script

elyezer:
Allow the script to accept an optional argument which will be the
version that should be released when specified.


elyezer:
We usually want to increment the `y` number on the version `x.y.z` but some times we may want to increment the other versions so this improvement enables it.

elyezer:
Thanks

PR #454, opened 2017-11-14
filter urllib3.exceptions not requests.urllib3.exceptions

evgeni:
requests.urllib3 exists when you install requests via pip, but it should
not. and it does not exist when you install requests via rpm/deb.

thus let's filter the warnings directly from the urllib3 namespace, thus
making stuff work regardless how requests was installed

### DIFF
--- a/nailgun/client.py
+++ b/nailgun/client.py
@@ -19,6 +19,7 @@ from json import dumps
 from warnings import simplefilter
 import logging
 import requests
+import urllib3
 
 
 logger = logging.getLogger(__name__)  # pylint:disable=invalid-name
@@ -48,7 +49,7 @@ logger = logging.getLogger(__name__)  # pylint:disable=invalid-name
 # pylint:disable=no-member
 simplefilter(
     'ignore',
-    requests.urllib3.exceptions.InsecureRequestWarning,
+    urllib3.exceptions.InsecureRequestWarning,
 )
 
 


evgeni:
@Ichimonji10 as the original change was yours (https://github.com/SatelliteQE/nailgun/pull/66), can you please review?

Ichimonji10:
I tested this with a virtualenv:

```console
python3 -m venv env
source env/bin/activate
pip install requests
```

I then executed a pair of test scripts. One left warnings on:

```python
#!/usr/bin/env python
# coding=utf-8
"""Make an insecure request with relevant warnings on."""
import requests


def main():
    requests.get('https://example.com', verify=False)


if __name__ == '__main__':
    exit(main())
```

The other turned warnings off:

```python
#!/usr/bin/env python
# coding=utf-8
"""Make an insecure request with relevant warnings off."""
import urllib3
import warnings

import requests


def main():
    warnings.simplefilter('ignore', urllib3.exceptions.InsecureRequestWarning)
    requests.get('https://example.com', verify=False)


if __name__ == '__main__':
    exit(main())
```

Output:

```console
$ ./warnings_on.py
/home/ichimonji10/env/lib/python3.6/site-packages/urllib3/connectionpool.py:858: InsecureRequestWarning: Unverified HTTPS request is being made. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.io/en/latest/advanced-usage.html#ssl-warnings
  InsecureRequestWarning)
$ ./warnings_off.py
```

LGTM

PR #466, opened 2017-11-30
Search subs

sean797:
This fixes

```
[root@katello ~]# cat tt.py 
from nailgun import entities, entity_fields
from nailgun.config import ServerConfig

server = ServerConfig(
             url=server_url,
             auth=(username, password),
             verify=False
         )
org = entities.Organization(server).search()[0]
entities.Subscription(server, organization=org).search(query={'search': 'name="Red Hat OpenShift Container Platform, Standard (1-2 Sockets)"'})[0].read()
[root@katello ~]# python tt.py 
Traceback (most recent call last):
  File "tt.py", line 15, in <module>
    entities.Subscription(server, organization=org).search(query={'search': 'name="Red Hat OpenShift Container Platform, Standard (1-2 Sockets)"'})[0].read()
  File "/usr/lib/python2.7/site-packages/nailgun/entities.py", line 5809, in read
    return super(Subscription, self).read(entity, attrs, ignore, params)
  File "/usr/lib/python2.7/site-packages/nailgun/entity_mixins.py", line 812, in read
    in _get_entity_ids(field_name, attrs)
  File "/usr/lib/python2.7/site-packages/nailgun/entity_mixins.py", line 284, in _get_entity_ids
    return [entity['id'] for entity in attrs[plural_field_name]]
KeyError: 'id'
```

PR #472, opened 2017-12-13
search subs

sean797:
from https://github.com/SatelliteQE/nailgun/pull/466

ping @renzon

### DIFF
--- a/nailgun/entity_mixins.py
+++ b/nailgun/entity_mixins.py
@@ -281,7 +281,10 @@ def _get_entity_ids(field_name, attrs):
     elif field_name in attrs:
         return [entity['id'] for entity in attrs[field_name]]
     elif plural_field_name in attrs:
-        return [entity['id'] for entity in attrs[plural_field_name]]
+        try:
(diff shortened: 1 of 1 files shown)


renzon:
Can you add a test case for it and also a example of working console section? You've provided only a failure on #466 and would be nice to check it's working. 

Since _get_entity_ids is used by all entities I would say changing it can have a very wide impact. If test I could check if there is a way to change the read method, maybe overriding it, so this change would affect only Subscription itself.

renzon:
I have created an issue for this, check #475

renzon:
Hey @sean797, I came up with a less intrusive workaroung. Instead of changing a function which is used for all entities, I've just overridden read method on Subscription. Take a look on #476

rochacbruno:
@renzon @sean797 as #476 is merged can we close this one?

sean797:
Yes, sorry i've been inactive the last couple of weeks. Thnkas @renzon & @rochacbruno 👍

PR #473, opened 2017-12-13
Remove org from sync_plan api_path

sean797:
form https://github.com/SatelliteQE/nailgun/pull/461

ping @renzon

### DIFF
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -5282,7 +5282,6 @@ class SyncPlan(
     """
 
     def __init__(self, server_config=None, **kwargs):
-        _check_for_value('organization', kwargs)
         self._fields = {
             'description': entity_fields.StringField(),
             'enabled': entity_fields.BooleanField(required=True),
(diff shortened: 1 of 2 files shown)


renzon:
This will work for listing but will break for every other SyncPlan API call. Look at the docs on your satellite (path /api/v2).
```javascript
sync_plans: {
List sync plans: "/katello/api/sync_plans",
Show a sync plan: "/katello/api/organizations/:organization_id/sync_plans/:id",
Create a sync plan: "/katello/api/organizations/:organization_id/sync_plans",
Update a sync plan: "/katello/api/organizations/:organization_id/sync_plans/:id",
Destroy a sync plan: "/katello/api/organizations/:organization_id/sync_plans/:id",
Add products to sync plan: "/katello/api/organizations/:organization_id/sync_plans/:id/add_products",
Remove products from sync plan: "/katello/api/organizations/:organization_id/sync_plans/:id/remove_products",
Initiate a sync of the products attached to the sync plan: "/katello/api/sync_plans/:id/sync",
```
So we need to came up with a solution that work for all cases.

sean797:
@renzon This didn't break anything in my testing as `self.organization.path()` will work, so the URL doesn't change in that case.

TBH I think its a little weird that the URL is different for listing compared to other actions.

renzon:
@sean797 It is strange for me as well. Once only listing does not require a Organization I would expect it listing all sync plans from all organizations. But having only the list with this behavior is odd. Have you tried create a sync plan and listing with this modified path? Can you provide a console session for it?

sean797:
@renzon From my testing with this patch backported everything worked as expected.. I no longer have access to the Satellite I was using as I'm no longer with that customer. I could reproduce on my laptop but time doesn't permit that right now.

renzon:
@sean797 I have already acked after seeing the test you wrote. This PR is only missing another person review. @rochacbruno maybe you can do it.

PR #474, opened 2017-12-13
handle raise MissingValueError if we cannt find ID

sean797:
from https://github.com/SatelliteQE/nailgun/pull/462

ping @renzon


nailgun/entity_mixins.py
```diff
@@ -242,19 +242,23 @@ def _get_entity_id(field_name, attrs):
...
-    elif field_name_id in attrs:
-        return attrs[field_name_id]
-    else:
-        raise MissingValueError(
-            'Cannot find a value for the "{0}" field. Searched for keys named '
-            '{1}, but available keys are {2}.'
-            .format(field_name, (field_name, field_name_id), attrs.keys())
-        )
+            raise MissingValueError(error_msg)
+    except KeyError:
```
renzon: There is too many lines under try, you should so the there can be several sources of KeyError here. Ideally you should wrap only the line your are expecting getting this exception, which seems to be line 257 since all others are checking for key existence before using it.

omaciel:
@sean797 ping?

PR #478, opened 2018-01-05
[6.3 Feature] SSH Key helpers added in User entity

jyejare:
$subject


jyejare:
@abalakh , worked on your comment, please revisit !

PR #482, opened 2018-02-19
don't install tests into the binary distribution

evgeni:
this tries to install a "tests" python package, which we don't own.
at the same time, also exclude docs and contrib as done in the PyPA
sample project [1]

[1] https://github.com/pypa/sampleproject/blob/master/setup.py


evgeni:
@abalakh yes. I can remove it if you like to.

abalakh:
@evgeni if you don't mind - pls remove it then.

evgeni:
@abalakh updated

PR #483, opened 2018-03-04
create_missing only if explicitly True

JacobCallahan:
Current create_missing will run if any non-negative value is passed in.
This change will make it run only if the user passes in True.
This shouldn't affect 'normal' usage, but rizza exploits this to abuse create_missing.
Example:
Executing: ActivationKey create with fields: {} and args {'create_missing': 'e9e5490b-6596-4f1f-817e-11c445c76d63'}


JacobCallahan:
@ehelms @renzon would this impact the ansible module?

ehelms:
Going to defer to @akofink to assess

PR #491, opened 2018-04-13
Fix fauxfactory compatibility

rochacbruno:
Fix travis Ci error.

Fauxfactory now dropped support for Python 2.x

We need to ensure the corrent pinned version in nailgun before release.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ python:
     - 2.7
     - 3.4
 install:
-    - pip install -r requirements.txt -r requirements-dev.txt
+    - pip install -r requirements.txt -r requirements-dev.txt --no-cache-dir
 script:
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -21,9 +21,12 @@ docs-clean:
 
 lint:
 	flake8 --ignore=E731 .
-	pylint -j $(CPU_COUNT) --reports=n --disable=I --ignore-imports=y \
+	pylint -j $(CPU_COUNT) --reports=n -E \
```
rochacbruno: Used `-E` to raise only Errors and fatal errors, while the above `flake` will still look for styling but is more permissive, this is recommended while transitioning to Py3k

rochacbruno:
This fixes #486

PR #495, opened 2018-05-16
RHEV APIv4 feature: Add checkbox to select between API v4 and v3.

lhellebr:


lhellebr:
Tested OVirtComputeResource(...).create() with use_v4= True, False and without use_v4. Works.

PR #504, opened 2018-06-27
allow searching Content View filters

sean797:


sean797:
Thanks @abalakh I've tested this when I raised the PR and confirmed it works - I wouldn't have raised it otherwise. I have no easy way to currently provide the output of `.search()`.

sghai:
@sean797 : hey.. thank you for your contribution.

We trust your testing but its a process for everyone where every contributor has to share test results. One of the main reason is that we are using same code base extensively in product testing and want to avoid issues in automation due to code changes.

Moreover, we are planning to branch for 6.4 version so would like to merge all open PRs sooner than later. Your test results would really help us to merge the PR quickly.

sean797:
@sghai I'm just struggling to understand why by copying and pasting output from `.search()` is going to actually provide a real value.

- I could fairly easily fake it, maybe copy from another PR and make applicable changes.
- The above is actually probably easier for me since I no longer have access to a Satellite/Katello and would have to build one.
- I wouldn't be testing against 6.4 so my results would be somewhat not relevant.
- This change just allows me to run a couple of `search` methods, I don't believe it can really break anything.

I don't mean to be destructive, I just think process doesn't actually provide any real value (at least not in this PR) If you are really worried about breakages we should consider setting up some integration tests to provide that reassurance.

---

If you really want to insist on the requested output I doubt I can get around to building a Satellite before 6.4 branching

sean797:
It's a shame you have processes like this, but enjoy.. 

```
 >>> AbstractContentViewFilter(server_config, name='my filter').search()
[nailgun.entities.AbstractContentViewFilter(repository=[], inclusion=True, content_view=nailgun.entities.ContentView(id=12), description=None, type=u'rpm', id=3, name=u'my filter')]

sghai:
@sean797 : We trust our contributors. Intent is to run the tests once with selected instance, not just copy paste. I'll talk to folks if we can come up with some other way out to ensure the code is working. 

Thank you for test results.

PR #514, opened 2018-07-17
reverted locked state to ptable

sghai:
This change was affecting 6.4 automation. Soon we are planning to do 6.4 branching and then we can accommodate this change.


rplevka:
ACK

sghai:
@sbernhard : here is the PR https://github.com/SatelliteQE/nailgun/pull/542 raised by @jyejare

PR #515, opened 2018-07-26
Ping Entity is not searchable

sean797:


nailgun/entities.py
```diff
@@ -4612,7 +4612,7 @@ def __init__(self, server_config=None, **kwargs):
         super(Permission, self).__init__(server_config, **kwargs)
 
 
-class Ping(Entity, EntitySearchMixin):
+class Ping(Entity):
```
abalakh: This change will break at least one existing test: https://github.com/SatelliteQE/robottelo/blob/master/tests/foreman/endtoend/test_api_endtoend.py#L946
And leave it with no possible workarounds.

You're basically right, this entity is not searchable and the test is rather trying to read it, not search. But removing ability to send GET request completely isn't the best option either since it will introduce regression which can't be fixed.
How about adding `EntityReadMixin` and proper fields to entity so we start using it properly?

sean797:
Updated, thanks @abalakh

san7ket:
Created issue https://github.com/SatelliteQE/robottelo/issues/6165 to fix the broken test , to use `EntityReadMixin` instead

abalakh:
Seems like this change won't be so trivial.
`read()` was not designed to work with base paths, but with specific entity instances, e.g. `Product().read()` will try to send GET request to `/katello/api/products/:id`, not to base `/katello/api/products`.
On the contrary `search()` is designed to send GET to base path, that's probably why it was chosen for Ping() entity.

Switching to EntityReadMixin will introduce at least 2 issues:
1. When sending GET request, entity instance path will be chosen instead of base path. This issue can be workarounded with custom `read_raw()` method like this:
```python
    def read_raw(self, params=None):
        return client.get(
            self.path('base'),
            params=params,
            **self._server_config.get_client_kwargs()
        )
```
2. When processing response, nailgun will expect `id` of entity to be present in outputs and will rise exception as it won't be found. `id` is set in entity's `__init__` method, which makes it a bit tricky to override. My best guess right now is to override `get_fields()` with smth like that:
```python
    def get_fields(self):
        fields = self._fields.copy()
        fields.pop('id')
        return fields
```

But introducing 2 such workarounds makes it debatable which approach is better - to use `search()` instead of `read()` or to use `read()` but with 2 workarounds. Thoughts?

abalakh:
@sean797 btw that's why we always ask for test results - despite your change right now is trivial one-linear, unfortunately it won't work due to reasons described above.

sean797:
@abalakh I'm thinking we should leave it as is - thanks.

> @sean797 btw that's why we always ask for test results - despite your change right now is trivial one-linear, unfortunately it won't work due to reasons described above.

As a maintainer, it's your job, as well as the contributor, to test changes. Relying on the contributor to post the output of his or her tests is mostly useless. You have no reason to trust the contributors' tests, how do you know there isn't something environmental to the contributor that is making it work? This is especially relevant with a library like Nailgun that could be used across many different Foreman, Katello or Satellite versions. I suggested a far better approach in https://github.com/SatelliteQE/nailgun/pull/504#issuecomment-405891723

Though if you are going to require something like posting test results when people make changes, ensuring that the requirement is documented will mean I as a contributor will know what I'm letting myself in for before I choose to contribute or fork some software.

PR #528, opened 2018-08-20
Make repository_sets use /repository_sets/

sean797:
This also removes the requirements for the `product` field.

### DIFF
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -5450,7 +5450,6 @@ class RepositorySet(
         EntitySearchMixin):
     """ A representation of a Repository Set entity"""
     def __init__(self, server_config=None, **kwargs):
-        _check_for_value('product', kwargs)
         self._fields = {
             'contentUrl': entity_fields.URLField(required=True),
             'gpgUrl': entity_fields.URLField(required=True),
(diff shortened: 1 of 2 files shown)


sean797:
ping?

san7ket:
@sean797 I tested this today with robottelo tests for enable and disable reposets, but I faced issues on 6.4, as it doesn't seem to find the product as we are directly traversing to reposets endpoint. Will try to test with with latest nighties and report back , if this breaks anything

sean797:
@san7ket thanks!. I'd suggest you use Katello nightly as https://github.com/Katello/katello/commit/33c8c329e268029674e5bb83c615a7e972ab737d is required really, (although that commit is mostly documentation)

sean797:
@san7ket reading your error again, it look like your need to pass in a product ID, for example:

```
RepositorySet().search(set(), {'search': 'product_id=528'})

sghai:
@san7ket : ping ? 

@jyejare could you please review too ?

jyejare:
@sghai , Currently the nightlies and upstream devel both are broken, So we are unable to test these PR's. Once its back, we will be able to test new and modified functionality as well.

PR #536, opened 2018-09-13
Added compliance policies entity

schlupov:


schlupov:
This isn't the final code. I just needed this because of one cli test.

nailgun/entities.py
```diff
@@ -4966,6 +4969,28 @@ def __init__(self, server_config=None, **kwargs):
         super(PuppetModule, self).__init__(server_config, **kwargs)
 
 
+class Policies(Entity, EntityReadMixin,):
```
sean797: Mind changing the class to `CompliancePolicies`? There could be many types of policies with more plugins.

nailgun/entities.py
```diff
@@ -4966,6 +4969,28 @@ def __init__(self, server_config=None, **kwargs):
...
+        self._fields = {
+            'location': entity_fields.OneToManyField(Location),
+            'name': entity_fields.StringField(
+                required=True,
+                str_type='alpha',
+                length=(4, 30),
+                unique=True
+            ),
+            'organization': entity_fields.OneToManyField(Organization),
+            'hosts': entity_fields.ListField()
```
jyejare:
```suggestion
            'hosts': entity_fields.OneToManyField(Host)
```

nailgun/entities.py
```diff
@@ -4966,6 +4969,28 @@ def __init__(self, server_config=None, **kwargs):
...
+        }
```
jyejare: The compliance Policies has other fields as well, If you add them now it would be helpful. Just a suggestion.

schlupov:
@jyejare @sean797 Changed what was requested.

nailgun/entities.py
schlupov: I'll keep it in my mind, I just work on something else now, I'm sorry.

PR #543, opened 2018-10-19
lower the chunk size for content uploads to 2MB

evgeni:
Pulp/Django limit the upload size to about 2.5MB, so let's play safe and
use 2MB here.


evgeni:
this will need to be cherry picked to `6.4.z` too, as this issue started with the Pulp/Django version there.

PR #548, opened 2018-10-25
Ignore arguable "W504 line break after binary operator"

lpramuk:
It started to fail on 4 years old commits.

Some say W503 and W504 are contradictory. We can expect some change regarding these rules. So I better ignore than go fixing it.


ldjebran:
Think we can fix that but not in the immediate, just say for the moment on the recommendation level for new PRs.

PR #554, opened 2018-11-02
Use content credentials katello API instead of old gpg_keys

sbernhard:
This change will allow us to support SSL certificate management with
foreman_ansible_modules

### DIFF
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -1618,6 +1618,41 @@ class DockerRegistryContainer(AbstractDockerContainer):
         super(DockerRegistryContainer, self).__init__(server_config, **kwargs)
 
 
+class ContentCredential(
+        Entity,
+        EntityCreateMixin,
+        EntityDeleteMixin,
+        EntityReadMixin,
+        EntitySearchMixin,
+        EntityUpdateMixin):
+    """A representation of a Content Credential entity."""
+
+    def __init__(self, server_config=None, **kwargs):
+        self._fields = {
+            'content': entity_fields.StringField(required=True),
+            'name': entity_fields.StringField(
+                required=True,
+                str_type='alpha',
+                length=(6, 12),
+                unique=True
+            ),
+            'organization': entity_fields.OneToOneField(
+                Organization,
+                required=True,
+            ),
+            'content_type': entity_fields.StringField(
+                choices=('cert', 'gpg_key'),
+                default='gpg_key',
+                required=True,
+            ),
+        }
+        self._meta = {
+            'api_path': 'katello/api/v2/content_credentials',
+            'server_modes': ('sat'),
+        }
+        super(ContentCredential, self).__init__(server_config, **kwargs)
+
+
 class ContentUpload(
         Entity,
         EntityCreateMixin,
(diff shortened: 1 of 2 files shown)


sbernhard:
Thanks for the review @sean797. See updated commit.

sean797:
Nice :+1: Fancy also adding a message to tell people `GPGKey` class is deprecated? I suspect a simple `print` would work just fine.

nailgun/entities.py
```diff
@@ -2672,6 +2706,7 @@ def __init__(self, server_config=None, **kwargs):
             'api_path': 'katello/api/v2/gpg_keys',
             'server_modes': ('sat'),
         }
+        print('GPGKey entity is deprecated. Use ContentCredential entity instead!')
```
ldjebran:
```python
warnings.warn('GPGKey entity is deprecated. Use ContentCredential entity instead!')
```

nailgun/entities.py
```diff
@@ -1618,6 +1618,40 @@ def __init__(self, server_config=None, **kwargs):
...
+                str_type='alpha',
+                length=(6, 12),
+                unique=True
+            ),
+            'organization': entity_fields.OneToOneField(
+                Organization,
+                required=True,
+            ),
+            'content_type': entity_fields.StringField(
+                choices=('cert', 'gpg_key'),
```
ldjebran: perhaps we should add a default value
```python
default = 'gpg_key'
```

sean797: I would prefer not to, Katello doesn't have a default (https://github.com/Katello/katello/blob/master/app/controllers/katello/api/v2/content_credentials_controller.rb#L15) I don't think Nailgun should be any different.

ldjebran: My suggestion has nothing to do with katello, if nailgun was a new implementation I will 100% agree with you guys, but this implementation will break the tests of entities (product, repository ) related to GPGKey.
but if for example we put that default value we can even inherit PGPKey from ContentCredential
```python
class ContentCredential( 
     ...
     default = 'gpg_key'
    ...

class GPGKey(ContentCredential):
    pass
```
with this nothing will be broken.

nailgun/entities.py
```diff
@@ -4688,7 +4724,7 @@ class Product(
     def __init__(self, server_config=None, **kwargs):
         self._fields = {
             'description': entity_fields.StringField(),
-            'gpg_key': entity_fields.OneToOneField(GPGKey),
+            'gpg_key': entity_fields.OneToOneField(ContentCredential),
```
ldjebran: This will break the current tests

nailgun/entities.py
```diff
@@ -5188,7 +5224,7 @@ def __init__(self, server_config=None, **kwargs):
                 default='immediate',
             ),
             'full_path': entity_fields.StringField(),
-            'gpg_key': entity_fields.OneToOneField(GPGKey),
+            'gpg_key': entity_fields.OneToOneField(ContentCredential),
```
ldjebran: This will break the current tests

nailgun/entities.py
ldjebran: Other suggestion is to not update Product and Repository entity, and let the warning in PGPKey. That warning is catched by the log system and open an issue in robottelo to make changes to the tests related to PGPKey entity (to be replaced by ContentCredential) and when done we can even delete PGPKey entity.

sbernhard:
thanks @ldjebran. I guess you are right regarding the default parameter in this context and the unnecessary duplication of code.

sean797:
Thanks @sbernhard :+1:

Issue #553, opened 2018-10-30
Create and Document the Nailgun Processes


san7ket:
For now, we are creating a rough draft of the processes that would make easier for contributions and review.
We are still at the initial phase, and will share across the document, once we get it outline and structure ready.
I would like everyone to participate to develop this, so that it would benefit the project and all other projects that consume nailgun.

oshtaier:
First of all, I think we need to enforce common sense things:
1) Keep the same style, standards and approach that code already has. That basically means, if people followed something for years that something has reasons behind.
2) First comment for every PR should contain test results that prove code is working. That means we, of course, trust every contributor commit, but let's be respectful to each other
3) At least two ACKs per each PR and more in case something changed to fundamental, basic architecture
4) Continue add unit tests at least in a way we did till that moment
5) That is fine to ping people in polite manner and if you are ready for constructive feedback

Will add more if remember something

sean797:
> Keep the same style, standards and approach that code already has. That basically means, if people followed something for years that something has reasons behind. Want to change it - provide reasons, don't wait that people behind existing approach provide reasons to you

It easy to make a change an not see "style, standards or approach" where you do. Any style or standards we want to enfore we should ideally do via linitng tests.

> First comment for every PR should contain test results that prove code is working.

I strongly disagree with this, data like this provide little to no value. There is a whole load of reasons why something that works for the contributor, does not work somewhere else. Consistent contributors should test a contribution before raising a PR.


> but let's be respectful to each other

I'm glad you said that.

oshtaier:
>It easy to make a change an not see "style, standards or approach" where you do. Any style or standards we want to enfore we should ideally do via linitng tests.
Okay, till we get to more lint tests and till contributor open his or her eyes to see obviously same/similar code in other methods we have to have proper review procedure. That still get us to the point that we need people who familiar with a framework to merge properly formed code from some dev branch to the master. And everyone who doesn't have time can use that dev branch for himself and his projects. We need to isolate proper commits and commits that done according some selfish rules.

>I strongly disagree with this, data like this provide little to no value. There is a whole load of reasons why something that works for the contributor, does not work somewhere else. 
Right, but that is only your point. We can provide reasons why that procedure worked for us for years to guarantee some safety even for yourself. Especially, when person fix own code after some review cycles.

>Consistent contributors should test a contribution before raising a PR.
So what's the problem to provide test results if you test something for real

>I'm glad you said that.
I'm glad **you** said that.

[...]

PR #561, opened 2018-11-08
Remove pyxdg dependency

sean797:
I have no idea why this was added in the first place?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -22,7 +22,6 @@ with open('VERSION') as handle:
 REQUIREMENTS = [
     'inflection',
     'packaging',
-    'pyxdg',
     'requests>=2.7',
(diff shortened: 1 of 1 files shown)


ntkathole:
@sean797 travis has the answer for your question. This was implemented to make nailgun smart enough to find default configuration file.
Ref : https://github.com/SatelliteQE/nailgun/issues/2#issuecomment-67708906

sean797:
ah thanks @ntkathole :+1:  If i'd `grep -r` for `xdg` instead of `dxg` made I would have noticed that :-1: 

Closing for now, I plan to revist this laster depending on the output of https://github.com/SatelliteQE/nailgun/issues/553

PR #568, opened 2018-11-15
Add FAM tests

sean797:
This commit makes use of pre-recorded tests that FAM uses, this has a couple of benefits:
- Increase test coverage, using real-world recorded API interactions between nailgun and server. Unit tests only take us so far.
- pre-recorded means they are fast, no need to build and server, ect..
- Using the pre-recorded tests in FAM means Nailgun doesn't need to record its own.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,6 @@ __pycache__/
 # Coverage information
 /.coverage
 /htmlcov/
+
+# FAM tests
+/foreman-ansible-modules/
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,10 @@ python:
     - 3.6
 install:
     - pip install -r requirements.txt -r requirements-dev.txt --no-cache-dir
+addons:
+  apt:
+    packages:
+    - rpm
 script:
     - make lint
     - make test-coverage
@@ -12,6 +16,7 @@ script:
     - make docs-clean
     - make package
     - make package-clean
+    - make test-fam
 after_success:
     coveralls
 notifications:
--- a/Makefile
+++ b/Makefile
@@ -48,4 +48,10 @@ package-clean:
 publish: package
 	twine upload dist/*
 
+test-fam:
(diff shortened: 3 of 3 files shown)


sean797:
I should also note, the goal here is to add integration tests, if anyone has a suggestion for a different type of integration test then please to suggest.

ldjebran:
I am a bit confused here, do not think that this is the right place for this tests, think that we have not to add tests for dependent projects here and that tests should goes in their relevant projects, but perhaps other people may have other opinion.

abalakh:
Same here. I don't want to put NACK and am open for discussion, but it feels like wrong direction to me - if FAM tests in nailgun is fine, then maybe we should put robottelo API tests here as well? AFAIK the intend is to build nailgun independent from any third-party, whether it's robottelo or FAM or something else, and i remember @sean797 you agreed with this point and wanted naigun to be less dependent from robottelo/satellite as well :)

And one more thing. Let's assume we merge this PR. Then i open a new PR with changes to nailgun, and FAM tests are starting to fail. On the one hand - it's cool we'll see some customers are affected by this change, but on the other - my PR will be blocked by Travis, and there's not much i can do to unblock the change - i'm not aware of how FAM works internally and i doubt i'll be able to rise a PR to fix the issue in FAM easily. So in the end nailgun development will be dependent on FAM. Currently master branch is dependent on robottelo/satellite and that's what we're trying to fix and will probably introduce different processes/branches (#553 and the doc sent by email which is currently under review) but that won't have much value if we introduce different dependency instead.

sean797:
Thanks for the responses guys :+1: as @abalakh has explained, I agree FAM tests are not appropriate here, for the sake of clarity this wasn't about stopping breaking changes that FAMs uses or anything like that, it was about setting up some more tests and using FAM was a quick and easy way to do it.

Closing this now - Hopefully, we can look at either VCR, https://github.com/ansible/foreman-test-container or another testing method in the future to complement the current unit tests.

Issue #203, opened 2015-09-07
File a Bz for SmartProxy.update()


sghai:
Looks like `PUT` request doesn't return all its attributes.
```
MissingValueError: Cannot find a value for the "organization" field. Searched for keys named ('organization_ids', 'organization', 'organizations'), but available keys are [u'name', u'url', u'created_at', u'updated_at', u'content_host_id', u'id'].
```

Ichimonji10:
:+1:

PR #570, opened 2018-11-23
Add MTU support to subnets

bagasse:
As MTU support for subnet has landed in [foreman 1.18](https://theforeman.org/manuals/1.18/index.html#Releasenotesfor1.18) and this version is included in [Satellite 6.4 (login required)](https://access.redhat.com/articles/1343683), this add MTU attribute in subnets for Satellite >= 6.4 and is needed by [theforeman/foreman-ansible-modules
#203](https://github.com/theforeman/foreman-ansible-modules/pull/203)

### DIFF
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -6397,6 +6397,7 @@ class Subnet(
             'network': entity_fields.IPAddressField(required=True),
             'to': entity_fields.IPAddressField(),
             'vlanid': entity_fields.StringField(),
+            'mtu': entity_fields.IntegerField(min_val=68, max_val=4294967295),
         }
         if _get_version(server_config) >= Version('6.1'):
             self._fields.update({


nailgun/entities.py
```diff
@@ -6361,7 +6361,11 @@ def __init__(self, server_config=None, **kwargs):
             'network': entity_fields.IPAddressField(required=True),
             'to': entity_fields.IPAddressField(),
             'vlanid': entity_fields.StringField(),
+            'mtu': entity_fields.IntegerField(),
         }
+
+        if _get_version(server_config) < Version('6.4'):
```
sean797: I wouldn't bother with this if I was you, the Ansible modules don't set the `version` attr in `server_config`. Plus we have different branches for different supported versions see https://github.com/SatelliteQE/nailgun/branches. Kind of like we 2 strategies for supporting different server versions.

Someone else may be able to give an authoritive answer on what server versions the master branch should support

nailgun/entities.py
```diff
@@ -6361,7 +6361,11 @@ def __init__(self, server_config=None, **kwargs):
             'network': entity_fields.IPAddressField(required=True),
             'to': entity_fields.IPAddressField(),
             'vlanid': entity_fields.StringField(),
+            'mtu': entity_fields.IntegerField(),
```
sean797: Is it worth setting `min_val=68, max_val=4294967295` ?

nailgun/entities.py
san7ket: Yes, we have a 6.4 branch, if needed we can backport this to 6.4 later. Currently, Master points to 6.5 or foreman 1.20. 
we may soon have documentation published around this, to have a clear idea

bagasse:
Thanks @san7ket, @sean797 !

PR #574, opened 2018-11-30
Send product_id to Enable/disable/available_repositories repository_s…

sean797:
…ets APIs

These endpoints require either `product_id` or `organization_id`.

### DIFF
--- a/tests/test_entities.py
+++ b/tests/test_entities.py
@@ -1883,7 +1883,6 @@ class GenericTestCase(TestCase):
         """
         cfg = config.ServerConfig('http://example.com')
         generic = {'server_config': cfg, 'id': 1}
-        repo_set = {'server_config': cfg, 'id': 1, 'product': 2}
         sync_plan = {'server_config': cfg, 'id': 1, 'organization': 2}
         cls.methods_requests = (
             (entities.AbstractDockerContainer(**generic).logs, 'get'),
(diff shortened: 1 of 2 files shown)


tests/test_entities.py
```diff
@@ -1990,7 +1990,7 @@ def test_generic(self):
                         response = method(**kwargs)
                 self.assertEqual(client_request.call_count, 1)
                 self.assertEqual(len(client_request.call_args[0]), 1)
-                self.assertEqual(client_request.call_args[1], kwargs)
+                assert set(kwargs).issubset(set(client_request.call_args[1]))
```
abalakh: I'm not sure it's a good idea to remove that strict equivalence check for all the entities as it can hide issues from us. Mind writing separate test for these 3 methods instead?

sean797:
thanks @abalakh updated :+1:

abalakh:
@sean797 thanks, by why did you remove synchronous from args, shouldn't we wait till the result? Or those 3 commands are always synchronous?

sean797:
Turns out just `available_repositories` is always synchronous, added `synchronous` option back in for enable/disable. 

https://github.com/Katello/katello/blob/master/app/controllers/katello/api/v2/repository_sets_controller.rb#L82

oshtaier:
`These endpoints require either product_id or organization_id.` ? So if we have passed organization_id in kwargs - we will not need product_id?

oshtaier:
I have no problem with that, but theoretically, we will not be able to test scenario when I want to pass only `organization_id` and see whether functionality will work with only one required parameter?

sean797:
Okay, lets only do this when data isn't supplied?

PR #576, opened 2018-11-30
Read Pool ID of Subscription

jyejare:
Reads the pool_id (cp_id) of subscription.


nailgun/entities.py
```diff
@@ -6483,6 +6483,7 @@ class Subscription(
     def __init__(self, server_config=None, **kwargs):
         self._fields = {
             'activation_key': entity_fields.OneToManyField(ActivationKey),
+            'cp_id': entity_fields.StringField(),
```
sean797: Mind adding `unique=True` ?

PR #577, opened 2018-12-03
added foreman_tasks_recurring_logic_id in sync plan entity

omkarkhatavkar:
**Description:** 
This is a new field was added in  sync plan for Satellite 6.5 feature while migrating sync plan from Pulp to Katello 

**How to use?**
```
In [10]: org = entities.Organization(id=1).read()

In [11]: sync_plan = entities.SyncPlan(organization=org, name='ddsd').search()

In [12]: sync_plan[0].foreman_tasks_recurring_logic.id
Out[12]: 2
```
**Objective** 

This will help to resolve https://github.com/SatelliteQE/robottelo/issues/6511 issue

### DIFF
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -6687,6 +6687,7 @@ class SyncPlan(
             ),
             'product': entity_fields.OneToManyField(Product),
             'sync_date': entity_fields.DateTimeField(required=True),
+            'foreman_tasks_recurring_logic': entity_fields.OneToOneField(RecurringLogic)
         }
         super(SyncPlan, self).__init__(server_config, **kwargs)
         self._meta = {


nailgun/entities.py
```diff
@@ -6687,6 +6687,7 @@ def __init__(self, server_config=None, **kwargs):
             ),
             'product': entity_fields.OneToManyField(Product),
             'sync_date': entity_fields.DateTimeField(required=True),
+            'foreman_tasks_recurring_logic_id': entity_fields.IntegerField(),
```
san7ket: Is the recurring_logic a entity, If yes, please raise a issue to add the entity and replace the IntegerField by it, after the issue is resolved.

omkarkhatavkar: @san7ket Recurring logic entity already exists in a nailgun code base here -> https://github.com/SatelliteQE/nailgun/blob/23d93a316197dd3089ec5e69a6ce08eea9b1dddb/nailgun/entities.py#L5254. If you want I can create the issue for this change to add entity_field in sync plan.

san7ket: `'foreman_tasks_recurring_logic_id': entity_fields.RecurringLogic(),`

omkarkhatavkar: make sense now, earlier I did not understand your question. It is better to make it as you mentioned as above.

san7ket: Ignore the comment above, 

'foreman_tasks_recurring_logic_id':entity_fields.OneToOneField(RecurringLogic),

omkarkhatavkar: Ok, Good to ACK then

sean797: Why not change this now? Instead of raising another PR later (which I think would be a breaking change)

omkarkhatavkar: @sean797 I'm holding this PR till the time I see the right way to implement this. I'm also thinking how to achieve this as foreman_tasks_recurring_logic_id is just a read-only field

san7ket:
@omkarkhatavkar can we get the O/P once please

ldjebran:
@san7ket think he updated the PR doc with output

PR #604, opened 2019-01-10
Make ComputeProfile searchable

sean797:
Hoping to be able to remove https://github.com/theforeman/foreman-ansible-modules/blob/master/module_utils/ansible_nailgun_cement.py#L200

Its been in FAM for 1 year now, https://github.com/theforeman/foreman-ansible-modules/blob/master/module_utils/ansible_nailgun_cement.py#L468

### DIFF
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -765,6 +765,7 @@ class ComputeProfile(
         EntityCreateMixin,
         EntityDeleteMixin,
         EntityReadMixin,
+        EntitySearchMixin,
         EntityUpdateMixin):
     """A representation of a Compute Profile entity."""
 


mirzal:
@sean797 this looks straightforward enough, but I'm missing some context. Do you know why we haven't had search mixin to begin with? Has something in the product changed since this entity was added?

Could you also add "test" result for that? Doesn't have to be a test, just run it from Python interpreter.

sean797:
Thanks for reviewing so quick!

Sorry - no I don’t know why it was missed before. Nothing has changed that
I’m aware of.

FAM has tests for compute_profile (
https://github.com/theforeman/foreman-ansible-modules/blob/master/test/test_playbooks/compute_profile.yml)
the module actually relies on the search function as it needs to search for
the compute profile in order to work out if it needs to
create/update/destroy entity. Give the above, do you still want test output?

oshtaier:
@mirzal we do not add all working mixins by default, but only ones that we need at exact moment. Of course, it is very nice if person do that, but it was never mandatory

Issue #606, opened 2019-01-22
Subnet class uses field with invalid (reserved) name - "from"


rplevka:
We are currently not able to instantiate a Subnet class with "from" parameter - this name is reserved in python and throws a syntax error.
https://github.com/SatelliteQE/nailgun/blob/master/nailgun/entities.py#L6779

PR #612, opened 2019-01-29
Medium org loc

rplevka:
fixes 3 issues:
1. https://github.com/SatelliteQE/nailgun/issues/606
2. Org + Loc entities use attribute called 'media', this does not work with generating the parameter correctly as the correct one uses singular form: `medium`
3. Subnet entity does not support assigning of the `template` proxy


nailgun/entities.py
```diff
@@ -6814,8 +6814,12 @@ def __init__(self, server_config=None, **kwargs):
                 'remote_execution_proxy':
                     entity_fields.OneToManyField(SmartProxy),
                 'subnet_parameters_attributes': entity_fields.ListField(),
+                'template': entity_fields.OneToOneField(SmartProxy),
                 'tftp': entity_fields.OneToOneField(SmartProxy),
             })
+        if kwargs.get('_from'):
```
abalakh: we use opposite form for reserved words - `'from_'`. E.g. `DiscoveryRule.search_`

abalakh: Please test this fix with all CRUD operations, previously we had to implement custom read/update/create with such fix (like in DiscoveryRule), so i'm not so sure it will work properly for all cases.

nailgun/entities.py
```diff
@@ -6814,8 +6814,12 @@ def __init__(self, server_config=None, **kwargs):
                 'remote_execution_proxy':
                     entity_fields.OneToManyField(SmartProxy),
                 'subnet_parameters_attributes': entity_fields.ListField(),
+                'template': entity_fields.OneToOneField(SmartProxy),
```
abalakh: btw, mind removing this check (`if _get_version(server_config) >= Version('6.1')`) completely? It's a leftover from times when we had 1 branch for all satellite versions, now it doesn't make sense and 2 separate dicts aren't that readable.

nailgun/entities.py
rplevka: sure. looks like we might still have trouble using this with `.update`

rplevka: done, pls, resolve this if happy with the solution

nailgun/entities.py
rplevka: Done, however Travis checks don't like that, looking into it

rplevka:
attached some logs which demonstrate the functionality

nailgun/entities.py
rplevka: looks like the [test_subnet_fields](https://github.com/SatelliteQE/nailgun/blob/master/tests/test_entities.py#L3626) is obsoleted, so I'm removing it

nailgun/entities.py
```diff
@@ -6789,33 +6789,31 @@ def __init__(self, server_config=None, **kwargs):
             'to': entity_fields.IPAddressField(),
             'vlanid': entity_fields.StringField(),
             'mtu': entity_fields.IntegerField(min_val=68, max_val=4294967295),
+            'boot_mode': entity_fields.StringField(
```
abalakh: Bonus karma if you sort all fields alphabetically 🙂

nailgun/entities.py
```diff
@@ -6825,8 +6823,13 @@ def create_payload(self):
...
         <https://bugzilla.redhat.com/show_bug.cgi?id=1151220>`_.
 
+        In addition, rename the ``from_`` field to ``from``.
+
         """
-        return {u'subnet': super(Subnet, self).create_payload()}
+        payload = super(Subnet, self).create_payload()
+        if 'from_' in payload:
+            payload['from'] = payload.pop('from_')
+        return {u'subnet': payload}
```
abalakh: Great, thanks. Now to the coverage - you should create test in `CreatePayloadTestCase` (similar to `test_discovery_rule`), `ReadTestCase` (same as existing `test_discovery_rule` there) and `UpdatePayloadTestCase` (same as `test_discovery_rule_search`).

nailgun/entities.py
```diff
@@ -6849,7 +6857,10 @@ def read(self, entity=None, attrs=None, ignore=None, params=None):
 
     def update_payload(self, fields=None):
         """Wrap submitted data within an extra dict."""
-        return {u'subnet': super(Subnet, self).update_payload(fields)}
+        payload = super(Subnet, self).update_payload(fields)
+        if 'from_' in payload:
+            payload['from'] = payload.pop('from_')
```
JacobCallahan: would this also work here?
```python
payload['from'] = payload.pop('from_', payload['from'])
```

JacobCallahan: Ahh looks like you're replacing from with from_ internally, but updating the payload back to from. What was the reasoning behind this?
Also, that is just to make it a one-liner

rplevka: see the bullet point 1 in the description: https://github.com/SatelliteQE/nailgun/issues/606

nailgun/entities.py
rplevka: @abalakh, done, could you review them? thanks

nailgun/entities.py
rplevka: done, pls. resolve the conversation if happy with the changes

tests/test_entities.py
```diff
@@ -1530,6 +1540,19 @@ def test_usergroup_with_external_usergroup(self):
         self.assertTrue(hasattr(usergrp, 'usergroup'))
         self.assertIn('usergroup', read.call_args[0][2])
 
+    def test_subnet(self):
+        """Call :meth:`nailgun.entities.Subnet.read`.
+
+        Ensure that the ``from_`` attribute is successfully set.
```
abalakh: This statement isn't covered, that was probably a bad advice to copy `test_discovery_rule` which doesn't verify it as well, sorry for that.
Just add one extra line in the end of test:
```python
self.assertIn('from_', read.call_args[0][1])
```

rplevka: oops, i actually had that in there, but screwed up the commiting part :)

nailgun/entities.py
JacobCallahan: ahhh gotcha. that is annoying, but unavoidable without a change to the API

tests/test_entities.py
rplevka: done

PR #616, opened 2019-02-22
Adding generic webhook configuration for readthedocs for auto docs bu…

san7ket:
…ilding

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,8 @@ script:
     - make package
     - make package-clean
 after_success:
-    coveralls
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -13,7 +13,8 @@ script:
     - make package
     - make package-clean
 after_success:
-    coveralls
+    - "curl -X POST -d \"token=${readthedocs_token}\" \"https://readthedocs.org/api/v2/webhook/nailgun/76623/\""
```
mirzal: What is this number (76623)? Where does it come from? readthedocs documentation says it's `<id>`, but nothing more. I'm wondering if this is dynamic or static.

san7ket: It's static project no. also I might change the implementation to use authenticated webhook

PR #622, opened 2019-03-08
Fix string date in sync plan

mdellweg:


nailgun/entities.py
```diff
@@ -7225,7 +7213,7 @@ def create_payload(self):
 
         """
         data = super(SyncPlan, self).create_payload()
-        if 'sync_date' in data:
+        if isinstance(data.get('sync_date'), datetime):
```
mdellweg: maybe this conversion should be done by `to_json_serializable`

ldjebran: Think we should pass a datetime always, what is the issue related to this change (as strftime has to fail if we supply other than datetime, otherwise we should do a validation: say if not datetime should raise some  Exception)?

mdellweg: This can be an already formatted string.
Also why should this field be handled any different than in the update case?

ldjebran: Perhaps there is an issue with the update that should accept only datetime too :)
as read always return datetime , create is with datetime now,  update str or datetime, 
but my original question is what is the issue associated to this change, why you need this, perhaps you want to create sync_date from yaml/json files or from text content, personally if this is the case I have no objection, just we need to justify this change, and put here some  example usage with some output.

mdellweg: The problem arises in https://github.com/theforeman/foreman-ansible-modules/pull/249 .

And that does not look like datetime to me:
```
In [6]: sp.search()[0].sync_date
Out[6]: u'2019/03/13 23:20:00 +0000'

In [7]: sp.search()[0].read().sync_date
Out[7]: u'2019/03/13 23:20:00 +0000'
```

mdellweg: @ldjebran 
BTW: I am ok with passing strings around here.

PR #639, opened 2019-06-03
[Cherry-pick] Removes Singnals.py as we no longer use it

san7ket:


san7ket:
ok to test : tests/foreman/api/test_activationkey.py -m tier1

PR #645, opened 2019-06-20
authSourceLDAP Search support (#644)

ntkathole:
(cherry picked from commit ffce8d8d039384bbf5af491b25c10171862498c7)


jyejare:
@ntkathole Probably for update_Search_mixin u need to write/cherrypick the unit tests here.

jyejare:
I suspect travis will fail with that.

jyejare:
> I suspect travis will fail with that.

hh ... The coverage remained same !

PR #666, opened 2019-10-15
Add ignorable_content field to Repository entity

kersommoura:
Add ignorable_content field to Repository entity.

### DIFF
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -6068,6 +6068,7 @@ class Repository(
             ),
             'full_path': entity_fields.StringField(),
             'gpg_key': entity_fields.OneToOneField(ContentCredential),
+            'ignorable_content': entity_fields.ListField(),
             'label': entity_fields.StringField(),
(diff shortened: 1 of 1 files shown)


nailgun/entities.py
```diff
@@ -6068,6 +6068,7 @@ def __init__(self, server_config=None, **kwargs):
             ),
             'full_path': entity_fields.StringField(),
             'gpg_key': entity_fields.OneToOneField(ContentCredential),
+            'ignorable_content': entity_fields.ListField(),
```
rochacbruno: Is there a fixed choice list to require here?

kersommoura: I will take a look. Thanks.

kersommoura: > List of content units to ignore while syncing a yum repository. Must be subset of rpm,drpm,srpm,distribution,erratum

rochacbruno: I was thinking to add a validator to ensure the values on that list are on that subset, but looks like the `ListtField` does not support `choices` or any other kind of validator. So nm.

PR #673, opened 2019-11-13
Remove container and registry entities, docker compute resource

mirekdlugosz:
These were removed from Satellite during 6.6 cycle, but we agreed to remove them from frameworks in release after 6.6.


mirekdlugosz:
@SatelliteQE/quality-engineers with https://github.com/SatelliteQE/robottelo/pull/7478 merged, nothing in Robottelo uses these entities anymore. 
Can I get second ACK and merge?

PR #676, opened 2019-11-14
Cherrypick reporttemplates commits

lhellebr:


jyejare:
@lhellebr , You can look at the coverage report : https://coveralls.io/builds/26992277 and cover that by adding tests.

lhellebr:
@jyejare , this is only a cherrypick, these commits have been in master for a long time already

lhellebr:
@jyejare , can you merge this please? Even when I add the tests, they will need to be in a new commit, in a new PR in master, then I can make a new cherrypick in here.

PR #692, opened 2020-01-06
GCE CR Support

jyejare:
$subject

### DIFF
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -1295,6 +1295,22 @@ class VMWareComputeResource(AbstractComputeResource):
             entity, attrs, ignore, params)
 
 
+class GCEComputeResource(AbstractComputeResource):
+    """A representation of a Google Compute Resource entity."""
+
+    def __init__(self, server_config=None, **kwargs):
+        self._fields = {
+            'email': entity_fields.StringField(required=True),
+            'key_path': entity_fields.StringField(required=True),
+            'project': entity_fields.StringField(required=True),
+            'zone': entity_fields.StringField()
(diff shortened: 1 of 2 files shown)


jyejare:
Was needed for https://github.com/SatelliteQE/robottelo/pull/7568 which is merged. Please help to review also the results are already submitted with the PR link in this comment.

nailgun/entities.py
```diff
@@ -1295,6 +1295,22 @@ def read(self, entity=None, attrs=None, ignore=None, params=None):
...
 
 
+class GCEComputeResource(AbstractComputeResource):
+    """A representation of a Google Compute Resource entity."""
+
+    def __init__(self, server_config=None, **kwargs):
+        self._fields = {
+            'email': entity_fields.StringField(required=True),
+            'key_path': entity_fields.StringField(required=True),
+            'project': entity_fields.StringField(required=True),
```
ntkathole: apidoc shows email, key_path, project as optional... required=true needed for them?

jyejare: Yes, All of them are absolutely mandatory.

jyejare: Mandatory for GCE, but not for others. So probably that's the reason APIDoc says it's optional.

mirekdlugosz:
I see OVirt and VMWare shadow `read` method to ignore password field - should GCE do something similar?

jyejare:
@mirekdlugosz , There is no password in GCE, but there is key_path. And I dont find any harm exposing keypath stored under admins/root account. 

What do u think ?

PR #697, opened 2020-01-29
Added support for AzureRM CR

vijay8451:


nailgun/entities.py
```diff
@@ -3661,7 +3691,7 @@ def __init__(self, server_config=None, **kwargs):
             'host_parameters_attributes': entity_fields.ListField(),
             'image': entity_fields.OneToOneField(Image),
             'interface': entity_fields.OneToManyField(Interface),
-            'interfaces_attributes': entity_fields.ListField(),
+            'interfaces_attributes': entity_fields.DictField(),
```
mirekdlugosz: Why is this change done? Isn't it backwards-incompatible?

Although I can't find any reference to 'interfaces_attributes' in robottelo, so probably it doesn't matter?

vijay8451: `interfaces_attributes` should be `dict` field only ... however, as it was not in use so no one noticed the same hence updated it.

PR #700, opened 2020-02-18
Add pre-commit configuration, remove pylint

mshriver:
Remove uses of pylint throughout code, remove its methods from make configuration

Add pre-commit config to use python-reorder-imports, flake, and some other basic code/whitespace checks
Update flake config with pylint default ignores

Update travis config to use pre-commit
Include some docs notes on it

Add .nailgun to gitignore, logical venv name


docs/create_organization_nailgun.py
```diff
@@ -5,9 +5,10 @@
 Use NailGun to accomplish this task.
 
 """
+from pprint import pprint
```
jyejare: Alphabetically this should be below nailgun. And looks like pre-commit doesn't cares much about alphabetically.

Are we going to change the rule of alphabetical order hence forth ?

docs/create_user_nailgun.py
```diff
@@ -2,10 +2,13 @@
...
 """Create an identical user account on a pair of satellites."""
 from __future__ import print_function
-from nailgun.config import ServerConfig
-from nailgun.entities import Organization, User
+
 from pprint import pprint
 
+from nailgun.config import ServerConfig
+from nailgun.entities import Organization
+from nailgun.entities import User
```
jyejare: Two different classes imports from same module? Can we instruct pre-commit to make this one or is it intentional?

jyejare:
@mshriver This is a great addition to the nailgun. Thanks for looking into it.

docs/create_user_nailgun.py
mshriver: This is the ordering done by reorder-python-imports, not by pre-commit itself.

https://github.com/asottile/reorder_python_imports#what-does-it-do

docs/create_organization_nailgun.py
mshriver: Imports are not sorted only alphabetically, but grouped into three distinct groups in order of: built-in imports, installed imports, and local imports.

In this case pprint is in the first group for built-in imports, there are no installed imports, and the nailgun imports are local to the project.

docs/create_user_nailgun.py
jyejare: Yes, Coz I don't see any advantage other than a bunch of import lines added. These days our IDE itself shows if there are any import conflicts.

But that's just my opinion and probably you would have already discussed with @JacobCallahan and @rochacbruno .

docs/create_organization_nailgun.py
jyejare: Why I said this cause, we have been following the practice of sorting imports alphabetically. So now, We can just send an Email to the satqe list saying:
"Hey ! You need not to waste more time in sorting the imports, keep them as it is and pre-commit will take care"

docs/create_user_nailgun.py
mshriver: Personally, I'm not a particular fan of splitting the imports like this, but I've learned to enjoy having the tool manage them for me, and know that they're always sorted and its always easy to locate an import when visually parsing them.

docs/create_organization_nailgun.py
mshriver: Exactly - its still sorting them alphabetically, just into groups as well.

Pre-commit should just modify the files automatically when a contributor makes a local commit, and they shouldn't have to worry about it (and reviewers don't have to worry or check the imports either!)

docs/create_user_nailgun.py
mirekdlugosz: @jyejare Rationale is given a little further down on page linked by Mike, under [Why this style](https://github.com/asottile/reorder_python_imports#why-this-style) heading. 

We don't really have that kind of problem in our repos, so it's more like a style choice. Some people will like it, some people won't. I didn't like it at first, but we all can agree that our time is better spent elsewhere. Having tool managing it for us means we don't have to waste a single second thinking about this.

docs/create_organization_nailgun.py
mirekdlugosz: To clarify, [Robottelo "code standards"](https://robottelo.readthedocs.io/en/latest/code_standards.html) mentions alphabetical sorting, but it also mentions grouping imports into three groups (standard library, third-party and application-specific). I believe most people understood it as alphabetical sorting within a group, not across the groups (because it doesn't make sense any other way).

`reorder-python-imports` uses the same rule. So it cleans up mess we have made, enforces following of rule for new code and (optionally) automatically applies it while developing.

mshriver:
@mirekdlugosz if this is going to be ported to various active branches (assuming it will be) then both commits, and all of the changes in them, should be included in the port. I don't see an advantage to splitting the pylint removals into another commit, when the inclusion of either commit here would negate their usefulness.

mirekdlugosz:
@mshriver as mentioned on list, we'll probably want to cherry-pick 0ea61f86e5ae20a58e82509c45b00db1bca8d84e and then run `pre-commit`, to avoid dealing with merge conflicts in automatically changed code.

My thought was in the same line - assuming pylint comments removal are responsible for merge conflicts, it might make sense to not cherry-pick them, but run the same automation on new branch. I have now tried cherry-picking 0ea61f86e5ae20a58e82509c45b00db1bca8d84e and it looks like there are no conflicts caused by pylint comments, so we don't have to worry about this.

mirekdlugosz:
@jyejare are your comments resolved, or is there anything else you would like to discuss?
@JacobCallahan your thoughts on that?

Mike already submitted cherry-pick PR, and I wonder if there is anything preventing us from merging this one.

docs/create_organization_nailgun.py
jyejare: @mirekdlugosz Why I was more focusing on this because people were very much used to the way we were following earlier. And as I said earlier now they need not to worry about it.

jyejare:
@mirekdlugosz Thanks for the reminder. I am all set and happy to merge this one. Kudos to @mshriver and team.

JacobCallahan:
@mirekdlugosz @mshriver I'd still like to see flake8 removed from the current repo (deprecated) and over to its new location.

PR #701, opened 2020-02-19
Add Discovery Reboot and Refresh Facts support

jyejare:
$subject along with unit tests.


mirekdlugosz:
@jyejare can you check Travis failure and fix it?

Looks like there is no code in Robottelo that depends on that change, so I would say it's not a branching blocker - we can branch 6.7.z without this PR and cherry-pick it later, if required.
What do you think?

jyejare:
@mirekdlugosz @hkx303 @JacobCallahan  Sorry for the late. Travis is now fixed.

Please ACK and merge.

PR #702, opened 2020-02-19
Drop py2 support, specify py 3.6

mshriver:
Update setup.py/travis supported python versions.

Make some updates in regard to py3.6 support including f-strings, and no `u''` string specs.

NOTE: At the moment this includes the commits from #700, they will be re-based out.

### DIFF
--- a/nailgun/client.py
+++ b/nailgun/client.py
@@ -114,10 +114,7 @@ def _log_response(response):
     :return: Nothing is returned.
 
     """
-    message = u'Received HTTP {0} response: {1}'.format(
-        response.status_code,
-        response.text
-    )
+    message = f'Received HTTP {response.status_code} response: {response.text}'
     if response.status_code >= 400:  # pragma: no cover
         logger.warning(message)
     else:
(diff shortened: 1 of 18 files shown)


mirekdlugosz:
@evgeni @sean797 do you still use Nailgun in any way in FAM? If we drop Python2 support here, will that break anything for you?

evgeni:
Nope. You can drop py2 as far as FAM is concerned, as we don't use nailgun in master or any recent releases anymore.

nailgun/client.py
```diff
@@ -114,10 +114,7 @@ def _log_response(response):
     :return: Nothing is returned.
 
     """
-    message = u'Received HTTP {0} response: {1}'.format(
-        response.status_code,
-        response.text
-    )
+    message = f'Received HTTP {response.status_code} response: {response.text}'
```
jyejare: :+1:

jyejare:
@mshriver , ACKed , We have not been using py2 in master at all. Coz the Robottelo and Satellite-upgrade itself has moved to Py3 recently. But this is a good cleanup. :)

PR #710, opened 2020-03-19
Backport #702

mshriver:


JacobCallahan:
After a basic run of API tests (maybe activation keys), we can merge.

mirekdlugosz:
Looks like we are in stalemate here, and I would **really** like this to be merged before we branch of 6.7.z.

@mshriver Would it be possible for you to provide tests results that Jake asked for in next day or two?
@JacobCallahan Are these tests results hard requirement for you? We have merged #702 without any tests, and requiring tests now is a little puzzling for me.

JacobCallahan:
@mirekdlugosz they're not a hard requirement, no. but being a QE, i prefer to test things before merge :)
if @mshriver isn't able to get some tests ran today, then we can assume some risk with a merge.

mirekdlugosz:
Based on last comment, I'm taking the risk and merging it now.

Issue #714, opened 2020-05-06
UI testing for Sat6.8 failed due to organization fields changed


hkx303:
UI testing failed due to the error below:

```
@fixture(scope='module')
    def module_org():
        """Shares the same organization for all tests in specific test module.
        Returns 'Default Organization' by default, override this fixture on
    
        :rtype: :class:`nailgun.entities.Organization`
        """
        default_org_id = (
            nailgun.entities.Organization()
            .search(query={'search': 'name="{}"'.format(DEFAULT_ORG)})[0]
            .id
        )
>       return nailgun.entities.Organization(id=default_org_id).read()

robottelo/tests/foreman/virtwho/conftest.py:28: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/usr/local/lib/python3.7/site-packages/nailgun/entities.py:5250: in read
    return super(Organization, self).read(entity, attrs, ignore, params)
/usr/local/lib/python3.7/site-packages/nailgun/entity_mixins.py:800: in read
    in _get_entity_ids(field_name, attrs)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

field_name = 'config_template'
attrs = {'ancestry': None, 'compute_resources': [], 'created_at': '2020-04-30 10:59:51 UTC', 'default_content_view_id': 1, ...}

    def _get_entity_ids(field_name, attrs):
        """Find the IDs for a one to many relationship.
    
        The server may return JSON data in the following forms for a
        :class:`nailgun.entity_fields.OneToManyField`::
    
            'user': [{'id': 1, …}, {'id': 42, …}]
            'users': [{'id': 1, …}, {'id': 42, …}]
            'user_ids': [1, 42]
    
        Search ``attrs`` for a one to many ``field_name`` and return its ID.
    
        :param field_name: A string. The name of a field.
        :param attrs: A dict. A JSON payload as returned from a server.
        :returns: An iterable of entity IDs.
    
        """
        field_name_ids = field_name + '_ids'
        plural_field_name = pluralize(field_name)
        if field_name_ids in attrs:
            return attrs[field_name_ids]
        elif field_name in attrs:
            return [entity['id'] for entity in attrs[field_name]]
[...]

PR #724, opened 2020-05-20
 deleting all occurencies of config template

tstrych:
config template was causing error in `test_positive_get_per_page`
I would like to finish here what was started in #714 

And I would like open discussion here, what we are going to do with all other occurencies which are still in `entities.py`: 
```
output of command grep -nr 'config_template'

entities.py:1561:            'api_path': 'api/v2/config_templates',
entities.py:1591:        return {'config_template': payload}
entities.py:1599:        return {'config_template': payload}
entities.py:1607:            /config_templates/build_pxe_default
entities.py:1609:            /config_templates/clone
entities.py:1611:            /config_templates/revision
``` 

I had local problems with `test_positive_get_per_page`:
```
 # Compute Resource related entities
        cls.compresource_libvirt = entities.LibvirtComputeResource(
            organization=[cls.org], location=[cls.loc]
        ).create()
>       cls.image = entities.Image(compute_resource=cls.compresource_libvirt).create()
```
Test result after change:
```
=================== 1 passed, 77 deselected in 42.77 seconds ===================
```

### DIFF
--- a/tests/test_entities.py
+++ b/tests/test_entities.py
@@ -107,7 +107,6 @@ class InitTestCase(TestCase):
                 entities.ComputeAttribute,
                 entities.ComputeProfile,
                 entities.ConfigGroup,
-                entities.ConfigTemplate,
                 entities.CompliancePolicies,
                 entities.ProvisioningTemplate,
                 entities.ReportTemplate,
@@ -274,7 +273,6 @@ class PathTestCase(TestCase):
         for entity, path in (
                 (entities.ActivationKey, '/activation_keys'),
                 (entities.Capsule, '/capsules'),
-                (entities.ConfigTemplate, '/config_templates'),
                 (entities.ProvisioningTemplate, '/provisioning_templates'),
                 (entities.ReportTemplate, '/report_templates'),
                 (entities.Role, '/roles'),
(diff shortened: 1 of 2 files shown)


hkx303:
Since the `config_template` was renamed to `provisoning_template`, I think it is better to remove the `config_template` entity

tstrych:
I removed all occurences of `ConfigTemplate` and `config_template` from nailgun.

tests/test_entities.py
```diff
@@ -107,7 +107,7 @@ def test_init_succeeds(self):
                 entities.ComputeAttribute,
                 entities.ComputeProfile,
                 entities.ConfigGroup,
-                entities.ConfigTemplate,
```
tstrych, line 110: I let this one, as here are other deprecated entities.

rplevka: hehe - I think that "see below" tries to point at:
https://github.com/SatelliteQE/nailgun/blob/master/tests/test_entities.py#L225-L228

- but some other contributors probably just replicated that comment without appending the commented entity in the deprecated list.
- Anyway, even if deprecated, the entity is kept in `entities` - taking `OperatingSystemParameter` for example.

Since you're removing the entity entirely - I would either simply delete this line here or kept the entity and appended it in the deprecated list.

I'm not 100% sure about what I'm saying here, I just assume from looking at this piece of code.
@mirekdlugosz - any comments on how to approach this one?

mirekdlugosz: @rplevka looks like deprecated entities still exist, they only raise exception when used on Satellite newer than when they existed. So if we want to call them deprecated here, we should leave them in nailgun, but with changes.

Since we are removing them from nailgun, I think we should remove them here as well.

And since we are the only consumers of nailgun, I don't see a point of going out of our way with deprecation warning etc.

tstrych:
Right, I removed all instances except that commented one. 
I checked now, and you can see all `provisioning_template` and `ProvisioningTemplate` have already been in place. So in my opinion it doesn't require any change there.

tstrych:
Squashing commits done.

PR #757, opened 2020-10-07
Override update_payload in some entities to only include fields that actually can be updated in the payload. Add update capability to Model. This is required by a robottelo change that tests updating of many entity types, using update_payload (rather than create_payload as it used to be).

lhellebr:
Used by https://github.com/SatelliteQE/robottelo/pull/8046 that fixes automation failure in `tests/foreman/api/test_multiple_paths.py::EntityIdTestCase::test_positive_put_status_code`

This has potential to break something; tested by running all the tests in `tests/foreman/api/test_multiple_paths.py` with a result of
```
============================= 3 failed, 13 passed, 7 warnings in 864.63 seconds =============================
```
The 3 failures don't seem to be related.

### DIFF
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -2007,6 +2007,7 @@ class ContentCredential(
     """A representation of a Content Credential entity."""
 
     def __init__(self, server_config=None, **kwargs):
+        self.updatable_fields = ['name', 'content_type', 'content']
         self._fields = {
             'content': entity_fields.StringField(required=True),
             'name': entity_fields.StringField(
@@ -2030,17 +2031,6 @@ class ContentCredential(
             'server_modes': ('sat'),
         }
         super(ContentCredential, self).__init__(server_config, **kwargs)
-
-    def update_payload(self, fields=None):
-        """Override the method to only update fields that actually can be updated
-        """
-        updatable_fields = set.intersection({'name', 'content_type', 'content'},
-                                            self.get_values().keys())
-        if fields is None:
-            fields = updatable_fields
-        else:
-            fields = set.intersection(set(fields), updatable_fields)
-        return super().update_payload(fields=fields)
 
 
 class ContentUpload(
@@ -3318,6 +3308,7 @@ class HostCollection(
     """A representation of a Host Collection entity."""
 
     def __init__(self, server_config=None, **kwargs):
+        self.updatable_fields = ['name', 'description', 'host_ids', 'max_hosts', 'unlimited_hosts']
         self._fields = {
             'description': entity_fields.StringField(),
             'host': entity_fields.OneToManyField(Host),
@@ -3368,18 +3359,6 @@ class HostCollection(
             self._server_config,
             id=self.create_json(create_missing)['id'],
         ).read()
-
-    def update_payload(self, fields=None):
-        """Override the method to only update fields that actually can be updated
-        """
-        updatable_fields = set.intersection({'name', 'description', 'host_ids',
-                                            'max_hosts', 'unlimited_hosts'},
-                                            self.get_values().keys())
-        if fields is None:
-            fields = updatable_fields
-        else:
-            fields = set.intersection(set(fields), updatable_fields)
-        return super().update_payload(fields=fields)
 
 
 class HostGroup(
@@ -5877,6 +5856,7 @@ class PuppetClass(
     """A representation of a Puppet Class entity."""
 
     def __init__(self, server_config=None, **kwargs):
+        self.updatable_fields = ['name']
         self._fields = {
             'name': entity_fields.StringField(
                 required=True,
@@ -5956,16 +5936,6 @@ class PuppetClass(
         kwargs.update(self._server_config.get_client_kwargs())
         response = client.get(self.path('smart_variables'), **kwargs)
         return _handle_response(response, self._server_config, synchronous)
-
-    def update_payload(self, fields=None):
-        """Override the method to only update fields that actually can be updated
-        """
-        updatable_fields = set.intersection({'name'}, self.get_values().keys())
-        if fields is None:
-            fields = updatable_fields
-        else:
-            fields = set.intersection(set(fields), updatable_fields)
-        return super().update_payload(fields=fields)
 
 
 class PackageGroup(Entity, EntityReadMixin, EntitySearchMixin):
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -2031,6 +2031,17 @@ class ContentCredential(
         }
         super(ContentCredential, self).__init__(server_config, **kwargs)
 
+    def update_payload(self, fields=None):
+        """Override the method to only update fields that actually can be updated
+        """
+        updatable_fields = set.intersection({'name', 'content_type', 'content'},
+                                            self.get_values().keys())
+        if fields is None:
+            fields = updatable_fields
+        else:
+            fields = set.intersection(set(fields), updatable_fields)
+        return super().update_payload(fields=fields)
+
 
 class ContentUpload(
         Entity,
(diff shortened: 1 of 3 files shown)


lhellebr:
Regarding removed renaming of system_ids to system_uuids and a corresponding test: I believe this is obsolete. Sat 6.8's apidoc mentions neither system_uuids or system_ids as Host Collections API's parameters.

nailgun/entities.py
```diff
@@ -2031,6 +2031,17 @@ def __init__(self, server_config=None, **kwargs):
...
+    def update_payload(self, fields=None):
+        """Override the method to only update fields that actually can be updated
+        """
+        updatable_fields = set.intersection({'name', 'content_type', 'content'},
+                                            self.get_values().keys())
+        if fields is None:
+            fields = updatable_fields
+        else:
+            fields = set.intersection(set(fields), updatable_fields)
+        return super().update_payload(fields=fields)
```
mshriver, line 2043: Because ContentCredential has multiple inheritence, and because the method that you want to call here is in the last class that is inherited, I think it would be better to provide the type here to `super`, in order to be as explicit as possible and avoid the potential for this to resolve and call the wrong method from one of the other inherited classes.

`super(EntityUpdateMixin, self).update_payload()`

lhellebr: You're right. I'll fix it in bulk after I get all the feedback.

nailgun/entities.py
```diff
@@ -2031,6 +2031,17 @@ def __init__(self, server_config=None, **kwargs):
         }
         super(ContentCredential, self).__init__(server_config, **kwargs)
 
+    def update_payload(self, fields=None):
+        """Override the method to only update fields that actually can be updated
+        """
+        updatable_fields = set.intersection({'name', 'content_type', 'content'},
```
mshriver, line 2037: Any reason why you're doing the intersection here, instead of just setting these fields directly? Concern about a field getting removed from the entity?

nailgun/entities.py
```diff
@@ -2031,6 +2031,17 @@ def __init__(self, server_config=None, **kwargs):
         }
         super(ContentCredential, self).__init__(server_config, **kwargs)
 
+    def update_payload(self, fields=None):
```
mshriver, line 2034: I think this behavior would be great to implement within `EntityUpdateMixin.update_payload` directly, so that the behavior is available to all entities using this mixin.

Each entity could define a class attribute of update-able fields, with a default behavior (no attribute defined) of all fields being update-able. 

This way, you wouldn't have to add this to every entity, but for any entity where you want to limit the fields included in update_payload, you just define `Entity.updateable_fields` as a list of field names.

The `EntityUpdateMixin.update_payload` method can then filter its return on `self.updateable_fields`.

mshriver (changes requested):
Given that the method is already needed on multiple entities, I think it should be behavior built into `EntityUpdateMixin.update_payload` itself

nailgun/entities.py
lhellebr: Because I hit the issue with not all fields being set on initialization, leading to errors when calling the update_payload. However, I don't know *why* they weren't set.

nailgun/entities.py
lhellebr: Yes, I could do that. I wasn't sure if we still want to do such a big change.
Anyway, that would require quite a lot of manual work and testing. How about splitting it to three PRs, where one would implement the functionality for the mixin (with the default of updatable fields == all fields, which is the current state), another would specify the fields for the entities I need here and the last one would eventually specify them for all the remaining entities?

nailgun/entities.py
mshriver: Would be good to include that as an inline comment, so anybody else reading/using this method understands why its done this way.

nailgun/entities.py
mshriver: I don't think its necessary to specify updateable fields for every entity - the `EntityUpdateMixin.update_payload` could safely reference `self.updateable_fields` without it being defined on all entities. For the entities where it is defined, it would be used. For those without it defined, the current behavior would remain consistent. 

As far as testing goes, the existing unit test structure could be updated to cover the new behavior.  You wouldn't need to test it with every "real" entity in nailgun.
https://github.com/SatelliteQE/nailgun/blob/master/tests/test_entity_mixins.py#L850

I'd prefer to scope it into a single PR, as defining updateable fields on the three entities included in this PR would make the updates to EntityUpdateMixin actually useful.

lhellebr:
```
$ pytest tests/foreman/api/test_multiple_paths.py::EntityIdTestCase::test_positive_put_status_code
[...]
====================================== 1 passed, 3 warnings in 175.88 seconds =======================================
```

nailgun/entities.py
```diff
@@ -3358,13 +3360,6 @@ def create(self, create_missing=None):
             id=self.create_json(create_missing)['id'],
         ).read()
 
-    def update_payload(self, fields=None):
```
mshriver, line 3361: Is this conversion from `system_uuids` to `system_ids` no longer necessary?

nailgun/entity_mixins.py
```diff
@@ -979,6 +979,9 @@ def update_payload(self, fields=None):
 
         """
         values = self.get_values()
+        updatable_fields = (self.updatable_fields if hasattr(self, 'updatable_fields') else
```
mshriver, line 982: Consider a bit different syntax here:

```
updatable_fields = getattr(self, 'updatable_fields', None) or list(values.keys())
```

nailgun/entities.py
```diff
@@ -2007,6 +2007,7 @@ class ContentCredential(
     """A representation of a Content Credential entity."""
 
     def __init__(self, server_config=None, **kwargs):
+        self.updatable_fields = ['name', 'content_type', 'content']
```
mshriver, line 2010: This could be a class attribute, not an instance attribute, since its not defined on instantiation. Just a note though, works fine either way.

nailgun/entity_mixins.py
```diff
@@ -979,6 +979,9 @@ def update_payload(self, fields=None):
...
+                            values.keys())
+        values = {field: value for field, value in values.items() if field in updatable_fields}
```
mshriver, line 984: @lhellebr great work!! Thanks for taking the step to implement this for the mixin itself, this looks great, and will be easy to adjust any other entities that are limited in what fields can be updated.

I think this new 'feature' of the mixin should be documented in its docblock, so future users of the class have some instruction to set `entity.updatable_fields` if they want to limit what fields can be modified on the entity.

I would document this in the class docblock, as its relevant for anyone using the mixin class, and they're more likely to see it there first.

mshriver, line 984: I don't think the current unit test covers this behavior, can you please update the current unit test class for this mixin to include an update on a class with `updatable_fields` defined?

nailgun/entities.py
lhellebr: Since neither of those parameters is mentioned in 6.8's APIdoc, I suppose it's obsolete. (But I'm not the author of that workaround so I can't be absolutely sure.)

mshriver (changes requested):
The implementation looks great! I think we need some updates to the unit tests to make sure there's coverage, though test results from within robottelo look good.

nailgun/entities.py
lhellebr: Since we don't use class attributes anywhere, I think it's better to leave it as it is.

mshriver (approved):

lhellebr:
@SatelliteQE/nailgun-tier-2-reviewers

JacobCallahan (approved):
ACK

PR #758, opened 2020-10-12
Tracer fixes

ogajduse:
This PR should address all concerns that were brought up in #748.

PR #748, opened 2020-09-09
katello-tracer component implementation

ogajduse:
This PR brings the possibility to work with Host Traces.
4 new methods in Katello API => 4 nailgun methods.

> GET /api/hosts/:host_id/traces | List services that need restarting on the host
> PUT /api/hosts/:host_id/traces/resolve | Resolve Traces
> POST /api/hosts/bulk/traces | Fetch traces for one or more hosts
> PUT /api/hosts/bulk/resolve_traces | Resolve traces for one or more hosts

PR #759, opened 2020-10-20
[cherry-pick] katello-tracer component implementation and Tracer fixes

mirekdlugosz:
Cherry-picking #748 and #758 into `stable-satellite`. Both of them, since #758 contains fixes from review for #748, which was merged early.

There was a conflict in entities during cherry-picking. Might impact other pending cherry-picks if this is merged first, or this will need rebase if something else is merged first. Shouldn't stop reviewers from ACKing and merging, we'll sort it out along the way.


nailgun/entities.py
```diff
@@ -4249,17 +4319,18 @@ def path(self, which=None):
...
-            return '{0}/{1}'.format(
-                super(Host, self).path(which='base'),
-                which
-            )
-        elif which in ('bulk/add_subscriptions',):
+        elif which in ('bulk/install_content',
+                       'bulk/add_subscriptions',
+                       'bulk/remove_subscriptions',
+                       'bulk/traces',
+                       'bulk/resolve_traces'):
```
mirekdlugosz, line 4333: This removal is effect of conflict, let me know if I messed it up.

ogajduse: Method `bulk_remove_subscriptions` is missing in the cherry-pick. Look at the path `'bulk/remove_subscriptions'`.

https://github.com/SatelliteQE/nailgun/blob/49f3c4fa27b1cc3f1f4081e152d37cee722e2488/nailgun/entities.py#L4206-L4221

ogajduse (changes requested):

nailgun/entities.py
ogajduse: As the PR is related to Tracer, it would be probably better to introduce the missing method in other PR.

mirekdlugosz: Thanks for pointing that out. I see this method was added in #474, which has not been cherry-picked yet. I removed reference to that endpoint in this PR.

ogajduse (approved):

JacobCallahan (approved):
ACK

PR #763, opened 2020-11-02
[cherry-pick to stable] traces status ignored

tstrych:
cherry pick of #761

PR #761, opened 2020-10-29
traces status ignored

tstrych:
Fixing the tests, 

Needs cherry-pick to `stable branch` and to `6.8.z`

Connected to #748 and  #758 

If there is no traces_status it should be ignored, on the other hand if there are traces it should be [read](https://github.com/tstrych/nailgun/blob/b46f8f13326dd4abb0fd3d3f24cd25cc0184f15f/nailgun/entities.py#L4252)

Thanks to @ogajduse it was tested also on host with  traces attributes:
Command below is using the same read method as the robotello tests mentioned below.
` entities.Host().search(query={'search': "dhcp-3-206.vms.sat.rdu2.redhat.com"})[0].read()`
shows:
Screenshot from 2020-10-29 15-22-00

Test results: 
```
collected 30 items / 26 deselected / 4 selected
test_host.py::test_positive_search_by_parameter 2020-10-29 15:30:09 - robottelo - DEBUG - Finished Test: test_host.py/test_positive_search_by_parameter
test_host.py::test_positive_search_by_parameter_with_different_values 2020-10-29 15:31:42 - robottelo - DEBUG - Finished Test: test_host.py/test_positive_search_by_parameter_with_different_values
test_host.py::test_positive_search_by_parameter_with_prefix 2020-10-29 15:33:18 - robottelo - DEBUG - Finished Test: test_host.py/test_positive_search_by_parameter_with_prefix
test_host.py::test_positive_search_by_parameter_with_operator 2020-10-29 15:34:54 - robottelo - DEBUG - Finished Test: test_host.py/test_positive_search_by_parameter_with_operator
=========== 4 passed, 26 deselected, 2 warnings in 383.27s (0:06:23) ===========
```

PR #764, opened 2020-11-02
[cherry-pick 6.8.z] traces status ignored

tstrych:
master commit #761 and cherry pick to stable #763


ogajduse (approved):

ogajduse:
Merging as this is really small change that is not supposed to break things.

PR #767, opened 2020-11-06
Remove long deprecated entities and Version based logic blocks

mshriver:
Removed:
- Deprecated System and SystemPackage entities
- Entity method workarounds based on old versions of Satellite (6.1 and 6.2)
- server_modes
- Unit tests related to all of the above

## Test Results

<Jenkins Instance>/view/Satellite 6/job/satellite6-standalone-automation/2089/testReport/

- tests.foreman.api.test_contentview.TestContentViewRedHatContent  Has some failures in the standalone in test setup, a 403 on access to katello/api/v2/repository_sets, I don't believe this is related to my changes.
- Additional failures are in unmodified classes, I have confidence this PR doesn't have any negative impact on test results in robottelo. I'll followup with fixes for anything that does see an impact, but the test results give me confidence in the change.

### DIFF
--- a/nailgun/entity_mixins.py
+++ b/nailgun/entity_mixins.py
@@ -378,7 +378,8 @@ class Entity:
         # Check that a valid set of field values has been passed in.
         if not set(kwargs.keys()).issubset(self._fields.keys()):
             raise NoSuchFieldError(
-                f'Valid fields are {self._fields.keys()}, but received {kwargs.keys()} instead.'
+                f'Valid fields are {list(self._fields.keys())}, invalid field(s) passed '
+                f'to entity: {set(kwargs.keys()).difference(set(self._fields.keys()))}'
             )
 
         # Iterate through the values passed in and assign them as instance
--- a/tests/test_entities.py
+++ b/tests/test_entities.py
@@ -9,6 +9,7 @@ from http.client import NO_CONTENT
 from unittest import mock
 from unittest import TestCase
 
+from fauxfactory import gen_alpha
 from fauxfactory import gen_integer
 from fauxfactory import gen_string
 
@@ -77,7 +78,6 @@ class InitTestCase(TestCase):
     def setUpClass(cls):
         """Set a server configuration at ``cls.cfg``."""
         cls.cfg = config.ServerConfig('http://example.com')
-        cls.cfg_610 = config.ServerConfig('http://example.com', version='6.1.0')
 
     def test_init_succeeds(self):
         """Instantiate every entity.
@@ -174,8 +174,6 @@ class InitTestCase(TestCase):
                 entities.Status,
                 entities.Subnet,
                 entities.Subscription,
-                # entities.System,  # see below
-                # entities.SystemPackage,  # see below
                 entities.TailoringFile,
                 entities.TemplateCombination,
                 entities.Template,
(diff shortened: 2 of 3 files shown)


JacobCallahan (changes requested):
Just as a reminder for this PR, references to "sat" and "sam" as server_modes should be removed as well.

JacobCallahan:
Could you include the results of a standalone automation run for say activation keys?

tests/test_entities.py
```diff
@@ -69,7 +70,7 @@ class InitTestCase(TestCase):
     def setUpClass(cls):
         """Set a server configuration at ``cls.cfg``."""
         cls.cfg = config.ServerConfig('http://example.com')
-        cls.cfg_610 = config.ServerConfig('http://example.com', version='6.1.0')
+        cls.cfg_610 = config.ServerConfig('http://example.com', version='6.8.0')
```
swadeley, line 73: I wonder why you have `cls.cfg_610` not 680 here. I further wonder why you cannot just remove all the 6.1 stuff. Satellite 6.1 was quite different, but we do not need that code in master.

swadeley (commented):
Have you tested with just removing all the 6.1 related code?

tests/test_entities.py
mshriver: Just missed updating the name here.

As you'll see I have removed entire tests for some entities, but that was focused on which unit tests started failing, and whether or not they were still relevant. I'll take a second look at how relevant this test is and at a minimum update the name to be consistent.

mshriver: Some removals may take some input from component owners, or a whole lot of investigation by me - we've got entity mixin function overrides that per their docblocks are to workaround BZ's that were written and closed years ago.

swadeley (approved):
ACK

tests/test_entities.py
```diff
@@ -689,8 +639,8 @@ class CreatePayloadTestCase(TestCase):
     @classmethod
     def setUpClass(cls):
         """Set a server configuration at ``cls.cfg``."""
-        cls.cfg = config.ServerConfig('http://example.com')
-        cls.cfg_610 = config.ServerConfig('http://example.com', version='6.1.0')
+        cls.url = 'http://example.com'
+        cls.cfg = config.ServerConfig(cls.url)
```
jyejare, line 643: Since cls.url is not below, cant we keep that in one line ?
```suggestion
        cls.cfg = config.ServerConfig('http://example.com')
```

jyejare (dismissed):
ACK with non blocking improvements.

nailgun/entity_mixins.py
```diff
@@ -378,7 +378,8 @@ def __init__(self, server_config=None, **kwargs):
         # Check that a valid set of field values has been passed in.
         if not set(kwargs.keys()).issubset(self._fields.keys()):
             raise NoSuchFieldError(
-                f'Valid fields are {self._fields.keys()}, but received {kwargs.keys()} instead.'
+                f'Valid fields are {list(self._fields.keys())}, invalid fields(s) passed '
```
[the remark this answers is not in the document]
mshriver: I'd rather explicitly call list here than unpack and re-pack

nailgun/entity_mixins.py
```diff
@@ -378,7 +378,8 @@ def __init__(self, server_config=None, **kwargs):
...
+                f'to entity: {set(self._fields.keys()).difference(set(kwargs.keys()))}'
```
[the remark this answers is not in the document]
mshriver: I have the set difference backwards, good catch, but I wanted to improve the exception message here by actually informing the caller of what invalid keys were passed - with the current error message, it just lists all the keys and you have to go lookup the entity class to find out which one is invalid.

Current exception in master looks like:

image


I'm going to update my branch commit with the `set`s switched, so that the error message looks like this:
image

JacobCallahan (approved):

nailgun/entity_mixins.py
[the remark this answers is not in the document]
jyejare: Make Sense!

jyejare (approved):

PR #768, opened 2020-11-10
Adding timeout to synchronous functions

valleedelisle:
It might be interesting to tweak the timeout parameter on a per-function
level. This adds it to all functions that have a synchronous option.

### DIFF
--- a/tests/test_entity_mixins.py
+++ b/tests/test_entity_mixins.py
@@ -228,6 +228,15 @@ class PollTaskTestCase(TestCase):
                         entity_mixins._poll_task(gen_integer(), self.cfg),
                     )
 
+    def test__poll_task_timeout(self):
+        """What happens when a foreman task timesout?
+        Assert that the task is still running.
+        """
+        with self.assertRaises(entity_mixins.TaskTimedOutError):
+            with mock.patch.object(client, 'get') as get:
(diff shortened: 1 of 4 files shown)


mshriver:
I think its a fine parameter to add, but there are also synchronous method implementations in EntityMixin classes, please include those in the update for a consistent API for synchronous entity methods.

A direct unit test would be great for adding confidence to the change.

Moving to draft state, please move back to ready to review when unit tests are passing and all method signatures/implementations have been updated.

mshriver (approved):
Unit test looks like adequate coverage and the changes look consistent across entities and the mixin functions, thanks for this enhancement!  I'd definitely like to see at least one more review by a tier2/maintainer and clean CI run before merging.

tests/test_entity_mixins.py
```diff
@@ -228,6 +228,15 @@ def test__poll_task_success(self):
                         entity_mixins._poll_task(gen_integer(), self.cfg),
                     )
 
+    def test__poll_task_timeout(self):
```
swadeley, line 231:
```suggestion
    def test_poll_task_timeout(self):
```
Was the double underscore intentional?

valleedelisle: yes it is because we're testing the `_poll_task` function.

JacobCallahan (approved):
ACK

PR #770, opened 2020-11-12
Minor change to allow for incoming robottelo Satellite class

JacobCallahan:
As part of the Satellite class, entity class inits are modified
to include the server_config by default.
This seems to conflict with existing argument structure when an
entity is created during create_missing.
This change will try the usual method and fall back to not passing
a server config if the TypeError is encountered


mshriver (approved):

jyejare (approved):

swadeley (approved):
ACK

PR #777, opened 2020-12-02
add host_ids field to host collection

swadeley:
Hello

This update is to enable adding hosts to a host collection. 

This has not been working before, but the problem only came to light after https://github.com/SatelliteQE/nailgun/pull/757

### DIFF
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -3212,6 +3212,7 @@ class HostCollection(
         self._fields = {
             'description': entity_fields.StringField(),
             'host': entity_fields.OneToManyField(Host),
+            'host_ids': entity_fields.ListField(),
             'max_hosts': entity_fields.IntegerField(),
             'name': entity_fields.StringField(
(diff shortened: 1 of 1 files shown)


mirekdlugosz (commented):
I'm not convinced this is the way forward. `_payload` in `nailgun/entity_mixins.py` is responsible for translating `OneToManyField` field to `{name}_ids` that is list of ids from entities: https://github.com/SatelliteQE/nailgun/blob/master/nailgun/entity_mixins.py#L182

I think change like that should be applied on framework level, not on each entity level. There are probably many more entities currently broken.

@lhellebr @mshriver @JacobCallahan

swadeley:
> I think change like that should be applied on framework level,

@mirekdlugosz Unfortunately I do not understand what to do

JacobCallahan:
@swadeley he is recommending you add something more general to the most applicable mixin.

swadeley:
the test is now passing without this change https://github.com/SatelliteQE/robottelo/pull/8197#issuecomment-765912236

PR #786, opened 2021-06-22
removal of puppet modules

tstrych:
##### Description of changes

removal of puppet modules for satellite 6.10

it's only a removal, so there is no functionality to demonstrate

##### Additional Information
I did remove it based on change in adidoc. I **DID NOT** investigate it further. I would like to get ACK from team members that know more about puppet and puppet modules. 

the test is no longer failing when creating a Content View, but hitting other error connected to proxy. (fake capsule)

### DIFF
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -126,15 +126,13 @@ def _check_for_value(field_name, field_values):
 
     An entity may use this function in its ``__init__`` method to ensure that a
     parameter required for object instantiation has been passed in. For
-    example, in :class:`nailgun.entities.ContentViewPuppetModule`:
+    example, in :class:`nailgun.entities.HostPackage`:
 
     >>> def __init__(self, server_config=None, **kwargs):
-    >>>     _check_for_value('content_view', kwargs)
+    >>>     _check_for_value('host', kwargs)
     >>>     # …
     >>>     self._meta = {
-    >>>         'api_path': '{0}/content_view_puppet_modules'.format(
-    >>>             self.content_view.path('self')
-    >>>         )
+    >>>         'api_path': f'{self.host.path()}/packages',
     >>>     }
 
     :param field_name: A string. A key with this name must be present in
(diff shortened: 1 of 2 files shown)


tstrych:
Only place I do know about puppet modules stayed there is in example below:
`ContentViewPuppetModule` and `content_view_puppet_modules`
https://github.com/SatelliteQE/nailgun/blob/d7c2508dff22e7c3c2c9651c4aa6e7b3dd0c12a4/nailgun/entities.py#L129-L135

lhellebr:
@vsedmik @latran

vsedmik:
Looks good, comparing the apidocs it looks you covered all the puppet module removals from API.

One thing remains - we should probably also remove the puppet.*module references from test_entities.py.

You will need to rebase on and resolve conflicts with #785 

Thank you!

tstrych:
`make test` 

```
python -m unittest discover --start-directory tests --top-level-directory .
........../home/tstrych/.virtualenvs/nailgun/lib/python3.8/site-packages/packaging/version.py:127: DeprecationWarning: Creating a LegacyVersion has been deprecated and will be removed in the next major release
  warnings.warn(
...........................................................................................................................................................................................................................................................
Ran 261 tests in 6.119s

```

vsedmik (approved):
LGTM, thank you :+1:

lhellebr (approved):
Looks good, according to my limited knowledge of puppet changes in 6.10.
I suggest running some tests that are related to puppet modules, like CV tests, to make reasonably sure nothing got broken.
Note that Lai is working on removing the tests that test puppet modules from robottelo and airgun.

vsedmik:
Just for reference - the PR with the changes: https://github.com/Katello/katello/pull/9209

tstrych:
I believe running tests when you don't know which are really affected is like looking for needle in haystack. 
Of course we can try some of the test that were failing with puppet_module error. But I don't know about easy way, how I can list all those tests.
I think easier option is to merge it and look at test results.

mshriver:
@tstrych I'm inclined to agree with you here, and move a bit quicker with these changes now that we branched for 6.9.z.

Any tests that are using these entities will be failing already anyway, and we've got someone accountable to removing those tests.

JacobCallahan (approved):

PR #787, opened 2021-06-25
More changes needed for robottelo's config injection

JacobCallahan:
This is a continuation of the change introduced in #770
Robottelo injects the server config in the init methods of each entity.
As a result, passing the server config as part of the instantiation in
these places will result in multiple server configs being passed in.


mshriver (dismissed):

mshriver (approved):

latran (approved):
ACK

PR #792, opened 2021-10-07
Adding size attribute to ContentUpload

Griffin-Sullivan:
##### Description of changes

I'm wondering if this is a valid way to add the size attribute to ContentUpload. I have to add the two lines in the `read()` function in order for the `super.read()` to pass. However, I am now having issues with the checksum not matching when the import upload is ran after this. I would like to get some feedback if this is valid, or if it could be causing the checksum issue.

None that I could find.

Example:
```
attrs
{'upload_id': '50ba7ca9-fc60-43ba-9...2a8ed31594', 'size': -9148830520927570858}
```

### DIFF
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -2090,7 +2090,7 @@ class ContentUpload(
         if ignore is None:
             ignore = set()
         ignore.add('repository')
-        ignore.add('size') 
+        ignore.add('size')
         return super().read(entity, attrs, ignore, params)
 
     def update(self, fields=None, **kwargs):
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -2090,7 +2090,7 @@ class ContentUpload(
         if ignore is None:
             ignore = set()
         ignore.add('repository')
-        ignore.add('size')
+        ignore.add('size') 
         return super().read(entity, attrs, ignore, params)
 
     def update(self, fields=None, **kwargs):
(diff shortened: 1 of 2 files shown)


nailgun/entities.py
```diff
@@ -2056,6 +2056,7 @@ def __init__(self, server_config=None, **kwargs):
                 Repository,
                 required=True,
             ),
+            'size': entity_fields.IntegerField(required=True, min_val=2097152),
```
Griffin-Sullivan, line 2059: I'm setting `min_val=2097152` because `content_chunk_size = 2 * 1024 * 1024` on L2150. Maybe we should set these to a global variable?

JacobCallahan (dismissed):

nailgun/entities.py
swadeley: I think that would be best, move `content_chunk_size = 2 * 1024 * 1024` to top of the class and `min_val=content_chunk_size`.

jyejare:
@Griffin-Sullivan Closed and Reopened PR to run the checks! Now they are running !

jyejare:
@Griffin-Sullivan The checks are failing, please investigate and fix accordingly.

jyejare:
@Griffin-Sullivan Ping, The high pipeline tests failures are blocked on this PR, please get this in on priority.

jyejare (approved):

swadeley (approved):
ACK

jyejare (approved):

PR #796, opened 2021-11-11
Nailgun Github Action Checks Added

jyejare:
##### Description of changes

Github Actions added to Nailgun that's missing since Travis removed.


.github/workflows/merge_to_master.yml
```diff
@@ -0,0 +1,87 @@
...
+    PYCURL_SSL_LIBRARY: openssl
+
+jobs:
+  codechecks:
+    name: Code Quality
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        # python 3.6 is still used by IQE integration pipeline
+        python-version: [3.6, 3.8, 3.9]
```
JacobCallahan, line 19: let's add 3.10, perhaps dropping 3.8 or 3.9

.github/workflows/pull_requests.yml
```diff
@@ -0,0 +1,65 @@
...
+env:
+    PYCURL_SSL_LIBRARY: openssl
+
+jobs:
+  codechecks:
+    name: Code Quality
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.6, 3.8, 3.9]
```
JacobCallahan, line 17: same suggestion about 3.10

setup.py
```diff
@@ -43,6 +43,8 @@
         'Intended Audience :: Developers',
         'License :: OSI Approved :: GNU General Public License v3 or later (GPLv3+)',
         'Programming Language :: Python :: 3.6',
+        'Programming Language :: Python :: 3.8',
+        'Programming Language :: Python :: 3.9',
```
JacobCallahan, line 47: assuming 3.10 checks are good we can also add that here

JacobCallahan (changes requested):

.github/workflows/merge_to_master.yml
jyejare: Removed `3.9` and added `3.10` which now checks in even python versions of 6,8 and 10.

.github/workflows/pull_requests.yml
jyejare: Removed `3.9` and added `3.10` which now checks in even python versions of 6,8 and 10.

setup.py
jyejare: Removed `3.9` and added `3.10` which now checks in even python versions of 6,8 and 10.

JacobCallahan: Probably should have kept 3.9, but i don't think it will hurt anything to remove it

JacobCallahan (approved):

jameerpathan111 (approved):
LGTM

swadeley (commented):
ACK

PR #799, opened 2021-12-13
Add support for missing Puppet environment entity

swadeley:
##### Description of changes
Hello

Nailgun failed to create an org in Satellite 7 snap 2

`E           nailgun.entity_mixins.MissingValueError: Cannot find a value for the "environment" field.`

I presume due to removal of default Puppet environment support.

##### Upstream API documentation, plugin, or feature links

On 6.10.1 and Sat7.0

http://<satellite_server>/apidoc/v2/organizations/create.html
organization[environment_ids]optional , nil allowed

##### Functional demonstration

Provide an execution of the modified code, with ipython code blocks or screen shots.
You can exercise the code as raw API calls or using any other method.

```
pytest tests/foreman/api/test_organization.py -k test_positive_create_with_name_and_description
=================================== test session starts ==========================
platform linux -- Python 3.9.9, pytest-6.2.5, py-1.10.0, pluggy-0.13.1
shared_function enabled - OFF - scope:  - storage: file
rootdir: /home/sjw/sat6/robottelo, configfile: pyproject.toml
plugins: mock-3.6.1, ibutsu-1.16, forked-1.3.0, services-2.2.1, cov-3.0.0, reportportal-5.0.8, xdist-2.4.0
collected 43 items / 36 deselected / 7 selected                                                                                                                                                   

tests/foreman/api/test_organization.py .......  

============== 7 passed, 36 deselected, 15 warnings in 55.27s =======================
```

Your contribution should include updates to the unit tests, covering the modified portions or adding new coverage.

Example:
```
# Demonstrate functional Snapshot read in ipython
In [1]: from nailgun.entities import Snapshot
In [2]: Snapshot(host='sat.instance.addr.com', id='snap_uuid').read()
Out [2]: <read method result>
```

##### Additional Information

Any additional notes for reviewers, comments about the change, TODO lists on WIP PRs, etc.

PR #804, opened 2021-12-27
[#799 - v1] Remove environment param from Organization entity

jyejare:
##### Description of changes
This is version - 1 of #799 that was initially raised by @swadeley to remove the environment param from the organization entity. 

This PR is being raised to fix the tons of failures we have due to #799 is not yet merged or the discussions are still happening on that to support the optional puppet feature even though it's not the default anymore with sat 7.0 but still some components would enable them and use them until they are depreciated.

So let's merge this PR to unblock with the current count of failures vs later we can work on #799 to address the entities using environment/puppet facility with fewer failures.

Refer #799 for [results](https://github.com/SatelliteQE/nailgun/pull/799#issue-1078591248) and discussions.

### DIFF
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -5073,7 +5073,6 @@ class Organization(
             'compute_resource': entity_fields.OneToManyField(AbstractComputeResource),
             'description': entity_fields.StringField(),
             'domain': entity_fields.OneToManyField(Domain),
-            'environment': entity_fields.OneToManyField(Environment),
             'hostgroup': entity_fields.OneToManyField(HostGroup),
             'label': entity_fields.StringField(str_type='alpha'),
             'medium': entity_fields.OneToManyField(Media),


JacobCallahan (approved):

jameerpathan111:
@jyejare @JacobCallahan I don't think this will only fix issues for Organization entity and not others for ex. Location.
We should remove it for all other entities while we are at it otherwise tests would still keep on failing.

JacobCallahan:
@jameerpathan111 that's a good point, @jyejare thoughts?

jyejare:
@jameerpathan111 @JacobCallahan Even first I thought to remove all of them from all entities, but I see some of the components still shows the `environment` parameter in their API doc! So why only organizations and not others because an organization entity is a dependent entity of many tests, not locations or other entities have that much dependency!

I agree that the tests will still fail but not as much as we are having now and those can easily be taken care by component owners whether to keep environment param by default or to remove them ! Or we can do that as part of #799 for all such entities to be flexible!

This PR specifically is to unblock the tests which are very dependent on org entity and failing, even some CLI and UI tests are failing because there are using the org API entity.

jameerpathan111 (approved):
Giving Ack and merging it. 
Note: This is a temporary workaround.
Alternate solutions for this issue:

- https://github.com/SatelliteQE/nailgun/pull/799
- https://github.com/SatelliteQE/nailgun/pull/805/

PR #802, opened 2021-12-20
Fix puppet-related URLs

vsedmik:
##### Description of changes

The puppet-related URLs have changed with movement into a plugin in 7.0.

##### Upstream API documentation, plugin, or feature links

https://community.theforeman.org/t/the-road-to-making-puppet-optional/17983
or see the apidoc of a puppet-enabled Satellite

##### Functional demonstration

CRUD like the following with entities passed:
```
>>> from nailgun import entities
>>> cg = entities.ConfigGroup(name='ahoy').create()
>>> res = cg.read()
>>> res
nailgun.entities.ConfigGroup(name='ahoy', id=1)
>>> cg.name = 'astalavista'; res = cg.update(['name'])
>>> cg = entities.ConfigGroup().search(query={'search':f'name="astalavista"'})[0].read()
>>> cg
nailgun.entities.ConfigGroup(name='astalavista', id=1)
>>> res = cg.delete()
>>> res
{'id': 1, 'name': 'astalavista', 'created_at': '2021-12-20T15:29:41.261Z', 'updated_at': '2021-12-20T15:29:57.987Z'}
```

PR #805, opened 2021-12-28
Fixes issue related to disabled puppet feature

jameerpathan111:
##### Description of changes

Fixes issue related to puppet or any other disabled feature

##### Additional Information

Refer for more info:

- https://github.com/SatelliteQE/nailgun/pull/804
- https://github.com/SatelliteQE/nailgun/pull/799
- https://github.com/SatelliteQE/nailgun/pull/802

### DIFF
--- a/nailgun/entity_mixins.py
+++ b/nailgun/entity_mixins.py
@@ -782,6 +782,7 @@ class EntityReadMixin:
             attrs = self.read_json(params=params)
         if ignore is None:
             ignore = set()
+
         for field_name, field in entity.get_fields().items():
             if field_name == 'environment':
                 if field.entity.__name__ == 'Environment':
--- a/nailgun/entity_mixins.py
+++ b/nailgun/entity_mixins.py
@@ -62,6 +62,14 @@ class TaskFailedError(Exception):
         self.task_id = task_id
 
 
+def _feature_list(server_config, smart_proxy_id=1):
+    """Get list of features enabled on capsule"""
+    path = f'{server_config.url}/api/v2/smart_proxies/{smart_proxy_id}'
+    response = client.get(path, **server_config.get_client_kwargs())
+    response.raise_for_status()
+    return [feature['name'] for feature in response.json()['features']]
+
+
 def _poll_task(task_id, server_config, poll_rate=None, timeout=None):
     """Implement :meth:`nailgun.entities.ForemanTask.poll`.
 
@@ -782,7 +790,8 @@ class EntityReadMixin:
             attrs = self.read_json(params=params)
         if ignore is None:
             ignore = set()
-
+        if 'Puppet' not in _feature_list(self._server_config):
+            ignore.add('environment')
         for field_name, field in entity.get_fields().items():
             if field_name in ignore:
                 continue
--- a/nailgun/entity_mixins.py
+++ b/nailgun/entity_mixins.py
@@ -62,14 +62,6 @@ class TaskFailedError(Exception):
         self.task_id = task_id
 
 
-def _feature_list(server_config, smart_proxy_id=1):
-    """Get list of features enabled on capsule"""
-    path = f'{server_config.url}/api/v2/smart_proxies/{smart_proxy_id}'
-    response = client.get(path, **server_config.get_client_kwargs())
-    response.raise_for_status()
-    return [feature['name'] for feature in response.json()['features']]
-
-
 def _poll_task(task_id, server_config, poll_rate=None, timeout=None):
     """Implement :meth:`nailgun.entities.ForemanTask.poll`.
 
@@ -790,9 +782,10 @@ class EntityReadMixin:
             attrs = self.read_json(params=params)
         if ignore is None:
             ignore = set()
-        if 'Puppet' not in _feature_list(self._server_config):
-            ignore.add('environment')
         for field_name, field in entity.get_fields().items():
+            if field_name == 'environment':
+                if field.entity.__name__ == 'Environment':
+                    continue
             if field_name in ignore:
                 continue
             if isinstance(field, OneToOneField):
--- a/nailgun/entity_mixins.py
+++ b/nailgun/entity_mixins.py
@@ -698,6 +698,9 @@ class EntityReadMixin:
 
     """
 
+    ignore_fields = ['environment']
+    ignore_entities = ['Environment']
+
     def read_raw(self, params=None):
         """Get information about the current entity.
 
@@ -784,9 +787,8 @@ class EntityReadMixin:
             ignore = set()
 
         for field_name, field in entity.get_fields().items():
-            if field_name == 'environment':
-                if field.entity.__name__ == 'Environment':
-                    continue
+            if field_name in self.ignore_fields and field.entity.__name__ in self.ignore_entities:
+                continue
             if field_name in ignore:
                 continue
             if isinstance(field, OneToOneField):
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -5073,6 +5073,7 @@ class Organization(
             'compute_resource': entity_fields.OneToManyField(AbstractComputeResource),
             'description': entity_fields.StringField(),
             'domain': entity_fields.OneToManyField(Domain),
+            'environment': entity_fields.OneToManyField(Environment),
             'hostgroup': entity_fields.OneToManyField(HostGroup),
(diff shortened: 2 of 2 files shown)


jyejare:
This looks like an ultimate solution, can you just add some ipython lines with the component using environment param with puppet enabled sat and without puppet enabled sat!

jameerpathan111:
@jyejare I tried testing my original fix https://github.com/SatelliteQE/nailgun/pull/805/commits/60e81acb0bc956543aa5e8745608e1eaf9eaa19e on puppet enabled satellite and got the same issue there as well.
It says `environments` field is deprecated, not sure this is product bug or not.
Basically what's happening is that, we can pass puppet environment while creating an entity (ex. Location/Org) but when we will try to read that entity we won't get values of puppet environment and instead it'll give 
`E           nailgun.entity_mixins.MissingValueError: Cannot find a value for the "environment" field. Searched for keys named environment_ids, environment, environments but available keys are dict_keys(['select_all_types', 'description', 'created_at', 'updated_at', 'ancestry', 'parent_id', 'parent_name', 'id', 'name', 'title', 'users', 'smart_proxies', 'subnets', 'compute_resources', 'media', 'ptables', 'provisioning_templates', 'domains', 'realms', 'hostgroups', 'organizations', 'hosts_count', 'parameters', 'deprecations']). ` error.

`attrs = {'ancestry': None, 'compute_resources': [], 'created_at': '2021-12-30 08:44:16 UTC', 'deprecations': {'environments': 'Environments got deprecated from this endpoint.'}, ...}`

So this leaves us with no choice but to completely ignore it for both kind of satellite(with and without puppet enabled).
https://github.com/SatelliteQE/nailgun/pull/805/commits/57524222dc4befd48ff7dc9d623b7352b8e7f3f8 could be our fix for now till we figure out if this is expected or not.
maybe @vsedmik would know something about this.

jameerpathan111:
@jyejare Here's what I tried with new fix on both kind of Satellite:
```

(Pdb) entities.ActivationKey(name=key_name).create()
nailgun.entities.ActivationKey(auto_attach=True, content_view=None, description=None, environment=None, host_collection=[], max_hosts=None, name='Ic6Pczeag7', organization=nailgun.entities.Organization(id=77), purpose_usage=None, purpose_role=None, purpose_addons=[], release_version=None, service_level=None, unlimited_hosts=True, id=4)
(Pdb) entities.Environment(name=name).create()
nailgun.entities.Environment(location=[], name='KBy4g81YNC', organization=[], id=9)
(Pdb) entities.Environment().search(query={'search': f'name="{name}"'})
[nailgun.entities.Environment(name='KBy4g81YNC', id=9)]
(Pdb) org = entities.Organization().create()
(Pdb) loc = entities.Location(environment='1').create()
(Pdb) entities.ContentView(organization=org.id).create()
nailgun.entities.ContentView(auto_publish=False, component=[], composite=False, content_host_count=0, description=None, environment=[], label='bXLTdQeUotSY', last_published=None, name='bXLTdQeUotSY', next_version='1.0', organization=nailgun.entities.Organization(id=79), repository=[], solve_dependencies=False, version=[], id=54)
(Pdb) entities.Organization().search(query={'search': f'name="{org.name}"'})[0]
nailgun.entities.Organization(description=None, label='yGpTiewqH', name='yGpTiewqH', title='yGpTiewqH', id=79)
(Pdb) entities.Location().search(query={'search': 'name="Default Location"'})[0]
nailgun.entities.Location(description=None, name='Default Location', parent=None, id=2)
(Pdb) host = entities.Host().create()
(Pdb) entities.Host().search(query={'search': f'name="{host.name}"'})[0]
nailgun.entities.Host(architecture=nailgun.entities.Architecture(id=5), build=False, build_status_label='Installed', capabilities=['build'], comment=None, compute_profile=None, compute_resource=None, domain=nailgun.entities.Domain(id=4), enabled=True, environment=nailgun.entities.Environment(id=10), hostgroup=None, image=None, ip=None, location=nailgun.entities.Location(id=81), mac='42:23:6b:10:85:db', managed=True, medium=nailgun.entities.Media(id=17), model=None, name='xtmcilhphu.arx3aj0yrz', operatingsystem=nailgun.entities.OperatingSystem(id=4), organization=nailgun.entities.Organization(id=82), owner=nailgun.entities.User(id=4), provision_method='build', ptable=nailgun.entities.PartitionTable(id=213), puppet_ca_proxy=None, puppet_proxy=None, realm=None, subnet=None, uuid=None, id=5, owner_type='User')
(Pdb) entities.HostGroup(location=[loc], organization=[org]).create()
nailgun.entities.HostGroup(architecture=None, domain=None, puppet_proxy=None, puppet_ca_proxy=None, content_source=None, compute_profile=None, lifecycle_environment=None, location=[nailgun.entities.Location(id=80)], medium=None, name='KOOQmhvbuA', operatingsystem=None, organization=[nailgun.entities.Organization(id=79)], parent=None, ptable=None, realm=None, subnet=None, group_parameters_attributes=[], content_view=None, id=4)
```
Note: Some of the tests depends on https://github.com/SatelliteQE/nailgun/pull/802

nailgun/entity_mixins.py
```diff
@@ -784,6 +784,9 @@ def read(self, entity=None, attrs=None, ignore=None, params=None):
             ignore = set()
 
         for field_name, field in entity.get_fields().items():
+            if field_name == 'environment':
+                if field.entity.__name__ == 'Environment':
```
jyejare, line 788: Why is this double condition needed? Is this to separate the Lifecycle environment from Puppet Environment?

if not stick the lower condition only.

jyejare (commented):
@jameerpathan111 Yeah! Agree that the universal ignoring the environment makes more sense instead of removing it completely! Though its depreciated its not completely removed yet :)

We also might need to raise a bug and get the reading of environment param fixed but let's get this in to unblock other tests !

ACK with a curious important question!

nailgun/entity_mixins.py
jameerpathan111: @jyejare yeah, I don't like it either. But when I tried to use both of them together it failed with error saying `field` has no object `entity`, and that's why this workaround.

pondrejk (approved):

swadeley:
Hi, I checked out this PR and ran `tests/foreman/api/test_organization.py -k test_positive_create_with_name_and_description` which passes, so looks good to me.

swadeley (dismissed):
ACK

vsedmik:
I played with the puppet-enabled SAT and I can confirm what @jameerpathan111 wrote above (we don't get env for org but still we can set it to create org). I will discuss it with DEVs tomorrow and will update the details.

nailgun/entity_mixins.py
jyejare: @jameerpathan111 Cool! I am almost fell settled with this app! But,

Historically, we never set / unset things like this for components outside `entities.py` and hence we need to decide if this is the right place to globally ignore the depreciated fields/components from entity_mixins or any other module as this may look scattered. So, let's talk on JFT call if we should:
1. Go with this way to ignore depreciated fields globally / Base class. (Which I like and prefer)
2. Or, ignore the field from the entity classes read the definition, which probably dont need but if component owners(those who needs `environment` field) can take the decision.

Finally, @jameerpathan111 if we wanted to go with approach 1 modify the code as:
```
# Class level list of ignore / depreciated fields
ignore_fields = ['environment', 'etc']
```
and finally:
```suggestion
            if field_name in ignore_fields and field.entity.__name__.lower() in ignore_fields:
```

jameerpathan111: @jyejare entity can be referred using more than one name. For ex. LifecycleEnvironment is referred using `environment`, `library`, `prior` and `lifecycle_environment`. How about doing it like this:

```
# Class level list of ignore / depreciated fields
    ignore_fields = ['environment']  
    ignore_entity = ['Environment']
```

```
            if field_name in self.ignore_fields and field.entity.__name__ in self.ignore_entity:
                continue
```
And we can ignore the double if condition, it's working fine without it.

jameerpathan111: @jyejare I have updated the pr.

swadeley (approved):
ACK

pondrejk (approved):

pondrejk:
merging for the sake of tests passing, I'm sure @jameerpathan111 and @vsedmik will keep an eye on dev findings and revisit the issue if needed

PR #807, opened 2022-01-20
sat 7.0 environment fix

tstrych:
##### Description of changes
TLDR 
reverts #805 
intorucing the new way of reading environment

global ignore fields were problematic once the puppet was enabled via installer

with puppet enabled read the entity that uses environment, but read method already ignores the environment variable, so test fail, because asserts on environment has no values inside the object as result of ignorance

introducing new way 
during read if puppet is enabled there is environment it will be returned in `if` or `elif` clauses above 
if is not and attr is environment append the variable with empty  list and return it -> no more errors :+1: 

One think I am not sure about.  
Changed method `_get_entity_ids` is used also for search.  If you search for environment and there is none,  should the error be raised here ? I think it does not, but had not time to dig more.
Let me know what you think. 

I did test it with some tests

### DIFF
--- a/nailgun/entity_mixins.py
+++ b/nailgun/entity_mixins.py
@@ -247,6 +247,9 @@ def _get_entity_id(field_name, attrs):
             return attrs[field_name]['id']
     if field_name_id in attrs:
         return attrs[field_name_id]
+    elif field_name == 'environment':
+        attrs['environment'] = []
+        return attrs['environment']
     else:
         raise MissingValueError(
             f'Cannot find a value for the "{field_name}" field. '
@@ -280,6 +283,9 @@ def _get_entity_ids(field_name, attrs):
         return [entity['id'] for entity in attrs[field_name]]
     elif plural_field_name in attrs:
         return [entity['id'] for entity in attrs[plural_field_name]]
+    elif field_name == 'environment':
+        attrs['environment'] = []
+        return attrs['environment']
     else:
         raise MissingValueError(
             f'Cannot find a value for the "{field_name}" field. '
@@ -698,9 +704,6 @@ class EntityReadMixin:
 
     """
 
-    ignore_fields = ['environment']
-    ignore_entities = ['Environment']
-
     def read_raw(self, params=None):
         """Get information about the current entity.
 
@@ -787,8 +790,6 @@ class EntityReadMixin:
             ignore = set()
 
         for field_name, field in entity.get_fields().items():
-            if field_name in self.ignore_fields and field.entity.__name__ in self.ignore_entities:
-                continue
             if field_name in ignore:
                 continue
             if isinstance(field, OneToOneField):


vsedmik:
I tried to run Org and Loc tests with this PR against

1) puppet-enabled SAT
```
tests/foreman/api/test_organization.py ...........................................
tests/foreman/api/test_location.py ............................F....
```
The failure was `index out of range` on L226 as we don't get env back in read.

2) puppet-disabled SAT
```
tests/foreman/api/test_organization.py ...........................................
tests/foreman/api/test_location.py ............................E...E
```
The errors are `404 Client Error : Not Found for url: <SAT_FQDN>/foreman_puppet/api/environments`, which makes sense on disabled SAT.

Environment tests passed all (combined with robottelo PR#9286)
```
tests/foreman/api/test_environment.py ..............................
tests/foreman/cli/test_environment.py ...........................
```

I'm goning to try the no-env search next week.

jameerpathan111:
@tstrych @vsedmik can you give https://github.com/SatelliteQE/nailgun/compare/master...jameerpathan111:puppet_environment?expand=1 a try and see if it works for all the scenarios.

jameerpathan111 (approved):

vsedmik:
The search works good IMHO.
1. Against enabled SAT:
```
>>> entities.Environment().search(query={'search':f'name="env-that-does-not-exist"'})
[]
>>> entities.Environment().search(query={'search':f'name="production"'})
[nailgun.entities.Environment(name='production', id=1)]
>>> entities.Environment().search(query={'search':f'name="production"'})[0].read()
nailgun.entities.Environment(location=[nailgun.entities.Location(id=2)], name='production', organization=[nailgun.entities.Organization(id=1)], id=1)
```
2. Against disabled SAT:
```
>>> entities.Environment().search(query={'search':f'name="production"'})
WARNING:nailgun.client:Received HTTP 404 response: <!DOCTYPE html> 
...
  File "<input>", line 1, in <module>
  File "/home/vsedmik/PycharmProjects/nailgun/nailgun/entity_mixins.py", line 1350, in search
    results = self.search_json(fields, query)['results']
  File "/home/vsedmik/PycharmProjects/nailgun/nailgun/entity_mixins.py", line 1203, in search_json
    response.raise_for_status()
  File "/home/vsedmik/PycharmProjects/nailgun/venv/lib64/python3.9/site-packages/requests/models.py", line 953, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 404 Client Error: Not Found for url: https://dhcp-3-91.vms.sat.rdu2.redhat.com/foreman_puppet/api/environments
```

jyejare:
@vsedmik @tstrych Raising 404 against disabled puppet Sat is not expected.

@tstrych Have you handled that already ?

tstrych:
@jyejare 
I don't understand. Can you explain your point, and what would you expect?

I will explain mine.
so sat 7.0 with disabled puppet has no Environment entity, so if you reach such an endpoint I would expect 404 Not found

rplevka (approved):

jyejare:
@tstrych My worry was about the entities that use Puppet Environment and those entities arent aware whether they are running on Puppet Enabled satellite or not!

All such tests need to be handled to accept this error for hitting the wrong endpoint and mold the test steps accordingly.

jyejare:
@tstrych I see that a significant amount of effort has already been put together to isolate the tests using puppet to use the destructive_sat, and hence raising the endpoint error shouldn't be a worry for both isolated and non-puppet tests!

So merging this with my ACK now !

jyejare (approved):

PR #809, opened 2022-01-21
#808 Ignore mirror_on_sync

lhellebr:
Parameter mirror_on_sync has been added to API
in Katello PR 9834.
This is not returned as part of JSON on entity creation.
Consequently, read() fails.
It could be workarounded by ignoring errors when reading
non-required parameters. Also, this would be fixed if
Satellite started returning that value.
This solves the issue without unnecessary changes
in Nailgun's guts that might break something.

### DIFF
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -6198,6 +6198,7 @@ class Repository(
             ignore = set()
         ignore.add('organization')
         ignore.add('upstream_password')
+        ignore.add('mirror_on_sync')
         return super().read(entity, attrs, ignore, params)
 
     def create_missing(self):
--- a/tests/test_entities.py
+++ b/tests/test_entities.py
@@ -1270,7 +1270,7 @@ class ReadTestCase(TestCase):
                 {'discovery', 'remote_execution_proxy', 'subnet_parameters_attributes'},
             ),
             (entities.Subscription, {'organization'}),
-            (entities.Repository, {'organization', 'upstream_password'}),
+            (entities.Repository, {'organization', 'upstream_password', 'mirror_on_sync'}),
             (entities.User, {'password'}),
             (entities.ScapContents, {'scap_file'}),
             (entities.TailoringFile, {'scap_file'}),


jyejare (approved):
ACK

This is exactly where I wanted to introduce 'if issue_is_open' similar to 'if bz_bug_is_open' so that we can skip/run steps based on issue is open or not and we need not to worry about temporary fixing these kinds of issues and rolling it back once its fixed.

lhellebr:
@jyejare , do you thinks this is a bug in Satellite? I'm not sure myself. Is it reported already? I'm thinking about reporting it.

Griffin-Sullivan:
I asked about this last week. They replaced `mirror_on_sync` with `mirroring_policy`. Here's the PR: https://github.com/Katello/katello/pull/9834

My suggestion would be to add a note to this before adding it to ignore and then add the parameter for `mirroring_policy` with these options `additive`, `mirror_complete`, `mirror_content_only`.

jyejare:
@Griffin-Sullivan Yeah I remember we talk. So does the `mirror_on_sync` completely removed (from POST request as well), if it is, we can simply replace `mirror_on_sync` here with `mirroring_policy`!

Griffin-Sullivan:
@jyejare I see both in POST request, but `mirror_on_sync` says (deprecated). Last week when I was debugging I originally ran into this problem because `mirror_on_sync` wasn't being returned by `mirroring_policy` was. So far `mirroring_policy` in places of `mirror_on_sync` has worked for me locally.

jyejare:
@Griffin-Sullivan So does even `mirror_on_sync` gets executed with POST request successfully and is there any use of it since we have new option for same now?

If there is no use, I will simply replace in this PR !

Griffin-Sullivan:
@jyejare From my understanding `mirroring_policy` is replacing `mirror_on_sync` as it provides more options to the user. In the UI, you will find it fully replaced, but through the API both are still available.... for now. You should be able to use either one right now, but it seems like they are fully replacing `mirror_on_sync`. I'm going to link the PR to provide more answers: https://github.com/Katello/katello/pull/9834

lhellebr:
I'm still not sure how to best approach this. But I would merge this PR as is because:
1) mirror_on_sync is still present somewhere, just not in all calls
2) ignoring it makes sense, even more since it's present but deprecated
3) this PR fixes the issue
We can add mirrorring_policy in a separate PR once we know more.

lhellebr:
Let's ACK and merge than?

jyejare:
@lhellebr Agree with you to ACK and Merge this to avoid related issues in test execution and @Griffin-Sullivan already has PR opened(I think) for mirroring_policy so lets get it in.

jyejare (approved):

omkarkhatavkar (approved):
ACK

PR #816, opened 2022-03-01
Removed py3.6 from checks as InsightsQE has updated to 3.9

jyejare:
##### Description of changes

Removing the python 3.6 checks from GitHub Actions since it was being used by InsightsQE CI to execute the FiFi/Insights tests using nailgun having the python 3.6 version but now they have moved to python 3.9!

### DIFF
--- a/.github/workflows/merge_to_master.yml
+++ b/.github/workflows/merge_to_master.yml
@@ -15,8 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        # python 3.6 is still used by IQE integration pipeline
-        python-version: [3.6, 3.8, '3.10']
+        python-version: [3.8, '3.10']
     steps:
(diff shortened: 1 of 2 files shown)


swadeley (approved):
ACK

JacobCallahan (approved):

mshriver:
Thanks @jyejare !

mshriver:
@jyejare there is something I would have included here, and that's updating our packaging/metadata to reflect the python version support.

https://github.com/SatelliteQE/nailgun/blob/master/setup.py#L45

Our classifiers still include python 3.6 which is no longer true, and we don't specify a minimum python version via `python_requires = >=3.8` as a setuptools option - which I think we should.

jyejare:
@mshriver Agree ! Opened https://github.com/SatelliteQE/nailgun/pull/818

PR #817, opened 2022-03-01
Add token among host fields

vsedmik:
##### Description of changes

Just adding the `token` field in the Host class.

```
>>> entities.Host().search(query={'search': 'joey-altieri.vms.sat.rdu2.redhat.com'})[0].token
'6d361318-95e9-4d30-8543-bfa53242fd80'
```

##### Additional Information

I would like to use this field in a test where I would check rendered provisioning template for a puppet-specific content.


swadeley (approved):
ACK

mshriver (approved):

PR #833, opened 2022-06-22
Multiples configs

jyejare:
##### Description of changes

Extending #787 with auto-parent addition to the required entities for read and search.


jyejare:
@mshriver Nop, planning to do though.

PR #836, opened 2022-08-02
Auto CherryPick commits to previous branches

jyejare:
##### Description of changes

This PR adds the AutoCherryPick Github Actions for enforcing CherryPick labels and the cherrypick labeled PRs has been autocherrypicked to previous branches!

The parent PR to master should cherrypick PRs in n-2 branches , n-1 parent PR should open in n-1 branch and n-2 doesnt opens a cherrypick PR !

### DIFF
--- a/.github/workflows/required_labels.yml
+++ b/.github/workflows/required_labels.yml
@@ -1,0 +1,17 @@
+# CI jobs to check specific labels present / absent
+name: required_labels
+
+on:
+  pull_request:
+    types: [opened, labeled, unlabeled, synchronize]
+
+jobs:
+  cherrypick_label:
+    name: Enforcing cherrypick labels
+    runs-on: ubuntu-latest
+    steps:
+      - uses: mheap/github-action-required-labels@v2
+        with:
+          mode: exactly
+          count: 1
+          labels: "CherryPick, No-CherryPick"
--- a/.github/workflows/auto_cherry_pick.yml
+++ b/.github/workflows/auto_cherry_pick.yml
@@ -1,0 +1,43 @@
+# CI stages to execute against all branches on PR merge
+name: auto_cherry_pick_commits
+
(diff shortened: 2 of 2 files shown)


.github/workflows/required_labels.yml
```diff
@@ -0,0 +1,17 @@
...
+jobs:
+  cherrypick_label:
+    name: Enforcing cherrypick labels
+    runs-on: ubuntu-latest
+    steps:
+      - uses: mheap/github-action-required-labels@v2
+        with:
+          mode: exactly
+          count: 1
+          labels: "CherryPick, No-CherryPick"
```
ogajduse, line 17: Can we be less enforcing in terms of labels? Can't we consider the PR without a label being the case when the author does not want the change to be cherry-picked?

jyejare: Reasons :
- [Very important] People forget to cherrypick their changes back; hence, the automation of zStreams is affected! Especially in the case of framework level changes.
- Enforcing label is the **reminder** for authors to at least decide on whether it should be cherrypicked or not! After all its just adding a label right :) OR even if you are not able to decide right away just say `No-CherryPick` and you are done !
- Without Enforcing labels, there is no guarantee that people will always remember to add CherryPick labels for their PRs for AutoCherrypick GHA to work and hence adding AutoCherryPick GHA is useless until we have enforcing labels!

I hope that clears your doubt !

ogajduse: Yes, It does! Thanks for expanding on it, @jyejare!
It might be annoying for people to fill the label in, but on the other hand, there is a huge advantage of this enforcing policy that strengthens our processes and helps keep robottelo in shape.

ogajduse (approved):

JacobCallahan (approved):

PR #850, opened 2022-10-24
Truncate large strings being logged

JacobCallahan:
Nailgun has no size protection for large string being logged in requests. This can result in entire manifest contents being logged. This change introduces a recursive function to truncate strings to a maximum length of 500 characters.


swadeley (approved):
ACK

synkd (approved):
Tested and working as expected in PRT build #1402. Nice improvement!

PR #841, opened 2022-09-29
Add support for Ansible Role sync and assign

sambible:
Adding support for Ansible Role sync, as well as fetch/assign of Ansible Roles to a Host. 

Leaving in draft while I get a paired Robottelo PR that can exercise these endpoints

PR #881, opened 2023-01-17
Change path to point to proper endpoint

sambible:
Small update to Ansible Nailgun PR recently merged, this properly sets the path to the endpoint it needs to point to.

### DIFF
--- a/tests/test_entities.py
+++ b/tests/test_entities.py
@@ -325,7 +325,7 @@ class PathTestCase(TestCase):
             (entities.Host, 'packages'),
             (entities.Host, 'puppetclass_ids'),
             (entities.Host, 'smart_class_parameters'),
-            (entities.Host, 'list_ansible_roles'),
+            (entities.Host, 'ansible_roles'),
             (entities.Host, 'assign_ansible_roles'),
             (entities.HostGroup, 'clone'),
             (entities.HostGroup, 'puppetclass_ids'),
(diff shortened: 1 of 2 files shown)


tests/test_entities.py
```diff
@@ -325,7 +325,7 @@ def test_id_and_which(self):
             (entities.Host, 'packages'),
             (entities.Host, 'puppetclass_ids'),
             (entities.Host, 'smart_class_parameters'),
-            (entities.Host, 'list_ansible_roles'),
+            (entities.Host, 'ansible_roles'),
```
adarshdubey-star, line 328: @sambible can you please try running this test case locally and see if that works for you, I purposely changed that to `list_ansible_roles` as I think we are putting the methods here and it worked for me.

sambible: Yea, I've run this locally - in the path function it needs to be the actual name of the endpoint in Satellite - which is a GET on ansible_roles, not list_ansible_roles

adarshdubey-star (approved):

tests/test_entities.py
adarshdubey-star: PRT run passed with these changes in my open PR. Good catch @sambible

jyejare (approved):

Gauravtalreja1 (approved):
Ack, please add specific branch labels where you want this to be cherrypicked, also open manual cpick for #841

PR #888, opened 2023-01-31
poll for task intended to fail

pondrejk:
##### Description of changes
`_poll_task` did check if the task succeeded, which caused a problem for negative tests where we want wait on the task to finish  but we expect failure. This pr adds a parameter to make this behavior configurable, the default is kept as it was, though no need to update existing callers.

Will provide a robottelo pr using the parameter.
[update] see https://github.com/SatelliteQE/robottelo/pull/10591

### DIFF
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -3100,7 +3100,7 @@ class ForemanTask(Entity, EntityReadMixin, EntitySearchMixin):
             return f'{super().path(which="base")}/{which}'
         return super().path(which)
 
-    def poll(self, poll_rate=None, timeout=None):
+    def poll(self, poll_rate=None, timeout=None, must_succeed=True):
         """Return the status of a task or timeout.
 
         There are several API calls that trigger asynchronous tasks, such as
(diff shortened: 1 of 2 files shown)


nailgun/entities.py
```diff
@@ -3114,6 +3114,8 @@ def poll(self, poll_rate=None, timeout=None):
             ``nailgun.entity_mixins.TASK_POLL_RATE``.
         :param timeout: Maximum number of seconds to wait until timing out.
             Defaults to ``nailgun.entity_mixins.TASK_TIMEOUT``.
+        :param must_succeed: Don't rise error when task finishes with other then success
```
Griffin-Sullivan, line 3117:
```suggestion
        :param must_succeed: Don't raise error when task finishes with other than success
```

Griffin-Sullivan (dismissed):

nailgun/entity_mixins.py
```diff
@@ -130,10 +130,11 @@ def raise_task_timeout():  # pragma: no cover
         timer.cancel()
 
     # Check for task success or failure.
-    if task_info['result'] != 'success':
-        raise TaskFailedError(
-            f"Task {task_id} did not succeed. Task information: {task_info}", task_id
-        )
+    if must_succeed:
+        if task_info['result'] != 'success':
```
adamruzicka, line 134: Any reason for nesting the ifs?

pondrejk:
@adamruzicka @Griffin-Sullivan comments addressed, mind revisiting? Thanks!

Griffin-Sullivan (dismissed):

nailgun/entities.py
```diff
@@ -3114,6 +3114,8 @@ def poll(self, poll_rate=None, timeout=None):
...
+            result. Needed for negative tests
```
Gauravtalreja1, line 3118: As it is only required for negative tests, shouldn't this arg be set to False by default?

pondrejk: Bad wording once again. True is good default, ability to set this to False is needed for negative tests, updated

pondrejk:
@Gauravtalreja1 @Griffin-Sullivan mind revisiting?

PR #898, opened 2023-02-23
Add support for reading a Provisioning Template

shubhamsg199:
Adding support for reading the Provisioning Template

 **Reason for this addition**: There are several closed loop scenarios where we need to read the content of provisioning templates.


Griffin-Sullivan (dismissed):
Code looks good! Ack pending results.

ogajduse (approved):

PR #914, opened 2023-04-14
Upgrade codecov dependency

pnovotny:
~~codecov is no longer in PyPI.~~
UPDATE: codecov==2.1.12 has been removed from PyPI but developers uploaded
newer version 2.1.13 for those who still need it - like nailgun does.

More info:
https://about.codecov.io/blog/message-regarding-the-pypi-package/

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -16,4 +16,4 @@ wheel
 twine
 
 # For code coverage
-codecov==2.1.12
+codecov==2.1.13


pnovotny:
Ok, this looks like a stalemate. `codecov` package is no longer in PyPI, but nailgun uses it in the Code Quality (3.8) check - Test Nailgun Coverage.

lhellebr:
I don't have an opinion on whether we need it or not. But if we do, maybe get it from some other soure?

pondrejk (dismissed):

jyejare (changes requested):
Dont remove the codecov package but just update the version to the latest one. The codecov is not entirely removed but they removed 2.1.12 version. So use 2.1.13 version

See my PR works for codecov installation https://github.com/SatelliteQE/nailgun/pull/915/files#diff-2b4945591edfeaa4cf4d3f155e66d4b43d1bda7a55d881d5cf3107f1b05abbbc

jyejare:
@pnovotny ping, see my last comment !

pnovotny:
@jyejare sorry for the late reply. Good news that they've kept the package. I've updated the requirements to codecov 2.1.13

jyejare (approved):

jyejare:
@pnovotny The codecov warning to fix the integration between this repo and code coverage upload is now enabled.

jyejare:
Merging this PR as its failing checks for other PRs.

PR #922, opened 2023-05-08
Add needs_publish flag

sambible:
Simple PR, adds a new flag for the content-view publish audit feature.


sambible:
This will be exercised via https://github.com/SatelliteQE/robottelo/pull/11433

sambible:
Added another nailgun change here for the needs_publish feature, figured it didn't make sense to make another PR

nailgun/entities.py
```diff
@@ -2863,8 +2863,6 @@ def add(self, synchronous=True, timeout=None, **kwargs):
         if 'data' not in kwargs:
             # data is required
             kwargs['data'] = dict()
-        if 'component_ids' not in kwargs['data']:
```
sambible, line 2866: This isn't present in the API anymore for this endpoint ( and it might have never been? not sure if this was ever used in tests. )

sambible: anyways, as written it makes this endpolnt non-functional for at least the case where you want to use the `latest` flag

ogajduse: Can you tell which Sat version removed this attribute? If it is not present in the API, it might be a good idea to cherry-pick thic patch also to the respective `.z` branches.

ogajduse (commented):
Pending question. LGTM

nailgun/entities.py
jyejare: I think that must be 6.14 and I agree that it must be cherrypicked back to the right version!

sambible: I'll see if I can't walk back a few versions of satellite, and see where this was dropped.

sambible:
Alright, so with respect to the API for components changing: It hasn't. This portion that I removed from the Nailgun entity seems to be only something within Nailgun, rather than anything within satellite. This entity doesn't have any tests utilizing it currently, and I can't find anywhere there _were_ tests. So I think this is just something that's a nailgun cleanup, and should be backported to any version of Satellite we care about writing tests retroactively for.

ogajduse:
> Alright, so with respect to the API for components changing: It hasn't. This portion that I removed from the Nailgun entity seems to be only something within Nailgun, rather than anything within satellite. This entity doesn't have any tests utilizing it currently, and I can't find anywhere there _were_ tests. So I think this is just something that's a nailgun cleanup, and should be backported to any version of Satellite we care about writing tests retroactively for.

Thanks for looking into it. Funny that #920 (open at the moment) makes use of the import that you are removing.

Gauravtalreja1 (dismissed):
Ack, pending comment for cherrypick @ogajduse and add respective branch labels with Cherrypick label

sambible:
https://github.com/SatelliteQE/nailgun/issues/943

Gauravtalreja1 (approved):

ogajduse (approved):
ACK. Can you re-review @Gauravtalreja1?

ogajduse:
Shouldn't we cherry-pick this PR to 6.14, @sambible?

sambible:
We definitely need to, thanks for the catch @ogajduse  - this PR has been up for so long, I had hoped it would be in before branching haha.

sambible:
@jyejare can you merge this today please? It is ready to go :)

PR #923, opened 2023-05-11
Add cacert to .fields in HttpProxy [WIP]

damoore044:
##### Description of changes

Add cacert to .fields in http_proxy, which is the path within the host to the cacert for the proxy.

Provide an execution of the modified code, with ipython code blocks or screen shots.
You can exercise the code as raw API calls or using any other method.

Your contribution should include updates to the unit tests, covering the modified portions or adding new coverage.

Example:
```
# Demonstrate functional Snapshot read in ipython
In [1]: from nailgun.entities import Snapshot
In [2]: Snapshot(host='sat.instance.addr.com', id='snap_uuid').read()
Out [2]: <read method result>
```


vsedmik (changes requested):
IMHO you also need to add the field into the ignore list, as this field is not returned back on GET request and thus should be ignored when missing in the response:
```
[root@satellite ~]# curl -sku admin:changeme -H "Content-type: application/json" -X GET https://$(hostname)/api/http_proxies/1 | json_reformat
{
    "id": 1,
    "name": "AwesomeProxy",
    "url": "http://my.awesome.proxy.com:3128",
    "username": null,
    "locations": [
        {
            "id": 2,
            "name": "Default Location",
            "title": "Default Location"
        }
    ],
    "organizations": [ ]
}
```
just few lines lower:
```
if ignore is None:
    ignore = set()
ignore.add('cacert')
ignore.add('password')
...
```

Griffin-Sullivan:
Yeah I think @damoore044  was about to try curl to see what is returned from the GET. For reference David, nailgun entities are designed to look for any of the attributes defined in `__init__()` when receiving a GET request (in this case the `read` function). So for this instance, we defined the `cacert` field in the `__init__`, and now when we try to run something like `proxy.read()` it will go through each field listed in the `init` and see what value was returned for that field. However, the GET request to the proxy endpoint doesn't include `cacert`, so when nailgun doesn't find it we get a `KeyError`. This situation is pretty common, especially with things like `password` since we don't want it to be readable. So when this happens, we have a set named `ignore` which we can add our key to so that nailgun knows to not look for it when we run the `read`.

Here's the code that implements the logic I just explained: https://github.com/SatelliteQE/nailgun/blob/master/nailgun/entity_mixins.py#L777

All you need to do is add `ignore.add('cacert')`, to the HTTPProxy's `read` method.

jyejare:
Very nice explanation @Griffin-Sullivan !!

@damoore044 Please do it before we merge !

jyejare (changes requested):
ACK pending @vsedmik comment for ignoring the field in read.

damoore044:
Thanks @Griffin-Sullivan for the helpful explanation, just addressed those

Griffin-Sullivan (approved):

vsedmik (approved):

jyejare (approved):

PR #967, opened 2023-07-24
Remove Python 3.9 support from nailgun

LadislavVasina1:
##### Description of changes

As discussed on slack, we are removing support for Python 3.9 from our frameworks.

### DIFF
--- a/.github/workflows/pull_requests.yml
+++ b/.github/workflows/pull_requests.yml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.8, '3.10']
+        python-version: ['3.10', '3.11']
     steps:
       - name: Checkout Nailgun
         uses: actions/checkout@v2


setup.py
```diff
@@ -42,8 +42,8 @@
         'Development Status :: 4 - Beta',
         'Intended Audience :: Developers',
         'License :: OSI Approved :: GNU General Public License v3 or later (GPLv3+)',
-        'Programming Language :: Python :: 3.8',
         'Programming Language :: Python :: 3.10',
+        'Programming Language :: Python :: 3.11',
     ],
     packages=find_packages(exclude=['docs', 'tests']),
     install_requires=REQUIREMENTS,
```
jyejare, line 49: The Below python requires minimum version needs to be updated to 3.10

jyejare (changes requested):
1. `Build and Publish` job in merge-to-master GHA needs to change the python-version to 3.10 now.
2. setup.py also needs to update the `python-requires` to minimum 3.10

adarshdubey-star (approved):

Gauravtalreja1 (approved):

PR #994, opened 2023-09-13
Introduce docstring standards for Ruff

JacobCallahan:
Most rules are used, but a few are ignored for style.


ogajduse (commented):
Two comments left.

ogajduse (approved):
Thank you, Jake!

sambible (approved):
i love ruff already

PR #995, opened 2023-09-14
Removing auto release WF from non-master branches

jyejare:
##### Description of changes

Removing the WF for `Automatic Releases` from non-master branches as only master branch commits are considered in Github Releases!

This was been cherrypicked into 6.14.z during the Branching process which should have been removed !


ogajduse (approved):
👍 
If we have a documentation for branching, adding an extra step to remove the file from the newly branched `.z` branch would be a good idea.

Gauravtalreja1 (approved):

PR #1002, opened 2023-09-26
Version 6.15.0

lpramuk:
We lack proper versioning of nailgun module. For all branches we use currently it says just `0.32.0` 
Lets get used to updating nailgun package version after each branching/release.
It will be more obvious from what branch we installed the nailgun.

Keeping 6.15.0 in master as nigtly version/future version so being compatible with current tagging.

BTW
For multiple nailguns being installed at the same time we need proper versioning.
`pip download` saves nailgun from different branches to the very same file `nailgun-0.32.0.zip`. Oops.


jyejare:
@lpramuk We have [releases](https://github.com/SatelliteQE/nailgun/releases) auto-created but that is for the master branch only and is being created to understand the changelogs per snap.

This looks like a good idea but we had to update the https://github.com/SatelliteQE/nailgun/blob/master/.github/workflows/merge_to_master.yml to update PyPI package for each separate version.

lpramuk:
> @lpramuk We have [releases](https://github.com/SatelliteQE/nailgun/releases) auto-created but that is for the master branch only and is being created to understand the changelogs per snap.
> 
> This looks like a good idea but we had to update the https://github.com/SatelliteQE/nailgun/blob/master/.github/workflows/merge_to_master.yml to update PyPI package for each separate branch version.

Sure. Module version doesn't go against this or does ?

JacobCallahan:
i'm not opposed to the change in this PR, but I don't think it solves the problem of multiple nailgun versions. it's almost certainly better to replace the currently installed version of nailgun every time with the latest branch using pip git installs directly.
see: https://github.com/SatelliteQE/robottelo/pull/12840

PR #1102, opened 2024-02-21
Remove codecov

JacobCallahan:
This continues a trend we've done with other projects. Codecov has some use in other types of projects, but really isn't a good fit for Nailgun since it is functionally exercised by Robottelo.

Without getting too deep into philosophy, CodeCov is a misleading metric for quality and minor changes are largely ignored in mature projects.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,7 +1,0 @@
-github_checks:
-  annotations: false
-
-# https://docs.codecov.io/docs/github-checks-beta
-# Annotations are shown for unchanged files on the diff
-# Turn off all annotations until this is more configurable
-# The codecov comment will still be made on the PR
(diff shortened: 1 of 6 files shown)


omkarkhatavkar (approved):

jyejare:
Here are some personal thoughts:

TBH we caught many issues in the past in nailgun using unit tests before merge the PR.

I agree that its an extra overhead but nailgun has simplified form of unit tests where sometimes adding just a line covers the new function.

Also In SatelliteQE testing, we are very dependent on API testing and unit tests here serves as a sanity for automation failures in API functions added in nailgun.

JacobCallahan:
I'm not saying get rid of the unit tests, I'm saying get rid of codecov as it's a poor measurement of useful tests.

jyejare:
Sure @JacobCallahan , but then could we run tests as part of Pull Request WF to ensure none is failing ?

JacobCallahan:
@jyejare oohh I didn't even notice that we weren't running tests. Added that to the pull request workflow and they're passing.

jyejare (approved):

ogajduse (approved):

PR #1143, opened 2024-05-23
Use type(self) instead of entity class name in create methods

tpapaioa:
##### Description of changes

Before this PR, the entity classes that override the `EntityCreateMixin.create` method return a new instance of the `nailgun.entities.ENTITY` class. In robottelo, which wraps each entity class in  the `robottelo.hosts.DecClass` class, this causes test failures, because `satellite.api.ENTITY().create()` returns an instance of `robottelo.hosts.DecClass` for some classes and `nailgun.entities.ENTITY` instances for others.

This PR updates those `create` methods to return an instance of `type(self)` instead of explicitly naming the class, so that an the same subclass / decorated class is used.

No functionality changes.

### DIFF
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -1602,7 +1602,7 @@ class DiscoveryRule(
         <https://bugzilla.redhat.com/show_bug.cgi?id=1381129>`_.
 
         """
-        return DiscoveryRule(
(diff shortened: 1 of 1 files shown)


JacobCallahan (approved):
ACK, nice!

PR #1203, opened 2024-07-02
Add support for debian repository creation

dosas:
##### Description of changes

Add support for debian repository creation

All tests are green

##### Additional Information

I'm not sure if there should be explicit tests for debian repo creation?

### DIFF
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -6833,7 +6833,7 @@ class Repository(
             ),
             'content_counts': entity_fields.DictField(),
             'content_type': entity_fields.StringField(
-                choices=('puppet', 'yum', 'file', 'docker', 'ostree'),
+                choices=('puppet', 'yum', 'file', 'docker', 'ostree', 'deb'),
                 default='yum',
                 required=True,
             ),
@@ -6874,12 +6874,17 @@ class Repository(
                 choices=('global_default_http_proxy', 'none', 'use_selected_http_proxy')
             ),
             'http_proxy_id': entity_fields.IntegerField(),
+            'deb_releases': entity_fields.StringField(),
+            'deb_components': entity_fields.StringField(),
+            'deb_architectures': entity_fields.StringField(),
         }
         if self._fields['content_type'].choices == 'yum':
             self._fields['download_policy'].required = True
         self._meta = {
             'api_path': 'katello/api/v2/repositories',
         }
+        if kwargs.get('content_type') == 'deb':
+            self._fields['deb_releases'].default = 'stable'
         super().__init__(server_config=server_config, **kwargs)
 
     def path(self, which=None):


omkarkhatavkar:

nailgun/entities.py
```diff
@@ -6874,12 +6874,19 @@ def __init__(self, server_config=None, **kwargs):
...
+            'deb_errata_url': entity_fields.StringField(),
+            'deb_components': entity_fields.StringField(),
+            'deb_architectures': entity_fields.StringField(),
         }
         if self._fields['content_type'].choices == 'yum':
             self._fields['download_policy'].required = True
         self._meta = {
             'api_path': 'katello/api/v2/repositories',
         }
+        if self._fields['content_type'].choices == 'deb':
```
dosas, line 6887: This does not seem to work:

```
ipdb> self._fields['content_type'].choices
('puppet', 'yum', 'file', 'docker', 'ostree', 'deb')

```

same for above in L6882

dosas:
@omkarkhatavkar Should I ping someone for review?

nailgun/entities.py
```diff
@@ -6874,12 +6874,18 @@ def __init__(self, server_config=None, **kwargs):
                 choices=('global_default_http_proxy', 'none', 'use_selected_http_proxy')
             ),
             'http_proxy_id': entity_fields.IntegerField(),
+            'deb_releases': entity_fields.StringField(),
+            'deb_errata_url': entity_fields.StringField(),
```
ogajduse, line 6878: @dosas Can you please reference the API doc, assuming it is publicly accessible, that references `deb_errata_url`? I could not find it at the first try.

dosas: All I can give you right now is an open PR https://github.com/Katello/katello/pull/7961/files#diff-8a86b1f719540a398a66ebd0d9bedc010e40e5d8fb0f516464bb36e93f9adba9R26

ogajduse (changes requested):
This is an interesting situation here. I checked out your code and tried to create a repository.
```python
[ins] In [1]: from robottelo.hosts import ContentHost, Satellite, Capsule, IPAHost

[ins] In [2]: sat = Satellite.get_host_by_hostname('mysatellite.acme.com')

[ins] In [8]: org = sat.api.Organization(id=1).read()

[ins] In [12]: product = sat.api.Product(organization=org).create()

[ins] In [13]: sat.api.Repository(product=product).create()
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
Cell In[13], line 1
----> 1 sat.api.Repository(product=product).create()

File ~/repos/nailgun/nailgun/entity_mixins.py:987, in EntityCreateMixin.create(self, create_missing)
    961 def create(self, create_missing=None):
    962     """Create an entity.
    963 
    964     Call :meth:`create_json`, use the response to populate a new object
   (...)
    985 
    986     """
--> 987     return self.read(attrs=self.create_json(create_missing))

File ~/repos/nailgun/nailgun/entities.py:6948, in Repository.read(self, entity, attrs, ignore, params)
   6946 ignore.add('upstream_password')
   6947 ignore.add('mirror_on_sync')
-> 6948 return super().read(entity, attrs, ignore, params)

File ~/repos/nailgun/nailgun/entity_mixins.py:846, in EntityReadMixin.read(self, entity, attrs, ignore, params)
    844         setattr(entity, field_name, referenced_entities)
    845     else:
--> 846         setattr(entity, field_name, attrs[field_name])
    847 return entity

KeyError: 'deb_errata_url'
```

The above happens when you run nailgun with your patch against a Satellite/Foreman instance that does not have the patch [you referenced](https://github.com/SatelliteQE/nailgun/pull/1203/files#r1696476552)  applied. The fix here is to add `deb_errata_url` to `ignore` as follows:

```diff
diff --git a/nailgun/entities.py b/nailgun/entities.py
index e20055a..ae2d070 100644
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -6942,6 +6942,9 @@ class Repository(
         """
         if ignore is None:
             ignore = set()
+        if 'deb_errata_url' not in attrs:
+            ignore.add('deb_errata_url')
+
         ignore.add('organization')
         ignore.add('upstream_password')
         ignore.add('mirror_on_sync')
```

Although it is not nice, I believe that this is the only way around it with the current nailgun structure/design.

ogajduse:
I believe that we should remove all labels that are currently set on this PR and use `NoCherryPick` as this looks like something that will not be backported to any of the released versions. Can you please confirm, @dosas?
However, if we add that ignore as I suggested and it does not break on the older versions, I do not see a problem in merging it to the other `.z` branches.

ogajduse:
Adding @SatelliteQE/team-phoenix since the patch is related to Repositories component of the product.

dosas:
@ogajduse Thanks for the review. That is a good catch, I did not consider that. I will remove the `deb_errata_url` until the above mentioned PR is merged.

ogajduse:
> I believe that we should remove all labels that are currently set on this PR and use `NoCherryPick` as this looks like something that will not be backported to any of the released versions. Can you please confirm, @dosas? However, if we add that ignore as I suggested and it does not break on the older versions, I do not see a problem in merging it to the other `.z` branches.

@dosas Can you please comment on this one?

ogajduse (approved):

jameerpathan111 (approved):

dosas:
> > I believe that we should remove all labels that are currently set on this PR and use `NoCherryPick` as this looks like something that will not be backported to any of the released versions. Can you please confirm, @dosas? However, if we add that ignore as I suggested and it does not break on the older versions, I do not see a problem in merging it to the other `.z` branches.
> 
> @dosas Can you please comment on this one?

Since I'm using 6.15.z downstream I would like to backport it to this branch but the other labels can be removed.

PR #1209, opened 2024-07-30
Add support for delete with args, and republish_repositories endpoints

sambible:
Adding support for one new endpoint, republish_repositories, as well as adding in a newly added field, is_container_push to support the podman push feature for 6.16.

Additionally, this adds a version of the delete endpoint for a repository that respects arguments passed to it. There certainly could be a more elegant way to do this, perhaps reworking the Delete mixin entirely to support arguments, but as far as I'm aware this isn't something that is required in the vast case of all entities, so I decided to make a more local change, and just add a separate endpoint for repositories. 

This is in support of this robottelo PR: https://github.com/SatelliteQE/robottelo/pull/14940


tests/test_entities.py
```diff
@@ -344,6 +345,7 @@ def test_id_and_which(self):
             (entities.Product, 'sync'),
             (entities.PuppetClass, 'smart_class_parameters'),
             (entities.Repository, 'docker_manifests'),
+            (entities.Repository, 'delete_with_args'),
```
JacobCallahan, line 348: This is failing because it isn't included in the entity's which tuple. Likely doesn't need to be added here since it isn't a true path.

JacobCallahan (dismissed):

vsedmik (approved):
LGTM

nailgun/entities.py
```diff
@@ -2684,6 +2686,25 @@ def verify_checksum(self, synchronous=True, timeout=None, **kwargs):
         response = client.post(self.path('verify_checksum'), **kwargs)
         return _handle_response(response, self._server_config, synchronous, timeout)
 
+    def republish_repositories(self, synchronous=True, timeout=None, **kwargs):
+        """Force a republish of the version's repositories' metadata.
```
vijaysawant, line 2690:
```suggestion
        """Force a republish of the version's repositories metadata.
```

vijaysawant (approved):
Small change require, Ack.

JacobCallahan (approved):

ColeHiggins2 (approved):

PR #1210, opened 2024-08-14
Add repo_data option to host registration command

dosas:
##### Description of changes

Add repo_data option to host registration command

The foreman upstream api is not up to date so I can only link the PR
https://github.com/theforeman/foreman/pull/10162

I am executing it via robottelo I do not know how to extract the raw api call.

##### Additional Information
None

### DIFF
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -6815,6 +6815,7 @@ class RegistrationCommand(Entity, EntityCreateMixin, EntityReadMixin):
             'update_packages': entity_fields.BooleanField(default=False),
             'force': entity_fields.BooleanField(default=False),
             'ignore_subman_errors': entity_fields.BooleanField(default=False),
+            'repo_data': entity_fields.ListField(),
         }
 
         self._meta = {'api_path': '/api/registration_commands'}


omkarkhatavkar:

dosas:
> @dosas When would you like to get this PR merged? Before the upstream PR is merged or after?
> 
> IIRC, we nailgun [can not define](https://github.com/SatelliteQE/nailgun/pull/1203#pullrequestreview-2212993895) fields that are not implemented in API unless an `ignore` exists for that field.

The above mentioned PR is already merged since a few months. The field should be part of the latest foreman?

nailgun/entities.py
```diff
@@ -6795,6 +6795,7 @@ def __init__(self, server_config=None, **kwargs):
             'update_packages': entity_fields.BooleanField(default=False),
             'force': entity_fields.BooleanField(default=False),
             'ignore_subman_errors': entity_fields.BooleanField(default=False),
+            'repo_data': entity_fields.ListField(),
```
Gauravtalreja1, line 6798: @dosas Thanks for adding this. Yes, I saw this option being added in foreman recently, and as per apidoc I see it is a `Array with repository URL and corresponding GPG key URL`, so I think ListField might not work here, so could you try `DictField` instead if it doesn't?

Also, I was wondering if you had any robottelo tests covering the repo_data? To verify this, we could run the PRT with that?

dosas: IMO it is a list of dicts, so `ListField` should be correct. I have robottelo tests using this feature but they are not upstream.

The tests that are using [`api_register`](https://github.com/search?q=repo%3ASatelliteQE%2Frobottelo+api_register&type=code) in robottelo do not use `repo_data` currently.

dosas: ping @Gauravtalreja1

amolpati30: I checked by creating a command in the UI, adding the repository and URL, and then compared it with the command I created through the API. Both commands appear identical. As Dosas mentioned, it is a list of dictionaries. I tested it with the output below, and it should work.

```
module_target_sat.api.RegistrationCommand(
        organization=module_org,
        location=module_location,
        activation_keys=[module_ak_with_cv.name],
        repo_data=[{'repo':repo1.name,'repo_gpg_key_url':'url1'}, {'repo':repo2.name,'repo_gpg_key_url':'url2'}]
    ).create()
```

amolpati30: Based on my above comment, I am approving this. However, I would prefer to have one more review.

amolpati30 (approved):

nailgun/entities.py
shubhamsg199: Thanks @amolpati30 , I've tested this locally as well and `ListField` is correct here.

shubhamsg199:
@dosas Can you please rebase the branch?

PR #1225, opened 2024-09-05
Fix for registration command packages

dosas:
it is a space separated string not a list
https://apidocs.theforeman.org/foreman/latest/apidoc/v2/registration_commands/create.html

### DIFF
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -6811,7 +6811,7 @@ class RegistrationCommand(Entity, EntityCreateMixin, EntityReadMixin):
             'jwt_expiration': entity_fields.IntegerField(default=4),
             'repo': entity_fields.StringField(default=''),
             'repo_gpg_key_url': entity_fields.URLField(default=''),
-            'packages': entity_fields.ListField(default=[]),
+            'packages': entity_fields.StringField(),
             'update_packages': entity_fields.BooleanField(default=False),
             'force': entity_fields.BooleanField(default=False),
             'ignore_subman_errors': entity_fields.BooleanField(default=False),


omkarkhatavkar:

nailgun/entities.py
```diff
@@ -6792,7 +6792,7 @@ def __init__(self, server_config=None, **kwargs):
             'jwt_expiration': entity_fields.IntegerField(default=4),
             'repo': entity_fields.StringField(default=''),
             'repo_gpg_key_url': entity_fields.URLField(default=''),
-            'packages': entity_fields.ListField(default=[]),
+            'packages': entity_fields.StringField(),
```
Gauravtalreja1, line 6795: @dosas Good catch, this field should only be a string field, but robottelo uses it as a list field only, will you be opening a PR for that too?

dosas: Shouldn't some tests be failing if packages was passed?

I cannot find any usage of packages in robottelo tests. 

`RegistrationCommand` is called by `api_register` : https://github.com/search?q=repo%3ASatelliteQE%2Frobottelo%20api_register&type=code

dosas: ping @Gauravtalreja1

Gauravtalreja1: ah, right, we've no registration tests for api's to check if packages works

Gauravtalreja1: Created PR to cover this in robottelo test - https://github.com/SatelliteQE/robottelo/pull/16914

Gauravtalreja1 (approved):
ACK, PRT passed on https://github.com/SatelliteQE/robottelo/pull/16914

dosas:
@Gauravtalreja1 Thank you for the review. DO you know anybody from your team who could supply the second approval?

shweta83 (changes requested):
Ack

shweta83 (approved):
Ack

PR #1257, opened 2024-11-27
Enable download utility selection

dosas:
##### Description of changes

Add download_utility property to RegistrationCommand

Option was added to foreman recently https://github.com/theforeman/foreman/blob/develop/app/controllers/api/v2/registration_commands_controller.rb#L28

### DIFF
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -6926,6 +6926,7 @@ class RegistrationCommand(Entity, EntityCreateMixin, EntityReadMixin):
             'force': entity_fields.BooleanField(default=False),
             'ignore_subman_errors': entity_fields.BooleanField(default=False),
             'repo_data': entity_fields.ListField(),
+            'download_utility': entity_fields.StringField(default='curl', choices=('curl', 'wget')),
         }
 
         self._meta = {'api_path': '/api/registration_commands'}


Satellite-QE:

nailgun/entities.py
```diff
@@ -6815,6 +6815,7 @@ def __init__(self, server_config=None, **kwargs):
             'force': entity_fields.BooleanField(default=False),
             'ignore_subman_errors': entity_fields.BooleanField(default=False),
             'repo_data': entity_fields.ListField(),
+            'download_utility': entity_fields.StringField(default=''),
```
vsedmik, line 6818: I would rather go with
```suggestion
            'download_utility': entity_fields.StringField(choices=('curl', 'wget')),
```
since the apidoc says "Must be one of: `curl`, `wget`" and the field is optional (`curl` is used when missing or `None`)

vsedmik (changes requested):
Just a small proposal to consider

nailgun/entities.py
Gauravtalreja1: Good suggestion, but I think we'll still need to set `default='curl'` when None is passed

vsedmik: I tried locally even with `download_utility=None` and it passed, but I agree we can state the default for descriptive reasons here.

dosas: Done

dosas:
@Gauravtalreja1 would you mind to review?

dosas:
Any news on this?

Gauravtalreja1 (approved):
ACK, @dosas Do you mind rebasing this PR for merge? 
Thank you!!

Gauravtalreja1 (approved):

PR #1363, opened 2025-10-10
Drop field mapping for Filter

adamruzicka:
##### Description of changes

A change in Foreman drops the option to override filters as well as dropping the unlimited field. Those fields previously required mapping, there won't be any need for that anymore.
Also taxonomies have been removed from filters.

https://github.com/theforeman/foreman/pull/10370


pnovotny (approved):
LGTM

lhellebr (approved):

PR #1420, opened 2026-05-29
Add clone method to PartitionTable entity

pnovotny:
## Summary
  
Add support for cloning partition tables through the NailGun API client.

## Changes

- Add `path()` method override to `PartitionTable` to handle the `clone` path parameter
- Add `clone()` method that POSTs to `/ptables/<id>/clone` endpoint
- Follows the same pattern used by `JobTemplate`, `ProvisioningTemplate`, `ReportTemplate`, `HostGroup`, and `Role` entities

## Usage

```python
ptable = PartitionTable(server_config=cfg, id=5)
cloned_ptable = ptable.clone(synchronous=True, data={'ptable': {'name': 'cloned_ptable'}})
```

### DIFF
--- a/nailgun/entities.py
+++ b/nailgun/entities.py
@@ -6802,6 +6802,38 @@ class PartitionTable(
         self._meta = {'api_path': 'api/v2/ptables'}
         super().__init__(server_config=server_config, **kwargs)
 
+    def path(self, which=None):
+        """Extend ``nailgun.entity_mixins.Entity.path``.
+
+        The format of the returned path depends on the value of ``which``:
+
+        clone
+            /ptables/<id>/clone
+
+        ``super`` is called otherwise.
+        """
+        if which in ('clone',):
+            return f'{super().path(which="self")}/{which}'
+        return super().path(which)
+
+    def clone(self, synchronous=True, timeout=None, **kwargs):
+        """Clone an existing partition table.
+
+        :param synchronous: What should happen if the server returns an HTTP
+            202 (accepted) status code? Wait for the task to complete if
+            ``True``. Immediately return the server's response otherwise.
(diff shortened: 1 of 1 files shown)


pnovotny:
>     * In `PartitionTable.path`, the `if which in ('clone',):` check is effectively a single-value comparison; using `if which == 'clone':` would be clearer and avoid implying support for multiple values.
Technically true, but I intentionally used `in` operator here so expanding the `path` with more actions in the future is more simple - just adding the new action in the comparison list.
> 
>     * The `clone` method relies on `self.path('clone')`, which in turn assumes an ID-backed `self` path; consider adding a clear guard or error if `id` is missing to avoid generating invalid clone URLs at runtime.
Also true. This ID-checking is however missing also in other entities' `clone` methods and would deserve a broader refactoring in a separate PR, like creating new CloneMixin class with `clone` method that contains also the ID check.

adamruzicka:
Test failures are most likely related.

pnovotny:
> Test failures are most likely related.

@adamruzicka the failure in Code Quality check is not related to this change, it's failing also in other nailgun PRs.
The cause is in the latest `requests` library using type annotations in the get/put/post/... methods while the `nailgun` client doesn't.
E.g.:

nailgun client:
```python
def get(url, params=None, **kwargs):
```
requests:
```python
def get(
    url: _t.UriType, params: _t.ParamsType = None, **kwargs: Unpack[_t.GetKwargs]
) -> Response:
```

adamruzicka:
> The cause is in the latest requests library using type annotations in the get/put/post/... methods while the nailgun client doesn't.

Eh, "good"

adamruzicka (approved):
This is in line with how other clones are implemented

sambible (approved):

PR #1437, opened 2026-09-04
SAT-36510 - Add download policy support Python repositories

ianballou:
* Also remove reference to old Pulp 2 "background" download policy

##### Description of changes

Corrects the API modeling for "generic content type" repositories. The API return data shape is different from the data you send to create the repo. For example you send "includes" to create a Python repo with includes, but in the return, the includes section is inside `generic_remote_options`.

To see the data shape of the return, you need to try accessing the API yourself with a Python repo.

```
# curl -s -k -u admin:changeme https://`hostname`/katello/api/v2/repositories/109 | python3 -m json.tool | grep -E 'includes|excludes|package_types|keep_latest|generic_remote'
    "generic_remote_options": "{\"excludes\":[],\"includes\":[],\"keep_latest_packages\":0,\"package_types\":[]}",
```

See https://github.com/SatelliteQE/robottelo/pull/22638


nailgun/entities.py
```diff
@@ -7399,6 +7409,15 @@ def read(self, entity=None, attrs=None, ignore=None, params=None):
         ignore.add('upstream_password')
         ignore.add('mirror_on_sync')
         ignore.add('download_concurrency')
+        # Generic remote options are sent as top-level params on create/update
+        # but returned as a JSON blob in generic_remote_options on read.
```
qcjames53, line 7413: I'd put this comment on top of the whole `ignore.add` cluster since it applies to everything here.

qcjames53 (changes requested):
One style change requested. This looks solid, though. Thank you!

qcjames53:
I'm not all that familiar with nailgun TBH. Leaned on Claude heavily for this one :)

nailgun/entities.py
ianballou: The ignores above don't relate to `generic_remote_options`, they're different.

ianballou: Take `organization`, for example, that is unrelated to any Pulp repo type of configuration. The other 3 are related to Pulp, but are excluded for different reasons. Upstream pass is never returned, mirror on sync is defunct, and download concurrency isn't returned by the API (oversight probably).

vijaysawant (approved):

qcjames53 (approved):
Thank you for the changes. This looks great to me!

